Issue #69, opened 2021-05-27
add .nvmrc


lstebner:
since this project desires a specific version of node to be installed, it would be nice to have an .nvmrc to help out with setup.

lstebner:
@jeremyckahn is `v15.14.0` ok?

jeremyckahn:
I love this idea! However, I think we'll need to use the previous LTS (`14.17.0`) because Vercel does not yet support anything higher. I recently tried to upgrade to 15 (the highest that I could get working last time I tried), but had to revert to 14: https://github.com/jeremyckahn/farmhand/commit/b9804b0ede00769528e952aff80e2c61488866a9 😕

PR #70, opened 2021-05-27
chore [#69] adding nvmrc

lstebner:
currently needs to be on node 14 because of vercel support, this adds an nvmrc with the LTS v14.17.0.

if using a tool like [nvm](https://github.com/nvm-sh/nvm) or [asdf](https://asdf-vm.com/#/) this makes it easy to keep everyone on the same version of node when working on the app

image


jeremyckahn:
I tested this out and it's working great. Thanks for adding this and welcome to Farmhand @lstebner! 🙂

.nvmrc
```diff
@@ -0,0 +1 @@
+v14.17.0
```
jeremyckahn, line 1: This is working great! 🚀

jeremyckahn (approved):
LGTM!

Issue #67, opened 2021-05-26
Change Farmer’s Log icon to something like a Journal


jeremyckahn:

lstebner:
ended up giving several nav items a new icon to make things more clear :)

PR #72, opened 2021-05-28
refactor [#67] new icons for some navigation elements

lstebner:
updated icon used for Farmer's Log, Stats, Achievements and Bank  Account

before this change:

Screen Shot 2021-05-27 at 5 09 50 PM

after:

Screen Shot 2021-05-27 at 5 09 26 PM

## before merge:

- [x] all tests passing
- [x] icons still work when you click them

image

### DIFF
--- a/src/components/Navigation/Navigation.js
+++ b/src/components/Navigation/Navigation.js
@@ -1,7 +1,10 @@
 import React, { useState, useEffect } from 'react'
 import classNames from 'classnames'
-import AssignmentTurnedInIcon from '@material-ui/icons/AssignmentTurnedIn'
-import AttachMoneyIcon from '@material-ui/icons/AttachMoney'
+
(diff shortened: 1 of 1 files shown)


src/components/Navigation/Navigation.js
```diff
@@ -1,7 +1,10 @@
 import React, { useState, useEffect } from 'react'
 import classNames from 'classnames'
-import AssignmentTurnedInIcon from '@material-ui/icons/AssignmentTurnedIn'
-import AttachMoneyIcon from '@material-ui/icons/AttachMoney'
+
```
lstebner, line 3: I added this line break so the rest can be sorted easily without these moving around

jeremyckahn: Good call, I like this `import` organization pattern!

jeremyckahn (approved):
Looks awesome to me, both visually and code-wise. LGTM, great work! 👍

PR #75, opened 2021-05-28
fix [closes #74] updating icons in Achievement component

### LINKED ISSUE #74: icon in Achievements list needs updated to match navigation
Screen Shot 2021-05-27 at 7 03 43 PM

[this PR](https://github.com/jeremyckahn/farmhand/pull/72) updated the navigation icons and now this list is out of sync so these need updated to match.

lstebner:
a recent change to the navigation icons caused this one to no longer
  match. this change updates the icon, and also changes the 'completed'
  state to be completely CSS based instead of using another icon.

current mismatched version:

Screen Shot 2021-05-27 at 7 03 43 PM

updated states:

Screen Shot 2021-05-28 at 10 37 57 AM

### DIFF
--- a/src/components/Achievement/Achievement.js
+++ b/src/components/Achievement/Achievement.js
@@ -1,5 +1,7 @@
 import React from 'react'
 import classNames from 'classnames'
+
+import AssignmentLateIcon from '@material-ui/icons/AssignmentLate'
 import Card from '@material-ui/core/Card'
 import CardHeader from '@material-ui/core/CardHeader'
 import CardContent from '@material-ui/core/CardContent'
@@ -21,9 +23,7 @@ const Achievement = ({
   >
     <CardHeader
       {...{
-        avatar: (
-          <BeenhereIcon className={classNames({ completed: isComplete })} />
-        ),
+        avatar: isComplete ? <BeenhereIcon /> : <AssignmentLateIcon />,
         title: name,
         subheader: <p>Reward: {rewardDescription}</p>,
       }}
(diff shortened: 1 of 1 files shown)


jeremyckahn:
Great idea! The new "success" achievement state icon matches the Achievements modal trigger, so that's a welcomed improvement.

I'm not quite sold on the new "incomplete" state icon, as now only the icon _color_ conveys the distinction between the two states. This is a step backwards in terms of accessibility (which admittedly isn't the best to begin with), so I'd like to revisit this before it gets merged. Unfortunately Material UI doesn't give us an icon that works totally cohesively as an antonym to the new success state icon. I spent some time looking though [the available icons](https://material-ui.com/components/material-icons/), and in my opinion the previous incomplete state icon (`AssignmentIcon`) is still the best fit here, even though it's not directly related to the new `BeenhereIcon`.

@lstebner what do you think about bringing back `AssignmentIcon` for the incomplete state? I'm not married to that particular icon for this feature, I think we just need the complete/incomplete state icons to be distinct from each other.

lstebner:
good call on it being much less accessible. and now that you've called that out i kinda like the clipboard for "todo", i don't mind going back to the previous, but i also see two others that seem like decent candidates. maybe you already saw them, but what do you think about one of these two @jeremyckahn ? 

Screen Shot 2021-05-28 at 10 03 00 AM

Screen Shot 2021-05-28 at 10 03 04 AM

jeremyckahn:
I think `AssignmentLateIcon` works pretty well here! I like that the incomplete state has a ❗️ and the success state has a ✅. Let's give that a try!

jeremyckahn (approved):
This looks great to me! Thanks for tweaking this. LGTM!

Screen Shot 2021-05-28 at 12 42 16 PM

Issue #73, opened 2021-05-28
Achievement: I am rich


lstebner:
This is the original idea that was contributed:
Reach $1,000,000
(will help speedrunners to save time)
Reward: 10% extra online time money reward

I think we should split this up into at least 2 achievements, maybe 3. I'm unsure what proper rewards would be so these are just ideas:

1. Highest Tax Bracket
how to get: Reach $518,401
reward: all sale prices increase 5%

2. Millionaire
how to get: Reach $1,000,000
reward: all sale prices increase 10%

3. Billionaire
how to get: Reach: $1,000,000,000
reward: all sale prices increase 25%

jeremyckahn:
I really like the tiers here, I think it gets us a bit more mileage out of the core idea of @AbinashWorlds' original suggestion. I'd like to tweak the lowest tier a bit, as I don't know that the joke is quite universal (it's pretty America-centric). I think lowering the goal to $500,000 and calling it something like "Road to Riches" might have a broader appeal.

jeremyckahn:
Implementation notes for this:

  - To create an achievement, add an entry to this array: https://github.com/jeremyckahn/farmhand/blob/3a3354a1394c3ad80cc2b872375558917a767f08/src/data/achievements.js#L33-L244
  - To boost the money earned from a sale, this reducer function will probably need to be modified: https://github.com/jeremyckahn/farmhand/blob/3a3354a1394c3ad80cc2b872375558917a767f08/src/reducers.js#L1111-L1170

lstebner:
haha yes i like "road to riches" and the $500k point for that. thanks for the code pointers! will be back with a PR for this

lstebner:
I added the new achievement definitions in the commit that just got linked, but on the other end of it.. it seems like we would want the price shown before selling to show the increased value, but I think this requires adding something to `getItemCurrentValue` and/or `valueAdjustments`. the way i've implemented it does increase the sales value (i left a console.log in place for quick and dirty testing). could use some advice whenever you can look at the commit @jeremyckahn  and then I'll add some unit tests before finalizing a PR

lstebner:
actually i went ahead and added a few unit tests to the achievements spec to start. i'm curious what you think of the style though @jeremyckahn . it's a little different than the other one in there, but it looks like we need to add some more coverage :). draft PR 👉 https://github.com/jeremyckahn/farmhand/pull/77/

PR #77, opened 2021-05-29
Lstebner/achievement i am rich

lstebner:
This PR implements 3 new achievements in a series referred to as "I am Rich". this closes issue #73 , and can be tested here https://farmhand-git-lstebner-achievement-i-am-rich-jeremyckahn.vercel.app/

this PR also adds a few new utils which were used to implement this feature

- `percentageString` util to convert a floating percent to a string value
- `getSalePriceMultiplier` looks at achievements and returns a multiplier value

completed achievements:

Screen Shot 2021-05-29 at 2 37 17 PM

incomplete achievements:

Screen Shot 2021-05-30 at 9 35 01 AM

increased price in sell view:

Screen Shot 2021-05-29 at 2 34 28 PM

unit tests:

Screen Shot 2021-05-29 at 1 32 34 PM

Screen Shot 2021-05-29 at 1 33 03 PM

### DIFF
--- a/src/components/Item/Item.js
+++ b/src/components/Item/Item.js
@@ -21,6 +21,7 @@ import {
   getFinalCropItemFromSeedItem,
   getItemCurrentValue,
   getResaleValue,
+  getSalePriceMultiplier,
   moneyString,
   integerString,
 } from '../../utils'
@@ -145,6 +146,9 @@ export const Item = ({
 
   const avatar = <img {...{ src: items[id] }} alt={name} />
 
+  const sellPrice =
+    adjustedValue * getSalePriceMultiplier(completedAchievements)
+
   return (
     <Card
       {...{
@@ -221,7 +225,7 @@ export const Item = ({
                     <span>
                       Sell price:{' '}
                       <AnimatedNumber
-                        {...{ number: adjustedValue, formatter: moneyString }}
+                        {...{ number: sellPrice, formatter: moneyString }}
                       />
                     </span>
                   </Tooltip>
--- a/src/constants.js
+++ b/src/constants.js
@@ -110,6 +110,8 @@ export const DAILY_FINANCIAL_HISTORY_RECORD_LENGTH = 7
 
 export const RECIPE_INGREDIENT_VALUE_MULTIPLIER = 1.25
 
+export const I_AM_RICH_BONUSES = [0.05, 0.1, 0.25]
+
 export const PERSISTED_STATE_KEYS = [
   'completedAchievements',
   'cowBreedingPen',
--- a/src/data/achievements.js
+++ b/src/data/achievements.js
@@ -7,9 +7,10 @@ import {
   integerString,
   memoize,
   moneyTotal,
+  percentageString,
 } from '../utils'
 import { cropLifeStage, standardCowColors } from '../enums'
-import { COW_FEED_ITEM_ID } from '../constants'
+import { COW_FEED_ITEM_ID, I_AM_RICH_BONUSES } from '../constants'
 import { addItemToInventory } from '../reducers'
 
 import { itemsMap } from './maps'
@@ -240,6 +241,39 @@ const achievements = [
       ...state,
       inventoryLimit: state.inventoryLimit + reward,
     }),
+  }))(),
+
+  ((goal = 500000) => ({
+    id: 'i-am-rich-1',
+    name: 'I am Rich!',
+    description: `Earn ${dollarString(goal)}.`,
+    rewardDescription: `All sales receive a ${percentageString(
+      I_AM_RICH_BONUSES[0]
+    )} bonus`,
+    condition: state => state.revenue >= goal,
+    reward: state => state,
+  }))(),
+
+  ((goal = 1000000) => ({
+    id: 'i-am-rich-2',
+    name: 'Millionaire',
+    description: `Earn ${dollarString(goal)}.`,
+    rewardDescription: `All sales receive a ${percentageString(
+      I_AM_RICH_BONUSES[1]
+    )} bonus`,
+    condition: state => state.revenue >= goal,
+    reward: state => state,
+  }))(),
+
+  ((goal = 1000000000) => ({
+    id: 'i-am-rich-3',
+    name: 'Billionaire',
+    description: `Earn ${dollarString(goal)}.`,
+    rewardDescription: `All sales receive a ${percentageString(
+      I_AM_RICH_BONUSES[2]
+    )} bonus`,
+    condition: state => state.revenue >= goal,
+    reward: state => state,
   }))(),
 ]
 
--- a/src/data/achievements.test.js
+++ b/src/data/achievements.test.js
@@ -43,3 +43,47 @@ describe('harvest-crop', () => {
     })
   })
 })
+
+const iAmRichVariants = [
+  ['i-am-rich-1', 500000, 'Earn $500,000.', 'All sales receive a 5% bonus'],
+  ['i-am-rich-2', 1000000, 'Earn $1,000,000.', 'All sales receive a 10% bonus'],
+  [
+    'i-am-rich-3',
+    1000000000,
+    'Earn $1,000,000,000.',
+    'All sales receive a 25% bonus',
+  ],
+]
+
+describe.each(iAmRichVariants)(
+  'I am Rich variants',
+  (id, goal, description, rewardDescription) => {
+    describe(id, () => {
+      test('has the expected description', () => {
+        expect(achievementsMap[id].description).toEqual(description)
+      })
+
+      test('has the expected rewardDescription', () => {
+        expect(achievementsMap[id].rewardDescription).toEqual(rewardDescription)
+      })
+
+      test(`is achieved when revenue is greater than or equal to ${goal}`, () => {
+        const achievement = achievementsMap[id]
+        const state = {
+          revenue: goal,
+        }
+
+        expect(achievement.condition(state)).toEqual(true)
+      })
+
+      test(`is not achieved when revenue is less than ${goal}`, () => {
+        const achievement = achievementsMap[id]
+        const state = {
+          revenue: goal - 1,
+        }
+
+        expect(achievement.condition(state)).toEqual(false)
+      })
+    })
+  }
+)
--- a/src/reducers.js
+++ b/src/reducers.js
@@ -41,6 +41,7 @@ import {
   getRandomUnlockedCrop,
   getRangeCoords,
   getResaleValue,
+  getSalePriceMultiplier,
   getSeedItemIdFromFinalStageCropItemId,
   inventorySpaceRemaining,
   isItemAFarmProduct,
@@ -1120,7 +1121,12 @@ export const sellItem = (state, { id }, howMany = 1) => {
   }
 
   const item = itemsMap[id]
-  const { itemsSold, money: initialMoney, valueAdjustments } = state
+  const {
+    completedAchievements,
+    itemsSold,
+    money: initialMoney,
+    valueAdjustments,
+  } = state
   const oldLevel = levelAchieved(farmProductsSold(itemsSold))
   let { loanBalance } = state
 
@@ -1129,7 +1135,6 @@ export const sellItem = (state, { id }, howMany = 1) => {
     : getAdjustedItemValue(valueAdjustments, id)
   const saleIsGarnished = isItemAFarmProduct(item)
   let saleValue = 0
-
   for (let i = 0; i < howMany; i++) {
     const loanGarnishment = saleIsGarnished
       ? Math.min(
@@ -1137,7 +1142,9 @@ export const sellItem = (state, { id }, howMany = 1) => {
           castToMoney(adjustedItemValue * LOAN_GARNISHMENT_RATE)
         )
       : 0
-    const garnishedProfit = adjustedItemValue - loanGarnishment
+    const garnishedProfit =
+      adjustedItemValue * getSalePriceMultiplier(completedAchievements) -
+      loanGarnishment
     loanBalance = moneyTotal(loanBalance, -loanGarnishment)
     saleValue = moneyTotal(saleValue, garnishedProfit)
   }
--- a/src/utils.js
+++ b/src/utils.js
@@ -48,6 +48,7 @@ import {
   COW_WEIGHT_MULTIPLIER_MINIMUM,
   DAILY_FINANCIAL_HISTORY_RECORD_LENGTH,
   HUGGING_MACHINE_ITEM_ID,
+  I_AM_RICH_BONUSES,
   INITIAL_FIELD_HEIGHT,
   INITIAL_FIELD_WIDTH,
   INITIAL_SPRINKLER_RANGE,
@@ -191,6 +192,12 @@ export const dollarString = number => formatNumber(number, '$0,0')
  * @returns {string} Number string with commas.
  */
 export const integerString = number => formatNumber(number, '0,0')
+
+/**
+ * @param {number} number A float
+ * @returns {string} the float converted to a full number with a % added
+ */
+export const percentageString = number => `${Math.round(number * 100)}%`
 
 /**
  * @param {string} itemId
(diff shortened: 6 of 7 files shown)


src/reducers.js
```diff
@@ -1129,6 +1129,17 @@ export const sellItem = (state, { id }, howMany = 1) => {
...
+
+  if (state.completedAchievements['i-am-rich-3']) {
+    salePriceMultiplier = 1.25
+  } else if (state.completedAchievements['i-am-rich-2']) {
+    salePriceMultiplier = 1.1
+  } else if (state.completedAchievements['i-am-rich-3']) {
+    salePriceMultiplier = 1.05
+  }
+
+  console.log('price multiplier value', salePriceMultiplier)
```
jeremyckahn, line 1142: I know you're aware of this, I just want to add this comment so we don't forget to remove it before merging! 🙂

src/reducers.js
```diff
@@ -1129,6 +1129,17 @@ export const sellItem = (state, { id }, howMany = 1) => {
...
   let saleValue = 0
+  let salePriceMultiplier = 1
+
+  if (state.completedAchievements['i-am-rich-3']) {
+    salePriceMultiplier = 1.25
+  } else if (state.completedAchievements['i-am-rich-2']) {
+    salePriceMultiplier = 1.1
+  } else if (state.completedAchievements['i-am-rich-3']) {
+    salePriceMultiplier = 1.05
+  }
```
jeremyckahn, line 1140: This comment is aimed at your question in the `getItemCurrentValue` util, but makes more sense to answer here.

> should valueAdjustments have a "global" adjustment or something for any permanent global increases?

We should avoid using `valueAdjustments` to calculate sale bonuses, as that object is intended to only define market fluctuations. With that said, my intuition is to move this `salePriceMultiplier` calculation to a helper function in `src/utils.js` and use that helper both here and in `getItemCurrentValue`. What do you think of that approach?

jeremyckahn (changes requested):
This is off to a great start! I added comments throughout, but let me know if anything is unclear and we can discuss further.

src/data/achievements.js
```diff
@@ -241,6 +241,33 @@ const achievements = [
       inventoryLimit: state.inventoryLimit + reward,
     }),
   }))(),
+
+  ((goal = 500000) => ({
+    id: 'i-am-rich-1',
+    name: 'I am Rich!',
+    description: `Earn ${dollarString(goal)}`,
+    rewardDescription: 'All sales get a 5% bonus',
+    condition: state => state.revenue >= goal,
```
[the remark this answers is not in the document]
lstebner: i scoped out the 'stats' window to see what that was using for a hint

src/data/achievements.js
```diff
@@ -241,6 +241,33 @@ const achievements = [
       inventoryLimit: state.inventoryLimit + reward,
     }),
   }))(),
+
+  ((goal = 500000) => ({
+    id: 'i-am-rich-1',
+    name: 'I am Rich!',
+    description: `Earn ${dollarString(goal)}`,
+    rewardDescription: 'All sales get a 5% bonus',
```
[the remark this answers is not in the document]
lstebner: makes sense! i made a hash of `I_AM_RICH_BONUSES`, the keys line up to the # used in the 'id', which I thought was simpler to understand than just an array, but am open to whatever you prefer here.

src/data/achievements.test.js
```diff
@@ -43,3 +43,42 @@ describe('harvest-crop', () => {
...
+      id: 'i-am-rich-2',
+      goal: { revenue: 1000000 },
+    },
+    {
+      id: 'i-am-rich-3',
+      goal: { revenue: 1000000000 },
+    },
+  ]
+
+  for (let variant of variants) {
```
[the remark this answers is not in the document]
lstebner: today i learned about `describe.each`. looks perfect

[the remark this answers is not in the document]
lstebner: updated those

src/reducers.js
```diff
@@ -1129,6 +1129,17 @@ export const sellItem = (state, { id }, howMany = 1) => {
     : getAdjustedItemValue(valueAdjustments, id)
   const saleIsGarnished = isItemAFarmProduct(item)
   let saleValue = 0
+  let salePriceMultiplier = 1
+
+  if (state.completedAchievements['i-am-rich-3']) {
+    salePriceMultiplier = 1.25
+  } else if (state.completedAchievements['i-am-rich-2']) {
+    salePriceMultiplier = 1.1
+  } else if (state.completedAchievements['i-am-rich-3']) {
```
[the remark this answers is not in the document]
lstebner: good catch!

src/reducers.js
lstebner: removing this now

src/reducers.js
lstebner: i had a feeling `valueAdjustments` was something market specific. can we update the variable name? (is that easy? to like, `marketValueAdjustments`?). or add some documentation around it. i didn't see any explanation for the variable anywhere. whatever you prefer @jeremyckahn . i can also do the work if you let me know which you prefer.

onto the util...that makes sense to me! that will scale nicely later on.

lstebner: lots of unit tests, and a util or two later i think i got there!

src/data/achievements.test.js
```diff
@@ -43,3 +43,47 @@ describe('harvest-crop', () => {
...
 })
+
+const iAmRichVariants = [
+  ['i-am-rich-1', 500000, 'Earn $500,000', 'All sales receive a 5% bonus.'],
+  ['i-am-rich-2', 1000000, 'Earn $1,000,000', 'All sales receive a 10% bonus.'],
+  [
+    'i-am-rich-3',
+    1000000000,
+    'Earn $1,000,000,000',
+    'All sales receive a 25% bonus.',
```
lstebner, line 54: not sure how you feel about these kinda copy tests, they could also be snapshots @jeremyckahn . i kept manually testing that the copy was correct though so it felt right to add an automated test for it

src/components/Item/Item.js
```diff
@@ -145,6 +146,8 @@ export const Item = ({
 
   const avatar = <img {...{ src: items[id] }} alt={name} />
 
+  adjustedValue *= getSalePriceMultiplier(completedAchievements)
```
jeremyckahn, line 149: Rather than mutating `adjustedValue` here, I think we're going to need to add an additional number to represent `adjustedValue * getSalePriceMultiplier(completedAchievements)`. The reason for this is that `adjustedValue`, at least for non-Shop items, specifically defines the value of an item adjusted for current market values. Modifying it here will break the calculations at these points deeper in the component:

https://github.com/jeremyckahn/farmhand/blob/26d249bac2678abda5e3417d951ff69224632d92/src/components/Item/Item.js#L204-L206

https://github.com/jeremyckahn/farmhand/blob/26d249bac2678abda5e3417d951ff69224632d92/src/components/Item/Item.js#L233-L235

This is a bit confusing because #73 makes the semantics of this variable a bit unclear. It may make sense to rename `adjustedValue` to something like `marketAdjustedValue`, since the current name is too vague now. I think "adjustedValue" is good for representing "the final number that is presented to the user," in case you need some help with the variable naming here.

src/utils.js
```diff
@@ -192,6 +193,12 @@ export const dollarString = number => formatNumber(number, '$0,0')
  */
 export const integerString = number => formatNumber(number, '0,0')
 
+/**
+ * @param {number} a float
```
jeremyckahn, line 197: Per JSDoc syntax, this needs to reference the name of the variable. So, this will need to be something like:

```
* @param {number} number A float
```

src/utils.js
```diff
@@ -989,3 +997,21 @@ export const getCostOfNextStorageExpansion = currentInventoryLimit => {
...
+  if (completedAchievements['i-am-rich-3']) {
+    salePriceMultiplier += I_AM_RICH_BONUSES[3]
+  } else if (completedAchievements['i-am-rich-2']) {
+    salePriceMultiplier += I_AM_RICH_BONUSES[2]
+  } else if (completedAchievements['i-am-rich-1']) {
+    salePriceMultiplier += I_AM_RICH_BONUSES[1]
+  }
+
+  return salePriceMultiplier
+}
```
jeremyckahn, line 1017: Love it!

jeremyckahn (changes requested):
Awesome progress so far! There's a few issues that need to be addressed, but I'm really excited to see how well this is coming along!

src/reducers.js
jeremyckahn: > i had a feeling `valueAdjustments` was something market specific. can we update the variable name? (is that easy? to like, `marketValueAdjustments`?). or add some documentation around it. i didn't see any explanation for the variable anywhere. whatever you prefer @jeremyckahn . i can also do the work if you let me know which you prefer.

Whoops, I saw this after the fact! Sorry about that. I've effectively addressed this here: https://github.com/jeremyckahn/farmhand/pull/77/files#r641988878

tl;dr: Yes, we should definitely rename `valueAdjustments` to something like `marketValueAdjustments` for the relative value calculations, but we'll also need to having some sort of "grand total" value that is presented to the user as well.

src/data/achievements.js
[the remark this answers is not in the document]
jeremyckahn: I noticed this reply late, but I addressed it in https://github.com/jeremyckahn/farmhand/pull/77/files#r641982208, but yes I (personally) think it would a little more intuitive for this to be an `Array`. 🙂

src/components/Item/Item.js
lstebner: i see now, thanks for pointing that out. i think it actually makes the most sense to add a `sellPrice` variable and only adjust the "Sell price" field. i would like to see `adjustedValue` updated as you said, but there are many other similar variables that i would also want to update like `previousDatAdjustedValue` and `historicalValueAdjustments` so I think it's best to leave those all for now. how does that sound?

src/utils.js
lstebner: oops, number/number! good catch

src/data/achievements.js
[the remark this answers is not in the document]
lstebner: sounds good! going to resolve this conversation since that one exists now

src/data/achievements.test.js
lstebner: sounds good to me!

src/data/achievements.js
```diff
@@ -241,6 +242,40 @@ const achievements = [
...
     }),
   }))(),
+
+  ((goal = 500000) => ({
+    id: 'i-am-rich-1',
+    name: 'I am Rich!',
+    description: `Earn ${dollarString(goal)}`,
+    rewardDescription: `All sales receive a ${percentageString(
+      I_AM_RICH_BONUSES[1],
+      '%'
```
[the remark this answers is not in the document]
lstebner: yup! a formatNumber attempt hah. thanks for noticing that

lstebner:
@jeremyckahn pushed another round of updates for all your latest comments

src/components/Item/Item.js
```diff
@@ -145,6 +146,9 @@ export const Item = ({
 
   const avatar = <img {...{ src: items[id] }} alt={name} />
 
+  const sellPrice =
```
jeremyckahn, line 149: This is exactly what I was thinking! I think we're going to need `sellPrice` instead of `adjustedValue` here on like 294 as well: https://github.com/jeremyckahn/farmhand/blob/cea1cb22029552865edf3a3f5c45556e57dd9f20/src/components/Item/Item.js#L288-L300

Once that's fixed, I think this should be all good! 🙂

src/constants.js
```diff
@@ -110,6 +110,8 @@ export const DAILY_FINANCIAL_HISTORY_RECORD_LENGTH = 7
 
 export const RECIPE_INGREDIENT_VALUE_MULTIPLIER = 1.25
 
+export const I_AM_RICH_BONUSES = [0.05, 0.1, 0.25]
```
jeremyckahn, line 113: I love this! Thank for for making this change! 🙏

src/data/achievements.js
```diff
@@ -241,6 +242,39 @@ const achievements = [
       inventoryLimit: state.inventoryLimit + reward,
     }),
   }))(),
+
+  ((goal = 500000) => ({
+    id: 'i-am-rich-1',
+    name: 'I am Rich!',
+    description: `Earn ${dollarString(goal)}`,
```
jeremyckahn, line 249: Sorry I missed this previously, but these `description` strings will also need a period at the end to be consistent with the other achievements.

src/utils.js
```diff
@@ -204,6 +211,7 @@ const getItemBaseValue = itemId => itemsMap[itemId].value
  * @returns {number}
  */
 export const getItemCurrentValue = ({ id }, valueAdjustments) =>
+  // should valueAdjustments have a "global" adjustment or something for any permanent global increases?
```
jeremyckahn, line 214: Since it seems like we're getting close to the end with this PR, please be sure to remove this comment (unless it's something you feel is unresolved). 🙂

src/utils.js
```diff
@@ -232,7 +240,8 @@ export const isItemSoldInShop = ({ id }) => Boolean(shopInventoryMap[id])
  * @param {farmhand.item} item
  * @returns {number}
  */
-export const getResaleValue = ({ id }) => itemsMap[id].value / 2
+export const getResaleValue = ({ id }, completedAchievements) =>
```
jeremyckahn, line 243: It looks like we can drop the `completedAchievements` parameter here.

jeremyckahn (changes requested):
Thanks for the latest iteration! This is looking great and I think we're almost there. I saw a few mostly minor regression issues to address, but I think the hard problems are solved at this point!

src/components/Item/Item.js
jeremyckahn: Oh yeah, those variables are functionally affected by this feature at all, and changing their name is actually quite a bit more involved because they are part of the persisted game data. They're named slightly more vaguely than I'd like right now, but it's not worth changing them in this PR.

src/components/Item/Item.js
lstebner: oh wow i did not see that one at all!

src/utils.js
lstebner: another good catch! we resolved this with the discussion about these being only for market adjustments 👍

src/utils.js
lstebner: oops! yep, didn't end up going that route. removing

src/components/Item/Item.js
lstebner: actually, i'm not sure this one was supposed to change 🤔 . this is for purchasing, right? we only want to affect sell price I think

jeremyckahn: Ah! Great catch, you're right. Thanks for scrutinizing the code extra closely here! 😅

jeremyckahn:
The code looks good to me, but I just tried to test out the new achievement rewards and they don't seem to factor in the sales bonus. 😕

Here's where I tested out the new feature: https://farmhand-git-lstebner-achievement-i-am-rich-jeremyckahn.vercel.app/

Here's the save file I used to test this out: https://gist.github.com/jeremyckahn/4dc3bd35f6e2db3c47b24dcc62a5db27

(You can save that JSON data to a local file and then import it into the game for testing.)

You'll see there that Carrots have a boosted value of $26.25, but once sold, the player only gets the non-boosted value of $25. I'll take a look to see what part of the code might need to be revisited to fix this too.

jeremyckahn:
> The code looks good to me, but I just tried to test out the new achievement rewards and they don't seem to factor in the sales bonus. 😕
> 
> Here's where I tested out the new feature: https://farmhand-git-lstebner-achievement-i-am-rich-jeremyckahn.vercel.app/
> 
> Here's the save file I used to test this out: https://gist.github.com/jeremyckahn/4dc3bd35f6e2db3c47b24dcc62a5db27
> 
> (You can save that JSON data to a local file and then import it into the game for testing.)
> 
> You'll see there that Carrots have a boosted value of $26.25, but once sold, the player only gets the non-boosted value of $25. I'll take a look to see what part of the code might need to be revisited to fix this too.

Based on some quick experimentation, it looks like this change might be all that's needed here: 

```diff
diff --git a/src/reducers.js b/src/reducers.js
index a10d58a..617bc03 100644
--- a/src/reducers.js
+++ b/src/reducers.js
@@ -1142,12 +1142,11 @@ export const sellItem = (state, { id }, howMany = 1) => {
           castToMoney(adjustedItemValue * LOAN_GARNISHMENT_RATE)
         )
       : 0
-    const garnishedProfit = adjustedItemValue - loanGarnishment
+    const garnishedProfit =
+      adjustedItemValue * getSalePriceMultiplier(completedAchievements) -
+      loanGarnishment
     loanBalance = moneyTotal(loanBalance, -loanGarnishment)
-    saleValue = moneyTotal(
-      saleValue * getSalePriceMultiplier(completedAchievements),
-      garnishedProfit
-    )
+    saleValue = moneyTotal(saleValue, garnishedProfit)
   }
 
   if (saleIsGarnished) {
```

The tests pass with this, and I'm seeing the correct behavior in-game with it!

jeremyckahn (approved):
I've QA'ed and verified this on https://farmhand-git-lstebner-achievement-i-am-rich-jeremyckahn.vercel.app/, and everything seems to be working perfectly! Thank you @lstebner for sticking with this and bringing it through to completion. This feature adds some much-needed late-game content and lays the groundwork for more interesting mechanics down the line. Awesome work! LGTM! 🚀

PR #96, opened 2021-06-01
[closes #95] feat: make cows look in moove direction

### LINKED ISSUE #95: make cows flip horizontally
the cows always face right currently, but it would add some extra life to them if they had a horizontally flipped version for when they are moving to the left, instead of the right

lstebner:
cows currently face right all the time, but this change adds some
  logic so that when they are moving left, they also look to the left.
  this was doing with CSS using transform, so it also inherited a nice
  little transition property that adds some animation to the whole
  thing.

  additionally: it was noticed that X position was being applied as the
  `top` style and Y as the `left` so this was flipped which seems more
  expected.

## here's a little gif of it in action

happycows

### DIFF
--- a/src/components/CowPen/CowPen.js
+++ b/src/components/CowPen/CowPen.js
@@ -33,7 +33,7 @@ export class Cow extends Component {
   // This MUST be kept in sync with $hug-animation-duration in CowPen.sass.
   static hugAnimationDuration = 750
 
-  // Only moves the cow withing the middle 80% of the pen
+  // Only moves the cow within the middle 80% of the pen
   static randomPosition = () => 10 + Math.random() * 80
 
   get waitVariance() {
@@ -163,7 +163,6 @@ export class Cow extends Component {
   componentWillUnmount() {
     ;[
       this.repositionTimeoutId,
-      this.animateMovementTimeoutId,
       this.animateHugTimeoutId,
     ].forEach(clearTimeout)
 
--- a/src/components/CowPen/CowPen.js
+++ b/src/components/CowPen/CowPen.js
@@ -27,8 +27,8 @@ export class Cow extends Component {
   animateHugTimeoutId = null
   tweenable = new Tweenable()
 
-  static movementAnimationDuration = 3000
   static flipAnimationDuration = 1000
+  static transitionAnimationDuration = 3000
 
   // This MUST be kept in sync with $hug-animation-duration in CowPen.sass.
   static hugAnimationDuration = 750
@@ -125,7 +125,7 @@ export class Cow extends Component {
       await this.tweenable.tween({
         from: { x, y },
         to: { x: newX, y: Cow.randomPosition() },
-        duration: Cow.movementAnimationDuration,
+        duration: Cow.transitionAnimationDuration,
         render: ({ x, y }) => {
           this.setState({ x, y })
         },
--- a/src/components/CowPen/CowPen.js
+++ b/src/components/CowPen/CowPen.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import { array, bool, func, object, string } from 'prop-types'
 import classNames from 'classnames'
+import { Tweenable } from 'shifty'
 import Tooltip from '@material-ui/core/Tooltip'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faHeart } from '@fortawesome/free-solid-svg-icons'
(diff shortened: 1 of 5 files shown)


src/enums.js
```diff
@@ -80,11 +80,7 @@ export const itemType = enumify([
  * @property farmhand.module:enums.fertilizerType
  * @enum {string}
  */
-export const fertilizerType = enumify([
-  'NONE',
-  'STANDARD',
-  'RAINBOW',
-])
+export const fertilizerType = enumify(['NONE', 'STANDARD', 'RAINBOW'])
```
lstebner, line 83: eslint or prettier did this :)

jeremyckahn:
I really love this idea! Thanks for putting this PR together. I'd like to work on the "flip" animation a bit more before moving forward with this. I feel that it would be a bit stronger if the cow completing its flip before moving to the new target. So the animation sequence would be:

  1. Flip, then
  2. Start bouncing
  3. Transition to the new target while still bouncing

I've mocked up how this might look here (I cheated and used jQuery): https://codepen.io/jeremyckahn/pen/xxqYxBZ?editors=0110

Kapture 2021-06-01 at 21 21 17


What do you think about this animation, @lstebner?

lstebner:
awesome @jeremyckahn ! i really like it. also very much appreciate the codepen! would have taken me a long time. i'll work on integrating that into this PR soon

jeremyckahn:
I FINALLY had some time to work on this! Sorry for the long delay on this. This commit implements the look I was going for: https://github.com/jeremyckahn/farmhand/pull/96/commits/682a162d43246083d0c4d12e2ec3753678a615dd

https://user-images.githubusercontent.com/366330/122500437-66c43980-cfb8-11eb-895d-a8b6c4834f32.mp4

I want to iterate on this a little bit more and see if I can replace the `left`/`right` CSS transitions with Shifty since it's already in the mix. We're getting close, though!

jeremyckahn:
Cows now transition to their new X/Y position with JavaScript in https://github.com/jeremyckahn/farmhand/pull/96/commits/6f25849453ae9f402140954ef6ba9e96cc50b0c3!

I'd like to take some more time to self-review and clean this up in the coming days, and then I'll send it back to you @lstebner to review as well. 🙂

jeremyckahn (approved):
I'm feeling pretty good about this change now! The code and animation all look great to me. Sorry I kinda commandeered this PR @lstebner... I got into the zone with tweaking things. 😅

The UX is so much better with this PR! Since this now contains a lot of my code too, I'll let you review and approve and either you or I can merge and ship as a `patch` release.

src/components/CowPen/CowPen.js
```diff
@@ -1,34 +1,39 @@
...
   }
 
   repositionTimeoutId = null
-  animateMovementTimeoutId = null
   animateHugTimeoutId = null
+  tweenable = new Tweenable()
 
-  // This MUST be kept in sync with $movement-animation-duration in CowPen.sass.
-  static movementAnimationDuration = 3000
+  static flipAnimationDuration = 1000
```
lstebner, line 30: we can move to native CSS variables and use `getComputedStyle` if you would like to remove this need to update code in 2 places. alternatively, the javascript can be used to seed the CSS variable valuable depending on which one you feel is easier to maintain (not intending this comment holds up this PR, but it's just some food for thought)

src/components/CowPen/CowPen.sass
```diff
@@ -29,6 +25,9 @@
       z-index: -20
 
     &.is-selected
+      // Render the selected cow on top of all others
+      z-index: 40
```
lstebner, line 29: z-indexes are my mortal enemy haha. i strongly believe 0, 1, and -1 are the only acceptable values, but it can take some reworking to get to this point. not sure if it's something you care about, but if it is we can work on taming these sometime

lstebner:
> I'm feeling pretty good about this change now! The code and animation all look great to me. Sorry I kinda commandeered this PR @lstebner... I got into the zone with tweaking things. 😅
> 
> The UX is so much better with this PR! Since this now contains a lot of my code too, I'll let you review and approve and either you or I can merge and ship as a `patch` release.

this is awesome, so glad you were able to find time to do this! i'm super excited for this update. i left 2 comments but they aren't blocking on this PR, i approve! also tested it out and it seems solid. ship it asap!

lstebner:
oh yeah and i can't approve this cause i am still the author of this PR, but i officially approve 🎉

src/components/CowPen/CowPen.js
jeremyckahn: I'm open minded here! I think the variables in question are inherently a display concern, so I'd prefer the source of truth live in a Sass/CSS file.

src/components/CowPen/CowPen.sass
jeremyckahn: I forgot where I learned to use factors of 20 for z-index, but it's served me reasonably well over the years. I think your CSS skills are quite a bit stronger than mine, so I'm happy to defer to your judgment for these sorts of things. Feel free to wrangle these as you see fit! 🙂

Issue #98, opened 2021-06-03
bug with number inputs on mobile device


lstebner:
when on my phone, i can't enter numbers into the number inputs. whenever i try to delete the initial 1, the cursor moves to the left of it instead, and then typing in new numbers results in the 1 always being at the end. i.e. i try to enter '20' and i get '201'. it seems impossible to actually enter digits into here. i'll get a video of this bug if needed, but it seems easy to recreate for me via any of the buy or sell quantity inputs

jeremyckahn:
Yeah, the UX for these inputs leave a lot to be desired. They do work on mobile (you need to double-tap to Select All and then overwrite), but they could be a lot better. There's a related card on the Idea Board and I just linked it to this issue: https://github.com/jeremyckahn/farmhand/projects/1#card-61674937

jeremyckahn:
With #121 merged, do we want to close this issue? It doesn't quite fix the problem as stated, but it implements a reasonable workaround.

lstebner:
sorry for not responding sooner, but i can safely say after using the updated version my initial complaints are resolved. the UI/UX is much better now!

PR #121, opened 2021-06-17
QuantityInput: clear input onFocus

seawolff:
#### What does this PR do?
Sorry this is somewhat of an opinionated change. But I found it annoying that I couldn't clear the default value of the controlled component completely when writing out a desired amount. i.e. if the value amount was `1` and I wanted `50` I'd have to start typing my desired amount and clear the value. So I am proposing we just clear the quantity input value on focus.

Screenflick Movie 135
Screenflick Movie 134

#### How can this change be manually tested?
Select the Quantity Input for both buying and selling an item and notice that when you select the input the previous default value clears. The Buy/Sell button should be disabled if an undefined value is current set.

#### Questions or concerns about this change?
I could perhaps make the disabled item button check a bit cleaner but for now I just added to the conditional to check if the purchase or sell quantity isn't defined.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,6 +1,7 @@
 Special thanks to everyone who has contributed to Farmhand's development!
 
 - Luke Stebner ([@lstebner](https://github.com/lstebner))
+- Chris Wolff ([@seawolff](https://github.com/seawolff))
 
 Take a look at the GitHub repo's [Contributors page](https://github.com/jeremyckahn/farmhand/graphs/contributors) to see who's contributed what.
 
--- a/src/components/Item/Item.js
+++ b/src/components/Item/Item.js
@@ -291,7 +291,8 @@ export const Item = ({
                 color: 'primary',
                 disabled:
                   purchaseQuantity === 0 ||
-                  adjustedValue * purchaseQuantity > money,
+                  adjustedValue * purchaseQuantity > money ||
+                  !purchaseQuantity,
                 onClick: handleItemPurchase,
                 variant: 'contained',
               }}
@@ -315,7 +316,7 @@ export const Item = ({
               {...{
                 className: 'sell',
                 color: 'secondary',
-                disabled: sellQuantity === 0,
+                disabled: sellQuantity === 0 || !sellQuantity,
                 onClick: handleItemSell,
                 variant: 'contained',
               }}
--- a/src/components/Item/Item.js
+++ b/src/components/Item/Item.js
@@ -290,9 +290,7 @@ export const Item = ({
                 className: 'purchase',
                 color: 'primary',
                 disabled:
-                  purchaseQuantity === 0 ||
-                  adjustedValue * purchaseQuantity > money ||
-                  !purchaseQuantity,
+                  !purchaseQuantity || adjustedValue * purchaseQuantity > money,
                 onClick: handleItemPurchase,
                 variant: 'contained',
               }}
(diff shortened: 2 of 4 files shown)


jeremyckahn:
It looks like this PR fixes #98, yeah @lstebner?

jeremyckahn (changes requested):
This is a really strong start to a much-appreciated change! I left a few notes for some minor things to revisit. There's one thing missing, and that's the same treatment being given to Recipe cards. It's worth adding that before shipping this to keep the UX consistent throughout the game.

You should be able to apply similar logic to [`Recipe.js`](https://github.com/jeremyckahn/farmhand/blob/wolff/quanity-input-onfocus-clear/src/components/Recipe/Recipe.js). I think you'll want to look here: https://github.com/jeremyckahn/farmhand/blob/2b86fba4fba8d9a0b4cbb8d76f6a64e57aa51bd9/src/components/Recipe/Recipe.js#L80-L90

And here: https://github.com/jeremyckahn/farmhand/blob/2b86fba4fba8d9a0b4cbb8d76f6a64e57aa51bd9/src/components/Recipe/Recipe.js#L133-L141

lstebner:
i'm noticing a small bug where after the input is cleared now it is impossible to use the up/down arrows to adjust the quantity. i wouldn't actually expect the input to clear on focus as a user, but instead after i hit 'delete' it should be ok with the input staying empty rather than trying to force a value to always be there.

seawolff:
@lstebner yeah perhaps this brings up a good point. Perhaps we should just remove the min-max validation check and allow for a non-controlled component so the input can be cleared out.

seawolff:
...actually, what if I just disabled the arrows as well?

lstebner:
that would make this a lot less user friendly in my opinion when the goal is to make it more intuitive. i still think the clearing on focus is already unexpected.

seawolff:
@jeremyckahn okay I think 🤞 this should be GTG

jeremyckahn:
> @jeremyckahn okay I think 🤞 this should be GTG

We’re getting there! I noticed that the arrow buttons don’t work at all for recipes. I’m mobile now but I can take a look at the code this weekend sometime to see what the issue may be!

jeremyckahn:
As for general UX around these components, I expect that it will be hard to get them perfect. We need some level of input constraint. There’s a lot of ways to solve that problem, ranging from simply showing an error UI to preventing invalid input (which is what we currently have). I’m open to solving this problem in a more effective or simple way. I don’t think we need to aim for perfection with this PR (or any other one for that matter), but each PR should at least be a net positive for the user and not be introducing greater problems for us to deal with later.

I think the basic idea of clearing the input upon focus is a step in the right direction and one I’d be happy to ship. I think there’s more work to be done here, but we can iterate with more PRs until we get it right!

seawolff:
> > @jeremyckahn okay I think 🤞 this should be GTG
> 
> 
> 
> We’re getting there! I noticed that the arrow buttons don’t work at all for recipes. I’m mobile now but I can take a look at the code this weekend sometime to see what the issue may be!

Ah sorry. I'll take a closer look Monday. I had a quick second and tried to fix. I'll take a more thorough approach Monday. Thanks for the feedback.

seawolff:
Ah I found the issue. I forgot to actually add `Math.max` and so it was just using 1 and was never incrementing the value. cc @jeremyckahn 

Sorry I normally never code on my day off, this is what you get 😜

jeremyckahn (approved):
This looks great to me! I think this change will make the Item/Recipe inputs much easier to work with, especially on mobile. We should keep iterating here, but this is a net positive. LGTM!

@seawolff please **squash merge** this PR and ship it. I'll walk you through how to run a release on Discord!

PR #126, opened 2021-06-19
Target `main` branch instead of `master`

jeremyckahn:
This PR brings Farmhand into the present! This shouldn't affect how we do anything, it just modernizes our branch structure a bit.

I have already created the [new `main` branch](https://github.com/jeremyckahn/farmhand/tree/main) and will delete `master` once this is merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - '**'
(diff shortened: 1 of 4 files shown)


lstebner (approved):
looks good to me! i checked out the branch and grepped and don't see anything remaining. thank you for making this change!

PR #125, opened 2021-06-19
Improve GitHub actions

jeremyckahn:
I made a few tweaks that we've discussed recently. Let me know what you think of the PR template!

PR #129, opened 2021-06-21
Attempt to fix broken CI deployment

seawolff:
#### What does this PR do?
This PR assumes that https://github.com/jeremyckahn/farmhand/runs/2876457694?check_suite_focus=true is broken due to a missing branch name under the github push action

#### How can this change be manually tested?
¯\_(ツ)_/¯ we hope this works

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,3 @@
+#### What does this PR do?
+
+#### How can this change be manually tested?
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,0 @@
-#### What does this PR do?
-
-#### How can this change be manually tested?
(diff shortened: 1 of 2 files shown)


.github/workflows/run-release.yml
```diff
@@ -51,6 +51,7 @@ jobs:
         with:
           github_token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
           tags: true
+          branch: main
```
jeremyckahn, line 54: 💯

.github/pull_request_template.md
```diff
@@ -0,0 +1,3 @@
+#### What does this PR do?
+
+#### How can this change be manually tested?
```
jeremyckahn, line 3: Can we revert this file? I actually have another PR in the works to address this: https://github.com/jeremyckahn/farmhand/pull/125

jeremyckahn (changes requested):
Looking really good! Just one thing to address before merging. 😃

Issue #133, opened 2021-06-25
Feature flag system


jeremyckahn:
We need a generic feature flag system so that we can incrementally implement larger features and continually merge stable progress back to `develop`. This will let us keep feature branches more closely aligned.

PR #139, opened 2021-06-29
Add support for enabling features via query params

jeremyckahn:
### What this PR does

It extends #136 (for #133) by allowing players to enable feature flags through URL query params.

Manually add query params as described by the new documentation to the Preview environment's URL and see that they are provided as `props` to the `Farmhand` component in the React DevTools.

### Additional information

Screen Shot 2021-06-29 at 8 39 22 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ REACT_APP_ENABLE_MINING=true
 
 Would enable the `MINING` feature only for the local development environment. You can access the enabled feature flags at runtime by `import`ing the `features` Object from [`config.js`](https://github.com/jeremyckahn/farmhand/blob/develop/src/config.js). See [Adding Custom Environment Variables](https://create-react-app.dev/docs/adding-custom-environment-variables/) for more information on how to use environment variables.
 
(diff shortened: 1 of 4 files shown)


seawolff (approved):
So conceptually this LGTM.

However I have a design request. Would be great to have a flag to enable / disable a feature flag. Something like `?feature=foo&enable=true`. Of course if you did this you'd prob have to save the flags into a cookie. But I could see this being helpful for multiple feature flags etc.

lstebner:
i did expect the value to be read as well, vs just the presence of the key. i think what @seawolff said could be done the single prop still by just also seeing if the value is "true" or "false". though, i suppose the thought is all features all disabled by default until they're not longer a feature that can be toggled, so maybe just the presence is ok. i'm fine with this until we have more use cases to show we need more flexibility

lstebner (approved):
🎉

jeremyckahn:
>i suppose the thought is all features all disabled by default until they're not longer a feature that can be toggled, so maybe just the presence is ok. i'm fine with this until we have more use cases to show we need more flexibility

Yep, that's exactly the idea here. It's for enabling specific in-development features for beta testing. I'm open to revisiting this later, but for now I think this is the simplest thing that meets our needs!

PR #142, opened 2021-07-01
fix: [closes #140] avoid increasing value of items sold from the shop

### LINKED ISSUE #140: "I am Rich" achievements affect resale items
#77 introduced a bug: The value of items purchased from the Shop (such as seeds) are raised along with harvested crops. These resale value items must be kept at a constant value multiplier of `0.5` to prevent players from selling back seeds for more than they bought them for.

Relevant sections of the code:

https://github.com/jeremyckahn/farmhand/blob/f9b12b12c0231d546e0992c71070c71d2950e1cf/src/utils.js#L238-L242

https://github.com/jeremyckahn/farmhand/blob/f9b12b12c0231d546e0992c71070c71d2950e1cf/src/utils.js#L1000-L1016

https://github.com/jeremyckahn/farmhand/blob/f9b12b12c0231d546e0992c71070c71d2950e1cf/src/components/Item/Item.js#L149-L150

/cc @lstebner 

Thanks to Kairi on Discord for finding and reporting this!

lstebner:
### What this PR does

  the "I am Rich" achievement PR introduced an exploitable bug by
  affecting the price of seeds, and other items sold from the shop.
  players could buy these and instantly resell them for a steady profit.

  this PR patches that hole by checking to see if an item is sold in the
  shop before applying the adjustment to the value. this should help
  prevent this bug from popping back up in the future.

To test this change you need to achieve one of the "i am rich" achievements; so sell at least $500,000 worth of goods. Then check the resale price of seeds, fertilizer, and the scarecrow. These should all stay consistent as to what they were before, even though crops and workshop items start to get boosted.

This logic being wrapped up in this component makes it very cumbersome to test. We should probably try to make the component purely presentational and get some good unit tests written around this logic.

### Additional information

image

### DIFF
--- a/src/components/Item/Item.js
+++ b/src/components/Item/Item.js
@@ -14,7 +14,7 @@ import classNames from 'classnames'
 import FarmhandContext from '../../Farmhand.context'
 import { items } from '../../img'
 import { itemsMap } from '../../data/maps'
-import { shopItemIds } from '../../data/shop-inventory'
+import { itemIds as shopItemIds } from '../../data/shop-inventory'
 import {
   inventorySpaceRemaining,
   isItemSoldInShop,
--- a/src/components/Item/Item.js
+++ b/src/components/Item/Item.js
@@ -33,7 +33,7 @@ import './Item.sass'
 
 const noop = () => {}
 
-const shopItemIds = shopInventory.map(item => item.id)
+const shopItemIds = new Set(shopInventory.map(item => item.id))
 
 const ValueIndicator = ({ poorValue }) => (
   <Tooltip
@@ -153,7 +153,7 @@ export const Item = ({
 
   // #140 - never increase the value of items the shop sells otherwise they
   // can be bought and instantly resold for a profit to game the.. game
-  if (!shopItemIds.includes(id)) {
+  if (!shopItemIds.has(id)) {
     sellPrice *= getSalePriceMultiplier(completedAchievements)
   }
 
--- a/src/components/Item/Item.js
+++ b/src/components/Item/Item.js
@@ -14,6 +14,7 @@ import classNames from 'classnames'
 import FarmhandContext from '../../Farmhand.context'
 import { items } from '../../img'
 import { itemsMap } from '../../data/maps'
+import shopInventory from '../../data/shop-inventory'
 import {
   inventorySpaceRemaining,
   isItemSoldInShop,
@@ -31,6 +32,8 @@ import AnimatedNumber from '../AnimatedNumber'
 import './Item.sass'
 
 const noop = () => {}
+
+const shopItemIds = shopInventory.map(item => item.id)
 
 const ValueIndicator = ({ poorValue }) => (
   <Tooltip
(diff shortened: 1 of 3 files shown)


src/components/Item/Item.js
```diff
@@ -146,8 +149,13 @@ export const Item = ({
...
 
-  const sellPrice =
-    adjustedValue * getSalePriceMultiplier(completedAchievements)
+  let sellPrice = adjustedValue
+
+  // #140 - never increase the value of items the shop sells otherwise they
+  // can be bought and instantly resold for a profit to game the.. game
+  if (!shopItemIds.includes(id)) {
+    sellPrice *= getSalePriceMultiplier(completedAchievements)
+  }
```
jeremyckahn, line 158: This logic all looks sound to me! I really like the simplicity of this solution.

I'm a bit concerned about the performance cost of doing the `O(n)` lookup operation (`shopItemIds.includes(id)`), as this is a fairly hot (or at least warm) code path. While `shopItemIds` isn't a massive data set, I think it should be pretty easy to change this to be an `O(1)` operation by changing `shopItemIds` to being a `Set`. So on line 36, we could set it up to be:

```js
const shopItemIds = new Set(shopInventory.map(item => item.id))
```

And on line 156, the check would then be:

```js
if (!shopItemIds.has(id))
```

(For reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#performance)

We could do this with another data structure as well (like a plain object). What do you think @lstebner?

jeremyckahn (changes requested):
Great work! Thanks for finding a minimally-invasive solution here. I've tested this myself and it seems to work perfectly.

I have a performance concern that I detailed in an inline comment. Once that's resolved, this should be good to merge!

jeremyckahn:
> This logic being wrapped up in this component makes it very cumbersome to test. We should probably try to make the component purely presentational and get some good unit tests written around this logic.

Agreed, the logic is pretty tightly coupled right now. Sorry about that. Much of this was written when I was more focused on velocity than test coverage (#soloDevLyfe). I am very open to refactoring this to decouple the game logic from the presentation!

src/components/Item/Item.js
lstebner: a very valid concern. glad you brought it up. i like the idea of using a Set a lot, i was also thinking this could maybe be exported from the shop inventory instead of calculated in the component. how does that sound to you?

jeremyckahn: Sounds smart to me, let’s do it! 👍

lstebner: updated! @jeremyckahn

src/components/Item/Item.test.js
```diff
@@ -7,7 +7,6 @@ import { testItem } from '../../test-utils'
 import { Item } from './Item'
 
 jest.mock('../../data/maps')
-jest.mock('../../data/shop-inventory')
```
lstebner, line 10: @jeremyckahn is there a specific reason you chose to mock this? it seems like a good thing to test the real data here to me. though, the reason i looked is because the tests all fail with the mock in place for the new `itemIds` export. let me know what you think

jeremyckahn: Overall I wanted to keep game data/content reasonably decoupled from the logic for the tests. Some of the test setup is fragile and less than ideal, admittedly. I'd be fine with removing this mock if that's what's works for you here!

I don't love the the test situation for Farmhand right now, and I'm open to new ideas and approaches!

jeremyckahn (approved):
Love it! LGTM!

src/components/Item/Item.test.js
lstebner: alright, I spend lots of time thinking about unit testing so I'll take a closer look at some things and make some PRs to suggest changes

Issue #134, opened 2021-06-25
Shoveling plots


jeremyckahn:
**BLOCKED BY #131**

With the Shovel field tool selected, empty plots can be shoveled (see #131). This ticket is meant to capture the logic and mechanics of a shoveled plot.

Acceptance criteria:

  - Once a plot is changed to be in the "shoveled state," it randomly produces either an ore from #135 or nothing
  - If an ore is produced, it is added to the player's inventory and some visual indicator is shown to the player indicating as such
  - When an ore is produced by shoveling, it stays that way for 2-5 game days (chosen randomly) and then resets the day after that period
  - Shoveled plots cannot be interacted with (until they are reset per rules above)

PR #146, opened 2021-07-02
[closes #131] feat: Shovel tool

### LINKED ISSUE #131: Shovel field tool
**BLOCKED BY #133**

This field tool will be be the achievement reward for paying back the loan (#132).

Acceptance criteria:

  - The Shovel is available as a field tool once the achievement described by #132 is complete
  - There is a `SHOVEL` field mode implemented. See: https://github.com/jeremyckahn/farmhand/blob/4438983848f83b4ed937a84548574ba677849449/src/enums.js#L34-L43
  - Selecting the Shovel enables the `SHOVEL` field mode similarly to how other field tools enable their field modes
  - When the `SHOVEL` field mode is enabled, empty plots are highlighted in green similarly to how mature crops are highlighted when the Scythe is selected
  - When the player clicks on such a highlighted plot, it is changed to the "shoveled" state (this will require new art)
  - **This feature is hidden behind a feature flag**

lstebner:
### What this PR does

all of this is currently behind the MINING feature flag

- adds the Shovel tool
- refactors some code around tools data and the Toolbelt
- adds Shovel art
- adds MINE fieldMode

- if running locally simply go to the field and look for the new Shovel tool to appear in the toolbelt
- if not local, add `?enable_MINING=true` to the URL to preview

This is a work in progress, and part of a larger feature. MINE mode currently does nothing, and the shovel will be an unlocked item ultimately.

### Additional information

Screen Shot 2021-07-02 at 12 14 22 AM

shovel 3x

hooked up the art:
Screen Shot 2021-07-11 at 4 16 34 PM


lstebner:
oops, i hit request review @jeremyckahn but there's a bit more work to do here. would love feedback, but no rush and feel free to wait

src/data/tools.js
```diff
@@ -0,0 +1,30 @@
...
+  scythe: {
+    alt: 'A scythe for crop harvesting.',
+    fieldMode: HARVEST,
+  },
+  hoe: {
+    alt:
+      'A hoe for removing crops and disposing of them. Also returns replantable items to your inventory.',
+    fieldMode: CLEANUP,
+  },
+}
```
jeremyckahn, line 21: I love that this has been moved out of the component! This will prevent the Object from being redefined on each render. Great call!

jeremyckahn (commented):
This is off to a great start so far! I only have positive feedback so far. And the art looks great! I can't wait to try this new feature out! 😄

src/components/Toolbelt/Toolbelt.js
```diff
@@ -6,72 +6,63 @@ import classNames from 'classnames'
...
+
+    toolbeltIcons.push(
+      <Tooltip
+        {...{
+          key: fieldMode,
+          placement: 'top',
+          title: (
+            <>
+              <p>{alt}</p>
+              <p>(shift + {i})</p>
```
jeremyckahn, line 31: Just to make things easy for you: You can wire up this keybinding for the Shovel by updating these maps: 

https://github.com/jeremyckahn/farmhand/blob/aff10c9069251c5d01b8f35ae10bb5fdf8a448e5/src/components/Field/Field.js#L39-L43

https://github.com/jeremyckahn/farmhand/blob/aff10c9069251c5d01b8f35ae10bb5fdf8a448e5/src/Farmhand.js#L436-L438

lstebner: thanks for the pointer!

lstebner: what do you think about moving the hotkey to the 'tools' data obj as well? i felt like it might also need an 'order' for the order to display them.

lstebner: right now these strings you pointed out rely on the order of the other loop, which i could see easily breaking

jeremyckahn: Yeah that's fine with me. It'd be nice for all of this tightly-coupled logic to be in one place!

jeremyckahn:
I'm guessing you meant to post [this comment](https://github.com/jeremyckahn/farmhand/issues/131#issuecomment-873646888) in this PR rather than the Issue:

>i have this most of the way there for real this time @jeremyckahn , naming has been rough so far haha. i can go with 'SHOVELED' but after using 'MINE' as the action i went with 'MINED' and the resulting tile being called 'Rubble' but yeah.. i think we should have a quick discussion around naming! also the highlighted color i chose is more orange/yellow, but i'm fine with going with green. check it out though and let me know what you think.

Sorry to hear this has been a pain! This is a pretty low-level and substantive change, so I'm not surprised this is proving more difficult than previous PRs. I think it'll be worth the trouble, though!

Between "mine" and "shovel" semantics, I prefer consistently using "shovel," but that should be easy to sort out and iterate on once we have things minimally working. I suggest just getting it to work however you can and we can clean up and polish afterwards, and then merge once we're satisfied with it.

Regarding colors, I think we can simplify things a bit and keep them consistent with crops. The idea with the red/yellow/green highlight is to send the following messages to the players:

  - Red: "You likely don't want to perform the currently selected action on this plot" (like using the hoe on mature crops)
  - Yellow: "You can perform the currently selected action on this plot, but there's a good chance you don't mean to, so really think about it first" (like using the hoe on growing crops)
  - Green: "You probably do want to perform the currently selected action on this plot" (like using the scythe on mature crops)

Considering this, I think we'd only want to have the follow color states for field plots for the mining feature:

  - Green: For un-shoveled, empty plots

And I think that should be it. Let me know if you want to chat a bit more about this on Discord. My availability is a bit spotty today but I will respond when I can!

>i also haven't attempted any art, i think that might fit better with #134 since we might want different art for each different outcome.

Sounds good to me! I can definitely help out here. 👍

src/constants.js
```diff
@@ -60,6 +60,7 @@ export const STORM_CHANCE = 0.5
 
 export const SCARECROW_ITEM_ID = 'scarecrow'
 export const SPRINKLER_ITEM_ID = 'sprinkler'
+export const MINED_ITEM_ID = 'mined'
```
jeremyckahn, line 63: I think we want get rid of this and all of its references (like in `items.js`). The idea with these constants is to associate specific items to their game logic. This pattern has become less useful over time and may not be terribly helpful anymore. In any case, I don't think we want to use or extend it for this PR.

src/data/items.js
```diff
@@ -117,6 +118,17 @@ export const scarecrow = freeze({
...
+// TODO: move this, name it something better.
+export const mined = freeze({
+  description: 'a plot that has been mined. will be restored tomorrow',
+  hoveredPlotRangeSize: 1,
+  id: 'mined',
+  isReplantable: false,
+  name: 'Rubble',
+  type: MINED,
+  value: 0,
+})
```
jeremyckahn, line 130: Your TODO hints to as much, but just to add in case it's helpful: The shoveling feature shouldn't be conceptually concerned with any items, just plot states.

Let me know if that's too vague to be helpful! 😅

src/enums.js
```diff
@@ -36,6 +36,7 @@ export const fieldMode = enumify([
   'CLEANUP',
   'FERTILIZE',
   'HARVEST',
+  'MINE',
```
jeremyckahn, line 39: 👌

src/reducers.js
```diff
@@ -162,6 +163,13 @@ export const setWasWatered = plotContent =>
 export const resetWasWatered = plotContent =>
   setWasWateredProperty(plotContent, false)
 
+/**
+ * @param {?farmhand.plotContent} plotContent
+ * @returns {?farmhand.plotContent}
+ */
+export const resetWasMined = plotContent =>
+  plotContent && plotContent.itemId === MINED_ITEM_ID ? null : plotContent
```
jeremyckahn, line 171: Rather than using items to indicate the shoveled state, I think we want to extend this typedef: https://github.com/jeremyckahn/farmhand/blob/44bde3011e4e33be6c157e2fdae0a6f8c2696a53/src/index.js#L37-L44

I think it would make sense to add a new `farmhand.shoveledPlot` typedef that extends `farmhand.plotContent` similarly to how `farmhand.plotContent` below it does. I think we'd also have to make `farmhand.plotContent`'s `itemId` optional, because there will not be an item associated with a shoveled plot. It might be worth renaming `farmhand.plotContent` to something like `farmhand.plotState`, but don't feel that you need to do that for this PR if it's too much for now (we can smooth out those semantics in a later PR).

jeremyckahn (commented):
This is look GREAT so far! I added a few comments to hopefully help clear up some confusion. Let me know if you want to talk through any of this further!

src/reducers.js
lstebner: 🔑 thanks for this info! this makes sense to me

lstebner:
thanks for all the info about the colors @jeremyckahn . that's very helpful, and makes a lot of sense to stick with the pattern. easy change there. also the comment about naming helps. i was being a bit illustrative with my "pain" remark, the codebase is pretty easy to navigate but this is definitely a trickier PR than the others I have taken on so far; it's a good challenge. this gives me a good amount of changes to make, so i'll let you know when it's updated again or if i get stuck! 🙏

src/components/Plot/Plot.sass
```diff
@@ -1,43 +1,56 @@
 $not-allowed-color: rgba(255, 0, 0, 0.5)
 
 .Plot
+  --color-red-danger: rgba(255, 0, 0, 0.5)
+  --color-blue-water: rgba(125, 245, 255, 0.75)
+  --color-brown-fertilize: rgba(125, 56, 0, .75)
+  --color-green-ok: rgba(0, 255, 0, 0.5)
+  --color-yellow: #ff0
```
lstebner, line 8: i had to look each of these up so i decided to name them all. i have really grown to like css native variables, but we can use sassy ones instead if you prefer @jeremyckahn

src/components/Toolbelt/Toolbelt.test.js
```diff
@@ -1,39 +1,82 @@
 import React from 'react'
-import { shallow } from 'enzyme'
+import { mount, render } from 'enzyme'
```
lstebner, line 2: i will update this spec to use react-testing-lib once #154 is merged.

lstebner:
everything seems to be working though i'm not feeling like it's the cleanest code. looking forward to your review @jeremyckahn ! here's a little video of it in action https://www.dropbox.com/s/maeqqzpqo58l7s5/farmhand-shovelin.mov?dl=0

src/reducers.js
lstebner: made these updates

jeremyckahn (changes requested):
This is coming along really well! Thanks for sticking with this and for the video. It's working just like how we talked about! 😃

I added a few comments inline about how we might make the changes here a little more scalable. Also I noticed two issues that need to be resolved before this can be merged:

1. The CI tests appear to be currently failing: https://github.com/jeremyckahn/farmhand/runs/3036590500?check_suite_focus=true
2. One of my local save files crashes immediately under this branch. Here's the save file: https://gist.github.com/jeremyckahn/4686d38bec217b97ab3f7caf37ea3cc5. You should just need to import this file to see the crash.

src/components/Plot/Plot.sass
lstebner: 🙏  this is not exactly the most powerful example of them, but maybe we can work on it together in the future. i couldn't figure out how to do a `:root` style block in sass (i typically use these in vanilla CSS with CSS modules these days, or SCSS). and the way they are able to be redefined is actually way more powerful than SASS allows.

src/components/Plot/Plot.js
```diff
@@ -132,7 +134,7 @@ export const Plot = ({
         placement: 'top',
         title: (
           <>
-            <Typography>{item.name}</Typography>
+            <Typography>{item ? item.name : 'shoveled'}</Typography>
```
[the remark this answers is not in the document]
lstebner: i think this logic can be improved in general with the update from your other comment to add something actually signaling this plot was shoveled. i do agree it's not great as the general fallback! 

re: strings, no problem moving this one, but what do you think about adopting something like [i18next](https://www.i18next.com/) in the near future? we don't need to actually localize strings (yet), but we would be prepared to do so, while also keeping them separated. this would also give us an interface for strings that use interpolation, or dynamic count values (the `count` arg is particularly powerful). happy to share some examples if you've never worked with i18next before!

src/components/Toolbelt/Toolbelt.test.js
```diff
@@ -1,39 +1,82 @@
...
+const noop = () => {}
 
-test('renders shop inventory', () => {
-  expect(component).toHaveLength(1)
-})
+// https://stackoverflow.com/questions/58070996/how-to-fix-the-warning-uselayouteffect-does-nothing-on-the-server
+jest.mock('react', () => ({
+  ...jest.requireActual('react'),
+  useLayoutEffect: jest.requireActual('react').useEffect,
+}))
```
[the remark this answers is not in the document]
lstebner: i think this will all go away with testing library. pretty sure this is weirdness with enzyme and material ui :/ (or maybe it was from the version mismatches.. either way, expecting it to go away)

src/event-handlers.js
```diff
@@ -27,13 +28,14 @@ const {
...
   HARVEST,
+  MINE,
   PLANT,
   SET_SCARECROW,
   SET_SPRINKLER,
   WATER,
 } = fieldMode
 
-const toolbeltFieldModes = [CLEANUP, HARVEST, WATER]
+const toolbeltFieldModes = [CLEANUP, HARVEST, WATER, MINE] // todo: where can this data come from that is more dynamic? data/tools.js?
```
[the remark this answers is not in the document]
lstebner: at the time i was thinking about if it needed to consider the `features` or not, but i think this is kind of out of place in general too. maybe it fits in `enums.js`? we have `fieldMode` there already which this is a subset of, or maybe it fits with the `Toolbelt` component better? 

we're also manually controlling the calls to `handleFieldModeSelect`, so while this feels like it's an "allowlist" type of array, i think just checking against `fieldMode` from enums could do the same job, and not require extra maintenance. 

we don't really have to do anything with this for this PR, but let me know if any of that resonates with you (or doesn't make sense 😅 )

src/utils.js
```diff
@@ -266,7 +266,8 @@ export const getPlotContentFromItemId = itemId => ({
  * @param {farmhand.plotContent} plotContent
  * @returns {string}
  */
-export const getPlotContentType = ({ itemId }) => itemsMap[itemId].type
+export const getPlotContentType = ({ itemId }) =>
+  itemId ? itemsMap[itemId].type : null
```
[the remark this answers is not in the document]
lstebner: good call, sorry i am not used to using the doc stuff!

src/components/Plot/Plot.js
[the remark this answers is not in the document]
jeremyckahn: >what do you think about adopting something like i18next in the near future?

I'm into it! I've used i18next lightly on a few projects and I think it would be good to use as a matter of best practice at the very least. That said, I don't feel a rush to integrate it, as it doesn't solve a problem that we currently have. I'll leave the timeline and strategy up to you for doing this, but I'll support you with whatever you need to get it done! 👍

src/event-handlers.js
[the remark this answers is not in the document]
jeremyckahn: I see what you're saying. I don't think it belongs in `enums.js` because it's a list and not an enum. However, I think it would make sense to move this to `constants.js`!

>we're also manually controlling the calls to handleFieldModeSelect, so while this feels like it's an "allowlist" type of array, i think just checking against fieldMode from enums could do the same job, and not require extra maintenance.

Yeah this is effectively an allowlist. I don't think we should reuse the `fieldModes` enum for this because `toolbeltFieldModes` is intentionally a subset of those values and we don't want to match with anything beyond it. 

Upon looking at this closer, `toolbeltFieldModes` should be a faster data structure (like an Object or a Set). Here's the only place it's used: https://github.com/jeremyckahn/farmhand/blob/2a4966b1c98fe62bf4c94d0be27e7757279328c2/src/event-handlers.js#L127-L135

I don't know why I went with an array for this initially (it's been a few years since I wrote this), but this is as good a time as any to improve it since this is the first time it has been touched in a while!

-------------------------

**tl;dr**: I'd say this is fine to leave as-is for this PR, but if you're interested in improving this then I'd support converting this to a  faster data structure and moving it to `constants.js`. 🙂

src/components/Plot/Plot.js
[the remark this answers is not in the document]
lstebner: makes sense to me!

src/event-handlers.js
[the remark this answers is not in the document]
lstebner: added it as a Set to constants!

lstebner:
> This is coming along really well! Thanks for sticking with this and for the video. It's working just like how we talked about! 😃
> 
> I added a few comments inline about how we might make the changes here a little more scalable. Also I noticed two issues that need to be resolved before this can be merged:
> 
> 1. The CI tests appear to be currently failing: https://github.com/jeremyckahn/farmhand/runs/3036590500?check_suite_focus=true
> 2. One of my local save files crashes immediately under this branch. Here's the save file: https://gist.github.com/jeremyckahn/4686d38bec217b97ab3f7caf37ea3cc5. You should just need to import this file to see the crash.

everything should be resolved, the code is much more resilient now! CI is happy, your save seems fine too, but let me know if it's still broken for you when you give it a try. i also moved the other strings that were in tools.js now to the strings file, and added a few more unit tests around the reducers

jeremyckahn:
FYI @lstebner, I just pushed https://github.com/jeremyckahn/farmhand/commit/68f48f60d287a17806c22d3ca3ac271a0da23b42, which will make the range selector work for shoveling! 

Kapture 2021-07-12 at 08 10 15

src/components/Toolbelt/Toolbelt.js
```diff
@@ -6,79 +6,61 @@ import classNames from 'classnames'
...
-                className: `square ${toolImageId}`,
-                src: pixel,
-                style: { backgroundImage: `url(${tools[toolImageId]}` },
+                className: classNames({
+                  selected: fieldMode === currentFieldMode,
+                }),
+                color: 'primary',
+                onClick: () => handleFieldModeSelect(fieldMode),
+                variant: fieldMode === currentFieldMode ? 'contained' : 'text',
+                'data-tool-id': id,
```
jeremyckahn, line 44: What's this used for? I don't see it referenced anywhere else. Should it be removed?

src/components/Toolbelt/Toolbelt.test.js
```diff
@@ -49,5 +56,14 @@ describe('<ToolBelt />', () => {
         true
       )
     })
+
+    test('marks the shovel selected for field mode MINE', async () => {
```
jeremyckahn, line 60: This function is `async`, but it has no `await`s. Should `async` be removed?

src/constants.js
```diff
@@ -176,3 +176,10 @@ export const MAX_LATEST_PEER_MESSAGES = 30
 
 export const LEFT = 'left'
 export const RIGHT = 'right'
+
+export const TOOLBELT_FIELD_MODES = new Set([
+  fieldMode.CLEANUP,
+  fieldMode.HARVEST,
+  fieldMode.WATER,
+  fieldMode.MINE,
+])
```
jeremyckahn, line 185: 💯

src/reducers.test.js
```diff
@@ -3310,3 +3310,36 @@ describe('prependPendingPeerMessage', () => {
...
+    const plotContents = fn.resetWasShoveled({ wasShoveledToday: true })
+
+    expect(plotContents).toEqual(null)
+  })
+
+  test('will return the plot contents if it is anything else', () => {
+    const plotContents = fn.resetWasShoveled('sprinkler')
+
+    expect(plotContents).toEqual('sprinkler')
+  })
```
jeremyckahn, line 3344: Awesome work with these tests!!

src/utils.js
```diff
@@ -264,9 +264,10 @@ export const getPlotContentFromItemId = itemId => ({
 
 /**
  * @param {farmhand.plotContent} plotContent
- * @returns {string}
+ * @returns ?{string}
```
jeremyckahn, line 267: The `?` needs to go inside of the braces here, I think. So, `{?string}`

jeremyckahn (changes requested):
Awesome work!! We're nearly there with this. There's just a few minor remaining questions/issues I'd like to get resolved before this is merged.

src/components/Toolbelt/Toolbelt.js
lstebner: ahh, yes, i thought i was going to use this for the unit tests at one point before we switched to react testing lib

src/components/Toolbelt/Toolbelt.test.js
lstebner: yup! also came from the testing adventure. user input with testing lib requires async/await behavior, but i should have removed this with the removal of the click action in the test

src/reducers.test.js
lstebner: thank you! 🙏

src/utils.js
lstebner: oops!

lstebner:
thanks for the early morning review and adding the range logic @jeremyckahn ! i pushed updates for all the things you noticed

jeremyckahn (approved):
Everything looks great to me! Thanks for being receptive for all my comments and sticking with this. We’re getting close to shipping the shoveling feature! 😃

LGTM!

PR #150, opened 2021-07-03
Update developer note on Home screen

jeremyckahn:
### What this PR does

It updates the developer note on the Home screen to be more reflective of the state of the project and the people building it. It also removes the payment links.

Read and proofread the copy changes.

### DIFF
--- a/src/components/Home/Home.js
+++ b/src/components/Home/Home.js
@@ -60,7 +60,7 @@ const Home = ({
             linkTarget: '_blank',
             className: 'markdown',
             source: `
-Hi, you're playing **Farmhand**! This is an open source game project created and lead by [Jeremy Kahn](https://github.com/jeremyckahn). The project has evolved over time and is now developed with the support of [a community of contributors](https://github.com/jeremyckahn/farmhand/blob/develop/CONTRIBUTORS.md).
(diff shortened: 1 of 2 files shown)


src/components/Home/Home.js
```diff
@@ -60,13 +60,13 @@ const Home = ({
             linkTarget: '_blank',
             className: 'markdown',
             source: `
-Hi! You're playing **Farmhand**, an open source game being developed by [Jeremy Kahn](https://github.com/jeremyckahn) and [a community of contributors](https://github.com/jeremyckahn/farmhand/blob/develop/CONTRIBUTORS.md).
+Hi, you're playing **Farmhand**! This is an open source game project created and lead by [Jeremy Kahn](https://github.com/jeremyckahn). The project has evolved over time and is now developed with the support of [a community of contributors](https://github.com/jeremyckahn/farmhand/blob/develop/CONTRIBUTORS.md).
```
lstebner, line 63: i think you want 'led by' here

jeremyckahn: I actually had to look this up, but I believe "lead" is correct here because this is not past tense: https://www.grammarly.com/blog/led-lead/

> The past tense of the verb lead is led, not lead.

jeremyckahn: On second thought this is a sort of unclear and confusing: https://newsroom.unl.edu/announce/snr/3776/21317

"Led" feels a little more correct here, so I'll switch to that. Good call!

PR #153, opened 2021-07-08
Add shoveled plot state art

jeremyckahn:
### What this PR does

It adds art for #134 (also relates to #131).

Screen Shot 2021-07-08 at 8 29 38 AM

You'll have to modify the code a bit to get this to show as it is not yet hooked up to anything. You can also just look at the screenshot above.

This art is 32x32, which is notably different than the 24x24 dimensions of crop art. I chose this because the fertilized plot state art is 32x32. Perhaps this should all be sorted out and made consistent later, but for now I think it's best to match the fertilized plot state style for consistency.

### Additional information

This is simply a preemptive PR for #134. The actual feature implementation should happen in a later PR, after this is merged to `main`.


seawolff (approved):
Not much to see here. LGTM 🚢

lstebner:
yeah i'm cool with this. we can always tweak it if we feel like it, thanks for creating this!

lstebner (approved):

PR #154, opened 2021-07-10
chore: adding react-testing-library

lstebner:
### What this PR does

  - adds react-testing-library dependency
  - updates Toolbelt tests to use react-test-library instead of enzyme
  
CI passing is all the indication we really need, but you can check this branch out, run `npm i` and then run the `Toolbelt` tests to make sure they are passing with the changes.

We now have a mix of testing approaches, but I think it's ok. We should probably put in an effort to convert to testing library though before we have too many more specs. 

### Additional information

I found this article helpful while converting the test 👉 https://www.robinwieruch.de/react-testing-library


jeremyckahn:
Github is refusing to render the full diff for `package-lock.json`, so here it is as a gist: https://gist.github.com/jeremyckahn/dee19f7de5efea5b1165773c860a8621

jeremyckahn (changes requested):
Very exciting! I'm on board with modernizing our testing infrastructure (it's long overdue), I just want to understand the implications a bit better first. Also I was wondering if it would make sense to switch to a functional component for the test setup.

src/components/Toolbelt/Toolbelt.test.js
```diff
@@ -1,39 +1,61 @@
...
-test('renders shop inventory', () => {
-  expect(component).toHaveLength(1)
-})
-
-describe('tool selection', () => {
-  test('selects no tools by default', () => {
-    expect(component.find('.selected')).toHaveLength(0)
+let buttons = []
+
+class ToolbeltTest extends React.Component {
```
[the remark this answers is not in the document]
lstebner: this is not going to be the response you were expecting i don't think... but i actually did this just because i wanted to be able to test clicking the button and having it become selected. you were using the noop for the click callback and this was the only way i could think to really hook it up, but now that you called it out i thought some more and the unit test here is simply that it marks the expected button selected based on the fieldMode prop. i added new tests to do that, i also added a "visually hidden" class to put some text in there that isn't actually seen. this should be helpful to screen readers as well, though we have a lot more work to do in that regard.

src/components/Toolbelt/Toolbelt.sass
```diff
@@ -1,4 +1,12 @@
 .Toolbelt
+  .visually_hidden:not(:focus):not(:active)
+    clip: rect(0 0 0 0)
+    clip-path: inset(50%)
+    height: 1px
+    overflow: hidden
+    position: absolute
+    white-space: nowrap
+    width: 1px
```
lstebner, line 9: this can probably live in more of a "base" stylesheet if there is one, also i didn't make these styles up so here's a bit more about them https://css-tricks.com/inclusively-hidden/

src/components/Toolbelt/Toolbelt.test.js
```diff
@@ -1,39 +1,53 @@
...
-      component.setProps({ fieldMode: fieldMode.WATER })
+  describe('tool selection', () => {
+    test('there are no selected tools by default', () => {
+      render(<Toolbelt fieldMode={fieldMode.OBSERVE} />)
+      expect(getSelectedButton()).toBeUndefined()
+    })
+
+    test('marks the watering can selected for field mode WATER', async () => {
+      render(<Toolbelt fieldMode={fieldMode.WATER} />)
+      const button = screen.getByText('select watering can')
```
lstebner, line 28: this variable name ended up being confusing, i will probably rename it in the morning

src/components/Toolbelt/Toolbelt.sass
jeremyckahn: I love this! This seems generally useful so let's move it to (what is effectively) the base stylesheet: https://github.com/jeremyckahn/farmhand/blob/develop/src/Farmhand.sass

src/components/Toolbelt/Toolbelt.js
```diff
@@ -22,19 +24,22 @@ export const Toolbelt = ({
           alt: 'A watering can for hydrating plants.',
           fieldMode: WATER,
           toolImageId: 'watering-can',
+          hiddenText: 'select watering can',
```
jeremyckahn, line 27: Thank you for prioritizing accessibility! Farmhand certainly has a lot of ground to cover in this regard, and this is a great way to get us on the right track.

Out of curiosity, are these `hiddenText` strings all lowercased intentionally? It's fine either way since they will not be visually read, I just want to know if this was done as a matter of best practice or otherwise.

src/components/Toolbelt/Toolbelt.test.js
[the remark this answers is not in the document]
jeremyckahn: The new approach is WAY nicer, thank you for taking another stab at this! I think it's simple, effective, and scalable. Nice work!

jeremyckahn (commented):
This is looking even better! I'm going to make and push a small change and then hand this back to you @lstebner. 🙂

src/components/Toolbelt/Toolbelt.sass
jeremyckahn: This is done in https://github.com/jeremyckahn/farmhand/commit/fd24e6fdd9d600683fb31c9f92f5df6415a68e46!

jeremyckahn (approved):
This is so helpful! LGTM — thanks for getting us started with React Testing Library! 🙌

src/components/Toolbelt/Toolbelt.js
lstebner: glad you brought this up actually, i just kinda threw them down here thinking about the bigger picture of this PR but we should think about this before merging. I don't actually know if casing matters (other than you definitely don't want all uppercase), so we probably want proper grammar here. this is what a screen reader would say out loud to tell the person what action will be taken, maybe we should do something more like "Select the watering can to water crops", and similarly for the others. that would be similar to what the tooltip does for sighted users. what do you think @jeremyckahn ?

lstebner: i went ahead and updated so you can just comment on them instead, hope that's ok!

src/components/Toolbelt/Toolbelt.test.js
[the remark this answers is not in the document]
lstebner: 🙌

src/components/Toolbelt/Toolbelt.js
jeremyckahn: Looks great, thanks for making the changes! 😃

PR #158, opened 2021-07-13
[closes #132] feat: Gold Digger Achievement and Shovel unlock

### LINKED ISSUE #132: "Gold Digger" achievement
**BLOCKED BY #133**

Challenge: Pay back the initial loan the player starts out with.

Reward: The Shovel is unlocked. (#131)

**This feature must hidden behind a feature flag.**

lstebner:
### What this PR does

  - adds 'gold-digger' achievement
  - adds logic to Toolbelt to filter out shovel before it is unlocked
  - minor syntax modifications
  - relevant unit test updates

- start a brand new game
- go to the field page and observe the 3 available tools
- open the finance screen and pay off the entire loan
- the achievement should trigger, and you should now be able to use the shovel in the field

😎 

### Additional information

Screen Shot 2021-07-12 at 9 34 34 PM
Screen Shot 2021-07-14 at 9 18 31 PM
Screen Shot 2021-07-14 at 9 18 36 PM
Screen Shot 2021-07-12 at 9 34 57 PM

### DIFF
--- a/src/Farmhand.js
+++ b/src/Farmhand.js
@@ -368,6 +368,7 @@ export default class Farmhand extends Component {
       sendPeerMetadata: null,
       selectedCowId: '',
       selectedItemId: '',
+      shovelUnlocked: false,
       fieldMode: OBSERVE,
       priceCrashes: {},
       priceSurges: {},
--- a/src/Farmhand.js
+++ b/src/Farmhand.js
@@ -368,7 +368,6 @@ export default class Farmhand extends Component {
       sendPeerMetadata: null,
       selectedCowId: '',
       selectedItemId: '',
-      shovelUnlocked: false,
       fieldMode: OBSERVE,
       priceCrashes: {},
       priceSurges: {},
--- a/src/components/Toolbelt/Toolbelt.js
+++ b/src/components/Toolbelt/Toolbelt.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import Button from '@material-ui/core/Button'
 import Tooltip from '@material-ui/core/Tooltip'
-import { func, string } from 'prop-types'
+import PropTypes from 'prop-types'
 import classNames from 'classnames'
 
 import FarmhandContext from '../../Farmhand.context'
@@ -9,14 +9,17 @@ import './Toolbelt.sass'
 import { tools as toolImages, pixel } from '../../img'
 import toolsData from '../../data/tools'
 
-const tools = Object.values(toolsData).sort(t => t.order)
-
 const noop = () => {}
 
 export const Toolbelt = ({
   fieldMode: currentFieldMode,
   handleFieldModeSelect,
+  shovelUnlocked,
 }) => {
+  const tools = Object.values(toolsData)
+    .filter(t => shovelUnlocked || t.id !== 'shovel')
+    .sort(t => t.order)
+
   return (
     <div className="Toolbelt">
       <div className="button-array">
@@ -62,8 +65,9 @@ export const Toolbelt = ({
 }
 
 Toolbelt.propTypes = {
-  fieldMode: string.isRequired,
-  handleFieldModeSelect: func,
+  fieldMode: PropTypes.string.isRequired,
+  handleFieldModeSelect: PropTypes.func,
+  shovelUnlocked: PropTypes.bool,
 }
 
 Toolbelt.defaultProps = {
@@ -74,7 +78,12 @@ export default function Consumer(props) {
   return (
     <FarmhandContext.Consumer>
       {({ gameState, handlers }) => (
-        <Toolbelt {...{ ...gameState, ...handlers, ...props }} />
+        <Toolbelt
+          fieldMode={gameState.fieldMode}
+          shovelUnlocked={gameState.shovelUnlocked}
+          handleFieldModeSelect={handlers.handleFieldModeSelect}
+          {...props}
+        />
       )}
     </FarmhandContext.Consumer>
   )
--- a/src/components/Toolbelt/Toolbelt.js
+++ b/src/components/Toolbelt/Toolbelt.js
@@ -1,24 +1,32 @@
 import React from 'react'
-import Button from '@material-ui/core/Button'
-import Tooltip from '@material-ui/core/Tooltip'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
 
+import Button from '@material-ui/core/Button'
+import Tooltip from '@material-ui/core/Tooltip'
+
+import { memoize } from '../../utils'
+
 import FarmhandContext from '../../Farmhand.context'
-import './Toolbelt.sass'
-import { tools as toolImages, pixel } from '../../img'
 import toolsData from '../../data/tools'
 
+import { tools as toolImages, pixel } from '../../img'
+
+import './Toolbelt.sass'
+
 const noop = () => {}
+const getTools = memoize(shovelUnlocked => {
+  return Object.values(toolsData)
+    .filter(t => shovelUnlocked || t.id !== 'shovel')
+    .sort(t => t.order)
+})
 
 export const Toolbelt = ({
   fieldMode: currentFieldMode,
   handleFieldModeSelect,
   shovelUnlocked,
 }) => {
-  const tools = Object.values(toolsData)
-    .filter(t => shovelUnlocked || t.id !== 'shovel')
-    .sort(t => t.order)
+  const tools = getTools(shovelUnlocked)
 
   return (
     <div className="Toolbelt">
@@ -78,12 +86,7 @@ export default function Consumer(props) {
   return (
     <FarmhandContext.Consumer>
       {({ gameState, handlers }) => (
-        <Toolbelt
-          fieldMode={gameState.fieldMode}
-          shovelUnlocked={gameState.shovelUnlocked}
-          handleFieldModeSelect={handlers.handleFieldModeSelect}
-          {...props}
-        />
+        <Toolbelt {...{ ...gameState, ...handlers, ...props }} />
       )}
     </FarmhandContext.Consumer>
   )
--- a/src/components/Toolbelt/Toolbelt.js
+++ b/src/components/Toolbelt/Toolbelt.js
@@ -24,9 +24,9 @@ const getTools = memoize(shovelUnlocked => {
 export const Toolbelt = ({
   fieldMode: currentFieldMode,
   handleFieldModeSelect,
-  shovelUnlocked,
+  completedAchievements,
 }) => {
-  const tools = getTools(shovelUnlocked)
+  const tools = getTools(completedAchievements['gold-digger'])
 
   return (
     <div className="Toolbelt">
@@ -75,11 +75,12 @@ export const Toolbelt = ({
 Toolbelt.propTypes = {
   fieldMode: PropTypes.string.isRequired,
   handleFieldModeSelect: PropTypes.func,
-  shovelUnlocked: PropTypes.bool,
+  completedAchievements: PropTypes.object,
 }
 
 Toolbelt.defaultProps = {
   handleFieldModeSelect: noop,
+  completedAchievements: {},
 }
 
 export default function Consumer(props) {
--- a/src/components/Toolbelt/Toolbelt.test.js
+++ b/src/components/Toolbelt/Toolbelt.test.js
@@ -24,8 +24,13 @@ describe('<ToolBelt />', () => {
     expect(screen.getAllByRole('button')).toHaveLength(3)
   })
 
-  test('renders the shovel tool once unlocked', () => {
-    render(<Toolbelt fieldMode={fieldMode.OBSERVE} shovelUnlocked={true} />)
+  test('renders the shovel tool once gold digger achievement is completed', () => {
+    render(
+      <Toolbelt
+        fieldMode={fieldMode.OBSERVE}
+        completedAchievements={{ 'gold-digger': true }}
+      />
+    )
     expect(screen.getAllByRole('button')).toHaveLength(4)
   })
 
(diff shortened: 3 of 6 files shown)


src/components/Toolbelt/Toolbelt.js
```diff
@@ -1,22 +1,25 @@
 import React from 'react'
 import Button from '@material-ui/core/Button'
 import Tooltip from '@material-ui/core/Tooltip'
-import { func, string } from 'prop-types'
+import PropTypes from 'prop-types'
```
lstebner, line 4: this is the standard i am used to and it's a bit easier to maintain, but i am open to whatever style is preferred in the codebase

src/components/Toolbelt/Toolbelt.js
```diff
@@ -74,7 +78,12 @@ export default function Consumer(props) {
   return (
     <FarmhandContext.Consumer>
       {({ gameState, handlers }) => (
-        <Toolbelt {...{ ...gameState, ...handlers, ...props }} />
+        <Toolbelt
+          fieldMode={gameState.fieldMode}
+          shovelUnlocked={gameState.shovelUnlocked}
+          handleFieldModeSelect={handlers.handleFieldModeSelect}
```
lstebner, line 84: seeing the entire gameState and handlers destructured was a bit confusing for me, so i would much rather be explicit here. again though, open to whatever style the codebase prefers

src/Farmhand.js
```diff
@@ -368,6 +368,7 @@ export default class Farmhand extends Component {
       sendPeerMetadata: null,
       selectedCowId: '',
       selectedItemId: '',
+      shovelUnlocked: false,
```
jeremyckahn, line 371: Instead of adding a new state value, we can use `completedAchievements['gold-digger']` to determine whether the shovel is available similarly to how you did it here: https://github.com/jeremyckahn/farmhand/blob/cb4734a9727f2268b9cd3cbc9a1ac70a9d6d8510/src/utils.js#L1008-L1014

Though it's more verbose, I think it's a clean pattern and have used it elsewhere myself: https://github.com/jeremyckahn/farmhand/blob/cb4734a9727f2268b9cd3cbc9a1ac70a9d6d8510/src/components/Item/Item.js#L209-L214

Do you think it would make sense to drop this new state property and use the new `completedAchievements['gold-digger']` flag instead where necessary?

src/components/Toolbelt/Toolbelt.js
jeremyckahn: I personally prefer the destructured format, but not enough to impose it on others. I'm okay with a manageable mix of code styles in the codebase, so this is fine with me!

src/components/Toolbelt/Toolbelt.js
```diff
@@ -1,22 +1,25 @@
...
 const noop = () => {}
 
 export const Toolbelt = ({
   fieldMode: currentFieldMode,
   handleFieldModeSelect,
+  shovelUnlocked,
 }) => {
+  const tools = Object.values(toolsData)
+    .filter(t => shovelUnlocked || t.id !== 'shovel')
+    .sort(t => t.order)
```
jeremyckahn, line 21: I'd say this is a warm code path, so I'd like to avoid repeating work where it's reasonable to do so. That said, this seems like a good candidate for memoization. Take a look at this util: https://github.com/jeremyckahn/farmhand/blob/cb4734a9727f2268b9cd3cbc9a1ac70a9d6d8510/src/utils.js#L123-L127

It should be pretty easy to move this calculation out of the render function and memoize it for faster re-execution. Let me know what you think of this idea!

src/components/Toolbelt/Toolbelt.js
jeremyckahn: In this case I feel pretty strongly about reverting this back to the previous style, it's a pattern that can be easily copy/pasted across the component files and generally not have to be though about too much. I'm actually surprised that you had to make any changes to the `default` export at all here. Was that a stylistic choice, or was it necessary to get things to work?

src/components/Toolbelt/Toolbelt.test.js
```diff
@@ -19,8 +19,13 @@ describe('<ToolBelt />', () => {
...
   }
 
-  test('renders a button for each tool', () => {
+  test('renders a button for each of the default tools', () => {
     render(<Toolbelt fieldMode={fieldMode.OBSERVE} />)
+    expect(screen.getAllByRole('button')).toHaveLength(3)
+  })
+
+  test('renders the shovel tool once unlocked', () => {
+    render(<Toolbelt fieldMode={fieldMode.OBSERVE} shovelUnlocked={true} />)
```
jeremyckahn, line 28: These new tests are so much simpler! 😃

src/constants.js
```diff
@@ -149,6 +149,7 @@ export const PERSISTED_STATE_KEYS = [
   'recordSingleDayProfit',
   'revenue',
   'showNotifications',
+  'shovelUnlocked',
```
jeremyckahn, line 152: Note that if we remove the new `shovelUnlocked` state property, we can remove this too!

src/data/achievements.js
```diff
@@ -275,6 +275,18 @@ const achievements = [
     condition: state => state.revenue >= goal,
     reward: state => state,
   }))(),
+
+  (() => ({
+    id: 'gold-digger',
+    name: 'Gold Digger',
+    description: `Pay Off Your Loan.`,
+    rewardDescription: `A Shovel to use to dig for ores in your field`,
```
jeremyckahn, line 283: This content is all the right format, but subjectively feels imbalanced compared to how the other achievements look. Here's a comparison: 

Screen Shot 2021-07-13 at 8 40 33 AM

I feel that most of the content here should be in the `description` and that `rewardDescription` should be as brief as possible. Here's an example of what I'm getting at: https://github.com/jeremyckahn/farmhand/blob/cb4734a9727f2268b9cd3cbc9a1ac70a9d6d8510/src/data/achievements.js#L70-L83

This makes for a more "bottom heavy" text display, which is more consistent with the other achievements, if nothing else. Do you think this could be reworded a bit to fit in with the other achievements a bit?

jeremyckahn (changes requested):
This is looking great so far!! I've tested this out and it's working great. There's a few places I think we can simplify the implementation a bit and have called them out inline!

src/Farmhand.js
lstebner: will those persist? i.e. if the player takes out a new loan, will the game remember they had completed that achievement without re-evaluating the loan balance? i _think_ this achievement is a bit unique in how it unlocks

src/components/Toolbelt/Toolbelt.js
lstebner: one of the issues it that the result of the filtering could change on any render. i originally had it outside of render, and it would not show the shovel after unlocking until completely refreshing the app, so i moved it into here. i'm sure we can come up with something, but wanted to say that in case it triggers any more thoughts for you. can chat in discord to sort this out

src/components/Toolbelt/Toolbelt.test.js
lstebner: 💯

src/data/achievements.js
lstebner: ah, yep. i think i did this same sort of "backwards" thing when i first did the i am rich implementation too 😄 . will update

src/components/Toolbelt/Toolbelt.js
lstebner: stylistic. your reasoning is sorta exactly why i prefer this, "not have to be thought about" is a bit of a trigger for me. you specifically also know the game code very intimately, but i did not know what props were coming from which object, and this multiple destructuring syntax you're using feels like it's trying to avoid passing things by reference accidentally, but it makes me very confused as to what props are coming into a component. i would be surprised if there aren't also performance implications from doing this vs passing only what is needed. so, my main reasons for wanting to change this are that it's not explicit the way it is now, and it's non-standard for the react community. 

there's actually a much bigger discussion here though because you use the destructure syntax a lot, and i have never seen it used so often. i would really prefer all JSX pass in specific props, it's supposed to look and work like HTML more than javascript from what i understand. you can choose to nip this discussion in the bud though of course as the ultimate code owner here.

one last thing i do want to point out though is what i called the "multiple destructuring".

this is the code that currently exists:
```<Toolbelt {...{ ...gameState, ...handlers, ...props }} />```

but this is what i would expect if the component actually wanted all the items in each of these objects:
```<Toolbelt { ...gameState} {...handlers} {...props } />```

the extra hash you create feels like a performance issue for sure, and that gameState object is huge, i just can't get over that. 😅

**Comment by lstebner** [comment]
responded to all your comments @jeremyckahn. happy to chat through this on discord later today so we can resolve these things!

File: src/Farmhand.js (hunk 3 above)
Comment (jeremyckahn) [reply src/Farmhand.js:371]: They will, yes. `completedAchievements` properties are stored along with the rest of the persisted state object and don't change after being set to `true`. Additionally, achievements are only evaluated and updated when they have not yet been met: https://github.com/jeremyckahn/farmhand/blob/cb4734a9727f2268b9cd3cbc9a1ac70a9d6d8510/src/reducers.js#L1855-L1877

File: src/components/Toolbelt/Toolbelt.js (hunk 4 above)
Comment (jeremyckahn) [reply src/components/Toolbelt/Toolbelt.js:21]: Hm, that's odd. It looks like the only state this depends on is `shovelUnlocked` property. I would expect a function outside of this component that looks something like this might do it:

```js
const getTools = memoize((shovelUnlocked) => {
  return Object.values(toolsData)
    .filter(t => shovelUnlocked || t.id !== 'shovel')
    .sort(t => t.order)
})
```

And then it would be called here with `getTools(shovelUnlocked)`. You may want to try this without the `memoize` wrapper to see if the base logic works first. Feel free to ping me on DIscord and I'll respond as I can!

src/components/Toolbelt/Toolbelt.js
jeremyckahn: There's definitely a bigger discussion to be had here! This is a part of the project's architecture that I definitely went my own way with, and I can see how that would create challenges for others. For what it's worth, I never actually thought anyone else would ever be working on this with me, and if I did I might have made some different choices. 😅

In any case, This pattern effectively gives us a sort of "connected component" in the Redux sense, but with pure React APIs. However, there are drawbacks. There is a performance penalty with this spread-ed object approach, but it hasn't manifested into anything significant yet. The bigger issue is the nonstandard nature of it, as you pointed out. I am open to rethinking this and even updating the codebase wholesale to a different, more standard pattern. That said, I'd like to avoid introducing it for one special case in a feature PR like this. If you'd like to change this aspect of the project's architecture, please open a ticket and we'll figure something out. I'm pretty open-minded with making the code more accessible to others, I just want to keep individual PRs as focused as possible. 🙂

>you use the destructure syntax a lot, and i have never seen it used so often. i would really prefer all JSX pass in specific props, it's supposed to look and work like HTML more than javascript from what i understand.

I have the opposite preference, but it is just a preference. I'm fine with you using the HTML-ish syntax where you'd like if that makes things easier for you!

>this is the code that currently exists:
>`<Toolbelt {...{ ...gameState, ...handlers, ...props }} />`
>
>but this is what i would expect if the component actually wanted all the items in each of these objects:
>`<Toolbelt { ...gameState} {...handlers} {...props } />`

This is a really interesting point. It's certainly possible that I got this wrong, and this would be an easy fix if so. We should definitely experiment to see if this can be improved! That said, I'd like to save that for another PR as well so that we can keep changes isolated and easily revert-able if ever necessary.

jeremyckahn (commented):
Responded to your comments!

src/components/Toolbelt/Toolbelt.js
lstebner: yup this works! thanks for the help, i see what i had wrong now too

src/components/Toolbelt/Toolbelt.js
lstebner: > I never actually thought anyone else would ever be working on this with me, and if I did I might have made some different choices.

ha! this made me laugh, because i know exactly what you mean here

>  I am open to rethinking this and even updating the codebase wholesale to a different, more standard pattern. That said, I'd like to avoid introducing it for one special case in a feature PR like this.

this makes a lot of sense to me. i appreciate the discussion here and i'm going to revert these stylistic changes. we can have a discussion about the tradeoffs as things grow. i'm glad you have the performance concern in mind as well most importantly, and the code is consistent now which is often more valuable than anything to be honest.

src/Farmhand.js
lstebner: ahh, great! that's perfect then. i like not adding to global state or save data 😄

lstebner:
another round of updates available for your morning coffee time haha @jeremyckahn ☕

src/components/Toolbelt/Toolbelt.js
```diff
@@ -1,22 +1,33 @@
...
 
 import FarmhandContext from '../../Farmhand.context'
-import './Toolbelt.sass'
-import { tools as toolImages, pixel } from '../../img'
 import toolsData from '../../data/tools'
 
-const tools = Object.values(toolsData).sort(t => t.order)
+import { tools as toolImages, pixel } from '../../img'
+
+import './Toolbelt.sass'
```
lstebner, line 15: i added some line breaks to the these imports which added some noise to the PR here

src/data/achievements.js
```diff
@@ -275,6 +275,18 @@ const achievements = [
...
+  (() => ({
+    id: 'gold-digger',
+    name: 'Gold Digger',
+    description: `Pay off your loan from the bank.`,
+    rewardDescription: `The Shovel`,
+    condition: state => state.loanBalance === 0,
+    reward: state => ({
+      ...state,
+      shovelUnlocked: true,
+    }),
```
jeremyckahn, line 288: This works way better visually now. Nice work with this!

jeremyckahn (approved):
Awesome work! The code looks great and the new feature works perfectly. LGTM! 👏

PR #160, opened 2021-07-13
chore: updating deprecated method name reference

lstebner:
### What this PR does 

 a warning was being output in the console that this method name
  changed from createMuiTheme to createTheme so this updates that import
  and call

run the branch and see that the warning no longer shows up in the console. it was also being output from the test runner and can be validated by no longer seeing it there as well.

### Additional information

Screen Shot 2021-07-12 at 10 23 56 PM


lstebner:
actualllly i think this was from me still having a different version of npm! closing

PR #168, opened 2021-07-20
fix: put gold-digger achievement behind feature flag

lstebner:
### What this PR does

the gold-digger achievement was not behind the MINING feature flag and was unlockable in production, even though the shovel didn't unlock. this fixes that bug and only adds the achievement when the feature is actually enabled.

- go to production farmhand game
- open the achievements list
- find "gold digger" at the bottom
- next, go to vercel preview URL
- open the achievement list
- notice the 'gold digger' achievement is no longer there!


jeremyckahn (approved):
LGTM! Thanks for the quick fix. I’ll merge and deploy this now!

PR #169, opened 2021-08-03
Prevent incompatible cow breeding

### LINKED ISSUE #130: Show error UI when incompatible cows are set to breed
It’s too easy to set up same-gendered cows to breed. This leads to player confusion as to why no offspring are produced. ~Let’s add an obvious error indicator to guide the user.~

Per a suggestion from @seawolff and @lstebner [on Discord](https://discord.com/channels/714539345050075176/714539345637408793/857699568211591178), it'd be better to prevent the user from putting incompatible cows in the breeding pen instead of building an error UI state.

jeremyckahn:
### What this PR does

This PR fixes #130. It doesn't actually show an error UI when the player tries to breed incompatible cows, it just disables the checkbox so that incompatible pairs can't be set up.

I also converted all of the `CowPenContextMenu.test.js` tests to use React Testing Library.

Try to add compatible (one male, one female) breeding pairs to the Breeding Pen, and incompatible ones.

Let me know how I can improve the new tests! I'm pretty green with React Testing Library and would love to know where I can improve with using it.

### Additional information

https://user-images.githubusercontent.com/366330/128024601-28774f4b-f87f-46bf-ac54-3071bd893e2b.mp4


package.json
```diff
@@ -33,6 +33,7 @@
   },
   "devDependencies": {
     "@jeremyckahn/minami": "^1.3.1",
+    "@testing-library/jest-dom": "^5.14.1",
```
lstebner, line 36: 👍

src/components/CowPenContextMenu/CowPenContextMenu.js
```diff
@@ -66,107 +67,134 @@ const CowBloodline = memo(({ colorsInBloodline }) => (
...
-          <FormControlLabel
+          >
+            <FormControlLabel
+              {...{
+                control: (
+                  <Checkbox
+                    {...{
+                      color: 'primary',
+                      // TODO: This Boolean cast is needed for backwards
+                      // compatibility. Remove it after 10/1/2020.
```
lstebner, line 150: can be removed?

src/components/CowPenContextMenu/CowPenContextMenu.test.js
```diff
@@ -1,122 +1,158 @@
...
+          />
+        )
 
-        expect(
-          component
-            .find('.card-list')
-            .find(CowCard)
-            .props().isSelected
-        ).toEqual(false)
+        expect(container).toMatchSnapshot()
```
lstebner, line 60: is there a specific rendered result we can check for instead? these snapshots that snap an entire rendered output are so sensitive to future changes i find they end up being a pain instead of helpful (literally every change breaks them). part of me think there's a chance this test isn't even needed with all the specific checks for the checkbox and heart statuses later on. and maybe the next snapshot too. what do you think? might be easier to chat in discord if you want!

src/components/CowPenContextMenu/CowPenContextMenu.test.js
```diff
@@ -125,90 +161,296 @@ describe('CowCard', () => {
...
+              ...baseProps,
+              cow: generateCow({
+                color: cowColors.WHITE,
+                happiness: 0.5,
+                name: '',
+                baseWeight: 100,
+              }),
+              isCowPurchased: true,
+            }}
+          />
```
lstebner, line 431: random comment, but you can do these as ```<CowCardSubheader { ...baseProps} cow={generateCow(...)}  isCowPurchased />``` instead of spreading everything. sorry if you already told me a reason some other time why you prefer the style you use 😄

src/components/CowPenContextMenu/CowPenContextMenu.js
jeremyckahn: Good catch! Thanks for calling this out. Removed in https://github.com/jeremyckahn/farmhand/commit/b438824f0debafd6e12df93d93e697f069641a53!

src/components/CowPenContextMenu/CowPenContextMenu.test.js
jeremyckahn: Hah this is just my preferred JSX style. :) I find it generally easier to treat all JSX properties as standard JavaScript objects. I realize that this is not the common React style, so I don't expect anyone else to use it when working on the project! 😄

src/components/CowPenContextMenu/CowPenContextMenu.test.js
jeremyckahn: >is there a specific rendered result we can check for instead?

I scoped this in as closely as I could in https://github.com/jeremyckahn/farmhand/commit/f5f4f9dd450b17a9206885b32bcfb9d24fa432ea, but it's still pretty large, unfortunately.

>these snapshots that snap an entire rendered output are so sensitive to future changes i find they end up being a pain instead of helpful (literally every change breaks them).

I definitely agree, but short of checking that the `Card` component received the specific `raised` prop value (which is an antipattern worth moving away from), there's no great way that I can think of to check the rendered state of selected vs. non-selected `CowCard` state. To put it in concrete terms, this test is specifically interested in the classlists here: https://github.com/jeremyckahn/farmhand/blob/f5f4f9dd450b17a9206885b32bcfb9d24fa432ea/src/components/CowPenContextMenu/__snapshots__/CowPenContextMenu.test.js.snap#L5

and here: https://github.com/jeremyckahn/farmhand/blob/f5f4f9dd450b17a9206885b32bcfb9d24fa432ea/src/components/CowPenContextMenu/__snapshots__/CowPenContextMenu.test.js.snap#L371

The fact that the `MuiPaper-elevationX` classes are the sole distinguishing characteristic of these two variations is a Material-UI implementation detail that our application tests should not be concerned with. The fact that there are hundreds of lines of unrelated noise are an unfortunate side effect of how React Testing Library rendering and Jest snapshots work.

I would _love_ a better solution here that is simpler, more robust, and adherent to best practices. Do you have any ideas that might meet all these needs better?

>there's a chance this test isn't even needed with all the specific checks for the checkbox and heart statuses later on. and maybe the next snapshot too.

I think it's worth testing that our component renders selected vs. non-selected `CowCard`s correctly. I suppose a lighter-weight approach to doing this is to conditionally set a `selected` class on the `CowCard`, but then we're adding functionality specifically for the sake of testing.

---------------------

To sum up, I can't find a solution that feels great here, but what's here feels most closely aligned to React and testing best practices. What are your thoughts about this?

jeremyckahn (commented):
Thanks for the review! I made some changes but I think we have a bit more to figure out with regards to the snapshot testing. Back to you @lstebner!

src/components/CowPenContextMenu/CowPenContextMenu.test.js
lstebner: in this case the long term solution may be that we need to add more to distinguish the two states, the className alone is quite meaningless. i'm cool with future improvements rather than holding up this PR longer, but something like rendering some [visually hidden](https://css-tricks.com/inclusively-hidden/) text saying "selected" for the selected one would be an option.

jeremyckahn: >something like rendering some visually hidden text saying "selected" for the selected one would be an option.

That's a great idea! I think I'll roll that into this PR before merging because it solves this problem well and also adds a meaningful user feature for little work. Thanks for the suggestion!

jeremyckahn: This is addressed in https://github.com/jeremyckahn/farmhand/commit/d7e793e3f426a56962ff9c2ce2c6b60ab29eb7df. This actually cleans up this whole PR quite nicely. Thanks again for the great suggestion!

lstebner (approved):
looks good to me! nice work on this

Issue #170, opened 2021-08-04
Document keybindings


jeremyckahn:
This should exist either in the wiki or in an in-game modal when the `?` key is pressed.

PR #172, opened 2021-08-07
Keyboard shortcut info

jeremyckahn:
### What this PR does

It adds onboard documentation for in-game keyboard shortcuts. (#170)

Press `shift + ?` anywhere in the game.

### Additional information

2021-08-07 11_05_02-Settings

### DIFF
--- a/src/Farmhand.js
+++ b/src/Farmhand.js
@@ -408,8 +408,9 @@ export default class Farmhand extends Component {
       )
     })
 
-    // The dialog view mappings here must be kept in sync with the
-    // dialogTriggerTextMap map in Navigation.js.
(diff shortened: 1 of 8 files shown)


jeremyckahn:
Per @worry on Discord, this should use some sort of accessibility affordance (such as [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute)) to inform non-sighted users about the availability of this modal.

jeremyckahn:
Rather than block this PR on pending accessibility enhancements, I've spun off that work into #175 to be addressed separately. I'd really love some help with that ticket, as I don't know what the correct solution for it is!

PR #176, opened 2021-08-15
Lstebner/163 smelter

lstebner:
### What this PR does

adds a new Smelter item that is purchasable as part of the MINING feature

run the app locally to have the MINING feature enabled, and then simply check out the store for the Smelter. it is purchasable in this PR, but not usable yet.

### Additional information

Screen Shot 2021-08-15 at 10 10 23 AM

Screen Shot 2021-08-15 at 12 06 42 PM
Screen Shot 2021-08-15 at 12 06 48 PM


src/data/ores/stone.js
```diff
@@ -8,7 +8,7 @@ const { freeze } = Object
  */
 export const stone = freeze({
   description: 'A piece of rock.',
-  doesPriceFluctuate: true,
+  doesPriceFluctuate: false,
```
lstebner, line 11: when reviewing this i started thinking the stone shouldn't fluctuate in price, but the other ores still should. open to changing this back though

src/components/Shop/Shop.js
```diff
@@ -135,6 +139,23 @@ export const Shop = ({
           }}
         />
       </li>
+      {features.MINING ? (
+        <li>
+          <TierPurchase
+            {...{
+              description:
+                'You can purchase a Smelter to convert ore into ingots, and other useful items.',
```
jeremyckahn, line 147: Grammar nit: Remove the comma and change this to:

```
'You can purchase a Smelter to convert ore into ingots and other useful items.'
```

(I'm trying to break myself of this habit too!)

src/components/Shop/Shop.js
```diff
@@ -135,6 +139,23 @@ export const Shop = ({
           }}
         />
       </li>
+      {features.MINING ? (
+        <li>
+          <TierPurchase
```
jeremyckahn, line 144: Good reuse of the `TierPurchase` component here! 😃

src/components/Shop/Shop.js
```diff
@@ -135,6 +139,23 @@ export const Shop = ({
...
+            {...{
+              description:
+                'You can purchase a Smelter to convert ore into ingots, and other useful items.',
+              handleTierPurchase: handleSmelterPurchase,
+              maxedOutPlaceholder: "You've already purchased the smelter.",
```
jeremyckahn, line 149: Grammar nit: Change this sentence to end with "!" to be consistent with the other `maxedOutPlaceholder` strings. I could be swayed to change them all to sound less excited, but either way they should be consistent.

src/constants.js
```diff
@@ -35,14 +35,18 @@ export const INITIAL_FIELD_HEIGHT = 10
 
 export const PURCHASEABLE_FIELD_SIZES = freeze(
   new Map([
-    [1, { columns: 8, rows: 12, price: 1000 }],
-    [2, { columns: 10, rows: 16, price: 2000 }],
-    [3, { columns: 12, rows: 18, price: 3000 }],
+    [1, { columns: 8, rows: 12, price: 1_000 }],
+    [2, { columns: 10, rows: 16, price: 2_000 }],
+    [3, { columns: 12, rows: 18, price: 3_000 }],
```
jeremyckahn, line 40: This is a cool way to format these number! 👍

src/data/ores/stone.js
jeremyckahn: This seems reasonable to me!

src/data/recipes.js
```diff
@@ -309,3 +309,55 @@ export const popcorn = itemify({
...
+ * @property farmhand.module:recipes.ironIngot
+ * @type {farmhand.recipe}
+ */
+export const ironIngot = itemify({
+  id: 'iron-ingot',
+  name: 'Iron Ingot',
+  ingredients: {
+    [items.ironOre.id]: 5,
+  },
+  condition: state => state.purchasedSmelter,
```
jeremyckahn, line 336: Are we sure we want the condition for all of these recipes to be the same (`state.purchasedSmelter`)? It might give the player a little more meaningful work to also have to dig up some number of each ingots' ingredient ores before they can be smelted.

jeremyckahn (changes requested):
Great work here so far! There's a few minor things to discuss, but nothing significant. I do want to get your thoughts on maybe changing the unlock conditions for the new Ingot items and left a comment for that.

I want to make some (extremely minor) tweaks to the new images, but I'll take care of that and push the changes myself!

jeremyckahn:
I just pushed some art tweaks in https://github.com/jeremyckahn/farmhand/pull/176/commits/48cc51c4d52747c8f6b899220f94a77be41d242d:

Screen Shot 2021-08-15 at 4 12 00 PM

Screen Shot 2021-08-15 at 4 12 11 PM

Let me know what you think!

src/components/Shop/Shop.js
lstebner: ah yeah i get comma happy hah

src/data/recipes.js
lstebner: i had this same thought too actually so i think that means we should update this to require having dug up a few ores. will update

lstebner: i added a "sold at least 50" to each. since they will each have a different rarity this should make their unlock timings fluctuate nicely

jeremyckahn: 50 is an interesting choice here. It's relatively high compared to most other recipes, but I think it's worth exploring how players respond to higher unlock requirements for mining-related items. Let's run with it and give players a challenge! 😄

lstebner: let's also see how it is once we have the ability to actually dig them up. if this ends up being a lot of grinding we should probably reduce it a bit

jeremyckahn (approved):
LGTM! Awesome work @lstebner!

jeremyckahn:
I noticed some incorrect `propTypes` after this was merged, and I corrected them in https://github.com/jeremyckahn/farmhand/commit/ad04e64083e0565ee47f05766bd6137e5cc167ae.

Issue #175, opened 2021-08-13
Improve nonsighted accessibility of keyboard shortcut modal


jeremyckahn:
> Per @worry on Discord, this should use some sort of accessibility affordance (such as [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute)) to inform non-sighted users about the availability of this modal.

_Originally posted by @jeremyckahn in https://github.com/jeremyckahn/farmhand/issues/172#issuecomment-897643802_

lstebner:
I am pretty sure the requirement here is to add an `id` to the `h2` element that inside of the modal, and then to use that as the `aria-describedby` value for the modal. so it would be something like:

```
<div class="the-modal" aria-describedby="hotkeys-modal-heading">
   <h2 id="hotkeys-modal-heading">Keyboard Shortcuts</h2>
   ...
```

PR #179, opened 2021-08-15
Modal accessibility enhancements

jeremyckahn:
### What this PR does

For #175, this PR adds `aria-describedby` and `aria-labelledby` attributes to the Keyboard Shortcuts modal. Since Farmhand's modals are implemented generically, this also enhances all of the other modals as well!

References for implementation: 
  - https://github.com/jeremyckahn/farmhand/issues/175#issuecomment-899082811
  - https://material-ui.com/components/modal/#accessibility

I'm actually not sure how this can be validated and would appreciate some help with determining this. So far I've just made our DOM look like the suggested solutions, but I don't know how to verify that it's providing the intended experience to relevant users.

### Additional information

Here's how the DOM for this looks:

Screen Shot 2021-08-15 at 5 15 48 PM

### DIFF
--- a/src/components/Navigation/Navigation.js
+++ b/src/components/Navigation/Navigation.js
@@ -23,6 +23,7 @@ import SettingsIcon from '@material-ui/icons/Settings'
 import Switch from '@material-ui/core/Switch'
 import TextField from '@material-ui/core/TextField'
 import Tooltip from '@material-ui/core/Tooltip'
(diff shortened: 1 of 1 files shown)


jeremyckahn:
This was [discussed a bit on Discord](https://discord.com/channels/714539345050075176/847884323330850847/876654450536775710) and @worry seems to think what's here should work.

PR #180, opened 2021-08-15
feat: [closes #134] shovel plots for ore

### LINKED ISSUE #134: Shoveling plots

### LINKED ISSUE #165: Coal
add the item Coal to be used for smelting. as part of this, i think we'll want to update the recipes for the ingots to require some coal as well

### LINKED ISSUE #198: add extra restriction to unlocking the shovel so it can't be unlocked immediately
need to add something additional to the shovel unlock rules so that it cannot be immediately unlocked when starting the game. a few ideas came up in discord around either having hit at least day X, or having sold something(s) in the shop in addition to the pay off your loan criteria

lstebner:
### What this PR does

  - adds spawnChance to data/ores objects
  - adds constants for spawn chances
  - adds `ResourceFactory` for spawning mined resources
  - adds `OreFactory`, `CoalFactory`, and `StoneFactory` for generating specific resources
  - change wasShoveledToday to wasShoveled and adding ability to have shoveled state last a varying amount of days based on randomness and also rarity of ore spawned if any was spawned
  - closes #198 - update gold-digger achievement to only unlock after day 1
  - closes #165 - add coal
  - closes #134 - general logic for shoveling plots
  - adds a `randomChoice` util for picking a random choice out of a list of weighted choices
  - new art assets for ores
  - art asset for coal
  - added coal requirements to ingot recipes

check out the branch and run it locally to get the mining feature enabled. 

- pay off your loan (with any save, after day 1)
- use the shovel on plots

mostly unsure about the spawn rate of things if anyone has any feedback about how frequently things spawn

### Additional information

Screen Shot 2021-09-04 at 2 48 07 PM

Screen Shot 2021-09-04 at 2 46 26 PM

Screen Shot 2021-09-04 at 2 53 00 PM

Screen Shot 2021-09-04 at 2 52 48 PM


src/data/ores/index.js
```diff
@@ -1,5 +1,24 @@
...
+  var oreToSpawn = null
+
+  for (let ore of ores) {
+    if (value < ore.spawnChance) {
+      oreToSpawn = ore
+      break
+    }
+  }
+
+  return oreToSpawn
```
lstebner, line 23: @jeremyckahn i would like your feedback on this approach before i go too far. the way i have it, one random number is generated and then compared against all the ore spawn chances so order is important (lowest chance to highest). i rolled that up into this `oreSpawner` method, and just put it here to test out. does this seem like the right approach?

src/reducers.js
```diff
@@ -167,8 +168,20 @@ export const resetWasWatered = plotContent =>
  * @param {?farmhand.plotContent} plotContent
  * @returns {?farmhand.plotContent}
  */
-export const resetWasShoveled = plotContent =>
-  plotContent && plotContent.wasShoveledToday ? null : plotContent
+export const resetWasShoveled = plotContent => {
```
lstebner, line 171: i updated the `wasShoveledToday` to be a bit more complex to allow for shoveled plots to stay in the shoveled state for multiple days. also would like your feedback on how this was done @jeremyckahn 😄

src/data/ores/index.js
jeremyckahn: As long as the implementation details are sufficiently abstracted away (which they seem to be), then it doesn't matter _too_ much how this logic works, so long as it behaves the way we want.

I'm hesitant to pick the code apart too much now because I think things will become much more intuitive and clear once we play test this a bit. I also don't have a great alternative implementation suggestion to offer at the moment. 😅

I think this is off to a good start, so let's get this minimally working and then iterate and polish it up once we have game behavior that feels right!

One thing I like about this is that since you're passing in the random value, this will be easy to test without mocking! 👍

src/reducers.js
jeremyckahn: I think this is a good addition to the game! Since there's now a truthy `daysUntilClear` value, I wonder if we need the `wasShoveled` boolean anymore. Could that be removed now?

If I'm reading this right, then `plotContent`s could have `daysUntilClear` of `0` but still not be cleared (they would be cleared the next time this is run). To me it seems that `plotContent.daysUntilClear === 0` should mean that the plot should be cleared _now_. What do you think @lstebner?

jeremyckahn (commented):
Great work so far! Since this is still a draft I'm holding off on a full review and just responding to the specific questions you asked.

jeremyckahn:
I just test played the current state a bit. This is really coming along! Great work so far. Here's a few comments that I shared in Discord that I'll put here for posterity:

  1. I feel that ores are mined too frequently right now. Unsuccessful `null` yields should be typical, perhaps 75% of the time or so.
  2. It may be a bug, but right now all plots that yield an ore are restored after a constant single day for me. I think it would liven up the gameplay a bit if restoration was variable between maybe 1-3 days. We could also consider basing the plot restoration time on the ore that was mined (e.g. better ores result in longer plot restoration) for a balanced cooldown. That's definitely scope creep, so feel free to defer that (if you're even into the idea)!

lstebner:
did quite a bit of rebalancing based on your feedback @jeremyckahn. i put a bit more logic into the cooldown duration and was able to base it on the rarity of the ore spawned plus some randomness without too much effort. good suggestions! I also added them to the inventory so you can actually see them now! 😄  let me know what you think when you get to try it out

jeremyckahn:
Awesome work @lstebner! I didn't get a chance to review the code yet (my morning got away from me), but I did have a chance to give this a test play. This is feeling much better to me now! The balance between ore yields and `null` yields feels right, as does the balance of high-value to low-value ore distribution. From a gameplay perspective, I think the behavior is just about where it should be. We'll definitely need some visual feedback for successful ore yields, but I'd be fine with saving that for a separate PR since this is all behind a feature flag right now.

I'll do a code review just as soon as I can make time for it. Did you want to keep working on this and leave it in Draft state, or do you think it's feature-complete enough to promote to a full PR?

src/reducers.js
lstebner: i overlooked this comment for some reason, but yeah what you said is true. these don't really "clear" until the value falls below 0. i can update so it clears at 0, but i will probably also then go +1 the days until clear value when shoveling since i've been testing with this code in place and it's felt pretty good.

lstebner: made this change!

lstebner:
agreed on visuals for ore @jeremyckahn . i'll add something to the project board so we don't forget. i gave this a review myself this morning and tweaked some things that stood out to me, but i'm ready for official review now for sure!

src/data/ores/index.js
lstebner: i ended up moving that random value bit because it ended up using more than 1 random number, but open to suggestions if you have them

src/factories/OreFactory.js
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Generate resources
+   * @returns {Array} an array of ore resources
+   */
+  generate() {
+    return [this.spawn()]
+  }
+
+  /*
```
jeremyckahn, line 29: nit: Extra asterisk here.

src/factories/ResourceFactory.js
```diff
@@ -0,0 +1,111 @@
...
+        type
+      ] = ResourceFactory.generateFactoryInstance(type)
+    }
+
+    return ResourceFactory._factoryInstances[type]
+  }
+
+  /**
+   * Use dice roll and resource factories to generate resources at random
+   * @returns {?Array} array of resource objects, or null if no resoures were spawned
```
jeremyckahn, line 93: What do you think about returning an empty array here instead of `null`? It might help us cut down on special case logic.

src/factories/ResourceFactory.js
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Used for spawning mined resources
+ * @constructor
+ */
+export default class ResourceFactory {
+  constructor() {
+    this.resourceOptions = [
+      { weight: ORE_SPAWN_CHANCE, name: itemType.ORE },
+      { weight: COAL_SPAWN_CHANCE, name: itemType.FUEL },
+      { weight: STONE_SPAWN_CHANCE, name: itemType.STONE },
```
jeremyckahn, line 23: Would it make sense for the `name`s to be called `itemType` here?

src/img/index.js
```diff
@@ -84,14 +84,15 @@ import scythe from './tools/scythe.png'
...
+import bronzeOre from './ores/stone-bronze-ore.png'
 import bronzeIngot from './ores/bronze-ingot.png'
-import ironOre from './ores/iron-ore.png'
+import ironOre from './ores/stone-iron-ore.png'
 import ironIngot from './ores/iron-ingot.png'
-import silverOre from './ores/silver-ore.png'
+import silverOre from './ores/stone-silver-ore.png'
 import silverIngot from './ores/silver-ingot.png'
-import goldOre from './ores/gold-ore.png'
+import goldOre from './ores/stone-gold-ore.png'
```
jeremyckahn, line 95: I'm curious, what's the thinking behind the new `stone-` prefix for these?

src/reducers.js
```diff
@@ -1678,8 +1691,33 @@ export const minePlot = (state, x, y) => {
     return state
   }
 
+  const spawnedResources = ResourceFactory.instance().generateResources()
+  let spawnedOre = null
+  let daysUntilClear = Math.round(Math.random() * 2) + 1
```
jeremyckahn, line 1696: Looks like we can use `chooseRandom` here as well.

src/index.js
```diff
@@ -58,7 +58,9 @@
  * Represents a shoveled plot
  * @typedef farmhand.shoveledPlot
  * @type {Object}
- * @property {boolean} wasShoveledToday
+ * @property {boolean} wasShoveled
```
jeremyckahn, line 61: Looking at this all again, what do you think about renaming this to `isShoveled`? It's a subtle (and perhaps a little annoying to make) semantic tweak that I think helps make the logic a little bit clearer since plots can remain shoveled for many days.

Sorry I didn't think of this before. Seeing this being used more in the code made me think of it! 😅

src/reducers.test.js
```diff
@@ -3317,12 +3319,37 @@ describe('minePlot', () => {
...
+    expect(gameState.field[0][0].oreId).toEqual(goldOre.id)
+  })
+
+  test('adds the spawned ore to the inventory', () => {
+    let itemIsInInventory = false
+
+    for (let item of gameState.inventory) {
+      if (item.id === goldOre.id) {
+        itemIsInInventory = true
+        return
```
jeremyckahn, line 3348: It looks like this should be `break` so that the `expect` below is always hit.

src/common/utils.js
```diff
@@ -19,3 +19,33 @@ export const generateValueAdjustments = (priceCrashes = {}, priceSurges = {}) =>
 
     return acc
   }, {})
+
+/**
+ * @param {Array} an array of objects each containing a `weight` property
+ * @returns {Object} one of the items from weightedOptions
+ */
+export function randomChoice(weightedOptions) {
```
jeremyckahn, line 27: Can this function be moved to `src/utils.js`? This file is specifically mean for utils that are shared between client and API code due to a quirk of how Vercel builds endpoints. See: https://github.com/jeremyckahn/farmhand/blob/f0c9fb8d5f5fdd33289b3df864670bc8848a2059/src/common/README.md#L1

Right now this is only being used by client code, so it should live with the rest of the client-specific utils.

jeremyckahn (changes requested):
Incredible work. I have a handful of change requests, but it's mostly tweaky stuff.

Thanks for sticking with this! I really like the factory pattern you've introduced; I think it's a good fit. And your pixel art skills are really improving — we should ship this art as-is! 🎨👏

src/common/utils.js
```diff
@@ -19,3 +19,33 @@ export const generateValueAdjustments = (priceCrashes = {}, priceSurges = {}) =>
...
+  sortedOptions.sort(o => o.weight)
+
+  let diceRoll = Math.random() * totalWeight
+  let option
+  let runningTotal = 0
+
+  for (let i in sortedOptions) {
+    option = sortedOptions[i]
+
+    if (diceRoll < option.weight + runningTotal) {
```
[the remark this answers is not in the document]
lstebner: it shouldn't, i added some more unit tests to cover the upper/lower bounds of Math.random to make sure of it. but if you have a scenario in mind you see failing can you maybe add an example of that?

src/common/utils.test.js
```diff
@@ -42,3 +42,16 @@ describe('generateValueAdjustments', () => {
...
   })
 })
+
+describe('randomChoice', () => {
+  const choices = [
+    { weight: 0.2, name: 'first-choice' },
+    { weight: 0.5, name: 'second-choice' },
+    { weight: 0.3, name: 'third-choice' },
+  ]
+
```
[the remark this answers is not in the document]
lstebner: i kinda responded to this in the other comment about this method, but there are some more tests now. let me know if you want some more!

src/constants.js
```diff
@@ -187,3 +187,19 @@ export const TOOLBELT_FIELD_MODES = new Set([
...
+export const ORE_SPAWN_CHANCE = 0.25
+export const COAL_SPAWN_CHANCE = 0.15
+export const STONE_SPAWN_CHANCE = 0.6
+
+// if spawning ore, which kind?
+// note: these values end being used relative to each other
+export const BRONZE_SPAWN_CHANCE = 0.4
+export const GOLD_SPAWN_CHANCE = 0.07
+export const IRON_SPAWN_CHANCE = 0.33
+export const SILVER_SPAWN_CHANCE = 0.2
```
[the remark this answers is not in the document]
lstebner: i'm on the fence with these. i would normally say that i like them all spelled out, but in this i did add some comments to group them and i don't really expect a comment to do the job forever. if i grouped them, i would probably group the above 3 as well so it ends up like:

```
RESOURCE_SPAWN_CHANCE =
RESOURCE_SPAWN_CHANCES = {
  ORE: ,
  COAL: ,
  STONE: ,
}
ORE_SPAWN_CHANCES = {
  BRONZE: , 
  etc
}
```

the only thing i don't like is that a single `S` separates those 2 very long variable names.

src/factories/CoalFactory.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+export default class CoalFactory {
+  /**
+   * Generate resources
+   * @returns {Array} an array of coal and stone resources
+   */
+  generate() {
+    let spawns = []
+
+    const amount = Math.floor(Math.random() * 2) + 1
```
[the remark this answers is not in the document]
lstebner: looks like it! didn't realize that was in there. that actually lets me weight the amount return a bit too which is cool. there might even be another place i can use this now that i know about it..

src/factories/OreFactory.js
lstebner: not a nit, a requirement actually! i actually just realized that yesterday hah. nice catch

src/factories/ResourceFactory.js
lstebner: i think it's a good call, updated this

src/factories/ResourceFactory.js
lstebner: good catch. when i started this i was thinking `randomChoices' would always expect "weight" and "name" but didn't end up sticking to that. updating

src/img/index.js
lstebner: me being unsure about replacing the old images 😅 . but it seems like we agree they turned out well so let me update this now..

src/index.js
lstebner: no worries, i think `isShoveled` makes more sense. updating

src/reducers.test.js
lstebner: oh yeah good catch

src/common/utils.js
lstebner: yup, that makes sense! thanks for explaining the difference

src/img/ores/bronze-ore.piskel
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"modelVersion":2,"piskel":{"name":"stone-bronze-ore","description":"","fps":12,"height":24,"width":24,"layers":["{\"name\":\"Layer 1\",\"opacity\":1,\"frameCount\":1,\"chunks\":[{\"layout\":[[0]],\"base64PNG\":\"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAABdklEQVRIS+2UvUoDQRSFJ4WkUIxvoIKFD5LGysJXsNhUCgb0ERQixG4LXyGFlU0ewkZYUDCCrYighVisnBvPemZn9qcJKDhNNjOb79x7zp103IJXZ8F89y/Q6PCvsSivKbW2yKYODHw2One9lWXTeH17d3tLU3uefPbd8fBItQNenUAOsML4DOLlw3rQ1Ok4xZ7HrBIw+MvTXQHZ33z0gNPurn3vf1wV3aCArYNrTyQm4FVeBuHX2AOYn1QWywpuWSC/v9gxb7lQlYrwWbPguzhLkqSyA4PrUiGGzHMIYI8WaSaahXZgArfPc0S2Ou+CkFioOEc27IqZaA6eQJraFLhZdhNMCEOOZcJcaFulwMmh+WdVqa+0oWyf2haDBzPrnMvRBYGxcFVslG2Y5lr356KX70IwRbSJY6hVK1wzQccQG27P2t0DQJPBwKtMbWPlFAcYq3zJYhZpwTlEdAEUg3+Do7xWf3bBSIUblZwmgRbs+lf+vsAXSn2tGWpeb/IAAAAASUVORK5CYII=\"}]}"]}}
```
jeremyckahn, line 1: Since the file names have been updated (thank you!), please make sure to update the `name`s in the Piskel source files to match.

src/utils.js
```diff
@@ -1015,3 +1015,33 @@ export const getSalePriceMultiplier = (completedAchievements = {}) => {
 
   return salePriceMultiplier
 }
+
+/**
+ * @param {Array} an array of objects each containing a `weight` property
```
jeremyckahn, line 1020: I just noticed this because my editor appropriately highlights the variable name, but for this JSDoc syntax to be correct, this line needs to begin with `@param {Array} weightedOptions [description...]`.

jeremyckahn (changes requested):
We're almost there! Just a few more small wrinkles to iron out, as well as figuring out what to do with the spawn chance constants.

src/constants.js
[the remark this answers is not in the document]
lstebner: i like "MINED_PLOT_SPAWN_CHANCE". let's do that! thanks for the suggestion

src/img/ores/bronze-ore.piskel
lstebner: oh yeah good catch

jeremyckahn (approved):
Everything looks great! Awesome work. Thanks for sticking with this and making a huge improvement to the game. LGTM!

PR #181, opened 2021-08-19
Lstebner/178 combine recipe messages

lstebner:
### What this PR does

Adds a new notification for when multiple recipes are unlocked at the same time. This fixes a bug where in this situation only the very last message would actually get displayed. This introduces a new template to handle this scenario with a formatted list notification.

The easiest way to test this is to import a save that has a bunch of achievements unlocked and then the message will display all of those at once. Unlocking multiple in game is actually pretty rare, but using this route to test can actually expose a new issue...

Screen Shot 2021-08-19 at 9 02 57 PM

not a huge deal because unlocking this many at once is unrealistic for normal gameplay, but we could probably add some styles for wider screens to limit the width if we want to prevent this.


src/templates.js
```diff
@@ -95,15 +95,14 @@ export const RECIPES_LEARNED = (_, learnedRecipes) => {
   const learnedRecipeNames = learnedRecipes.map(({ name }) => name)
 
   if (learnedRecipes.length === 2) {
-    recipesString = learnedRecipeNames.join(' and ')
+    recipesString = `**${learnedRecipeNames[0]}** and **${learnedRecipes[1]}**`
```
jeremyckahn: Looks like `learnedRecipes` should be `learnedRecipeNames`. 😄

Screen Shot 2021-08-20 at 8 27 00 AM

Fixed in https://github.com/jeremyckahn/farmhand/commit/b1566fe4d2dc9cb1710ce7c4cecdb9d00cc99275!

jeremyckahn (approved):
LGTM! Thanks for pitching in and driving the solution here @lstebner.

Screen Shot 2021-08-20 at 8 37 36 AM

Screen Shot 2021-08-20 at 8 27 51 AM

Screen Shot 2021-08-20 at 8 25 14 AM


Let's ship it! Regarding your initial comment:

>not a huge deal because unlocking this many at once is unrealistic for normal gameplay, but we could probably add some styles for wider screens to limit the width if we want to prevent this.

Good to call out, but I'm also not too worried about it. It looks a little silly but it's not game-breaking, and I imagine it will mostly only happen under contrived scenarios.

src/templates.js
lstebner: good catch!

Issue #177, opened 2021-08-15
add Smelter area to Workshop


lstebner:
once the Smelter is purchased it should be a separate  "area" (tab maybe?) of the workshop. this interface might take a little bit of design to sort out, or maybe MUI provides something already.

jeremyckahn:
Per @lstebner [on Discord](https://discord.com/channels/714539345050075176/878706542604591124/878728088136060978):

>i wonder if we should flip “recipes” and “workshop”. everything that is made could be seen as “recipes” and then the tabs are sorta like locations of the farm to craft in. maybe forge and workshop to start? or forge and kitchen? since it is mainly foods

My response:

>I think breaking the tabs down into "forge" and "kitchen" makes sense!

jeremyckahn:
@lstebner were you envisioning anything beyond a tabbed space for Forge recipes? Do you think we need some additional bit of UI that the Kitchen area doesn't have?

<img width="397" alt="Screen Shot 2021-08-22 at 7 04 06 PM" src="https://user-images.githubusercontent.com/366330/130374645-682f544e-c3a6-438e-afb3-2559b194b86c.png">

lstebner:
> @lstebner were you envisioning anything beyond a tabbed space for Forge recipes? Do you think we need some additional bit of UI that the Kitchen area doesn't have?

hmm the only thing that this makes me think of is an area to insert fuel that could be outside of any recipe, but we could also just make recipes require some fuel. either way.. i think it's scope creep for this PR so I wouldn't worry about it. let's discuss for the mining feature in general. this is looking great!

jeremyckahn:
Thanks for spot-checking this @lstebner! I agree with reining in the scope here — I'd like this ticket to be about setting up a space in the UI for Forge recipes. Let's continue to discuss what will go here in a separate thread!

As you pointed out [on Discord](https://discord.com/channels/714539345050075176/878706542604591124/879199622772432937), tab navigation is currently broken so I'd like to fix that if I can before opening a PR, as well as maybe some code cleanup.

jeremyckahn:
Ugh I just realized that MUI tabs expect you to use the left/right arrow keys to move between tabs, but that clashes with Farmhand's global left/right view switch key handlers. I'll have to look into whether we can use a tab-based navigation here, or perhaps intelligently disable the global left/right key handler. 😕

jeremyckahn:
[...]

PR #183, opened 2021-08-25
Add Forge area to Workshop

jeremyckahn:
### What this PR does

For #177, this PR adds a space in the Workshop area for Forge recipes (like Ingots). While only available when the `MINING` feature is enabled, the new tabbed Workshop interface will be available to all players once this is shipped.

This also fixes a minor bug with miscounting the Kitchen recipes, so I'll deploy this after merging it.

Go to this preview URL: https://farmhand-git-jeremyckahn-177add-smelter-to-workshop-jeremyckahn.vercel.app/?enable_MINING=true, or any other environment where the `MINING` feature is enabled (like localhost) and unlock some Ingot recipes. I modified my local Ingot recipe conditions to make this a little easier.

I went with different semantics from what was originally requested (Forge instead of Smelter), so let me know if this works!

There's some preexisting informational copy below the kitchen recipes, but not the Forge recipes. @lstebner, let's work together on what should go in there (though I'm fine with punting this to another PR as well).

### Additional information

Screen Shot 2021-08-25 at 8 41 25 AM


src/components/Workshop/Workshop.js
```diff
@@ -1,60 +1,139 @@
...
+
+  return (
+    <section
+      role="tabpanel"
+      hidden={value !== index}
+      id={`workshop-tabpanel-${index}`}
+      aria-labelledby={`workshop-tab-${index}`}
+      {...other}
+    >
+      {value === index && children}
```
lstebner, line 31: i've been warned this style can sometimes cause unexpected behavior and a ternary with a `null` case is preferred. i forget the exact problem scenario, but can try to look it up if you're interested

src/components/Workshop/Workshop.js
```diff
@@ -1,60 +1,139 @@
...
-          <li key={recipeId}>
-            <Recipe
-              {...{
-                recipe: recipesMap[recipeId],
-              }}
-            />
-          </li>
-        ))}
-      </ul>
+function TabPanel(props) {
```
lstebner, line 20: i like sub components like this for organization a lot, but prefer to move them into their own file. the pattern i typically follow is to just add a sub folder (in this case `TabPanel/`) to this current dir and move all js/css needs into there

src/components/Workshop/Workshop.js
```diff
@@ -1,60 +1,139 @@
...
+  'aria-controls': `workshop-tabpanel-${index}`,
+})
+
+const Workshop = ({ learnedRecipes }) => {
+  const [currentTab, setCurrentTab] = useState(0)
+
+  const learnedKitchenRecipes = []
+  const learnedForgeRecipes = []
+
+  for (const learnedRecipeId of Object.keys(learnedRecipes)) {
```
lstebner, line 53: this would make a nice little utility helper to break out as well. i actually wouldn't mind two (`getLearnedKitchenRecipes` and `getLearnedForgedRecipes`) but i see the appeal of a singular loop too. either way it seems like it could be plucked out of this component pretty easily

src/data/maps.js
```diff
@@ -20,11 +20,30 @@ const {
...
-}, {})
+export const kitchenRecipesMap = []
+export const forgeRecipesMap = []
+
+for (const recipeId of Object.keys(recipes)) {
+  const recipe = recipes[recipeId]
+
+  switch (recipe.recipeType) {
+    case recipeType.KITCHEN:
+      kitchenRecipesMap[recipe.id] = recipe
```
lstebner, line 31: did you choose to be explicit with separate lists on purpose? it would simplify this to make a data object that is keyed by `recipe.recipeType` instead, something like:

```
recipes = {
    [recipeType.KITCHEN]: [],
    [recipeType.FORGE]: [],
}```

the loop could actually build this out without requiring any defaults, and you can still ultimately have the same `recipesMap` export. this would just simplify the maintenance on this loop, but maybe that was intentional which is why i asked that first

lstebner:
> There's some preexisting informational copy below the kitchen recipes, but not the Forge recipes. @lstebner, let's work together on what should go in there (though I'm fine with punting this to another PR as well).

definitely something we can work out, but i agree with filing a separate issue and we'll tackle it there. should be a small change, but would hold up this PR if we try to sort it out now

lstebner (approved):
i think this looks good to go! 🎉

PR #211, opened 2021-09-11
Lstebner/166 scythe upgrade

lstebner:


src/data/upgrades.js
```diff
@@ -0,0 +1,64 @@
+import { itemType, toolType, toolLevel } from '../enums'
+
+import * as items from './items'
+import * as recipes from './recipes'
+
+const upgrades = {
```
lstebner, line 6: definitely open to feedback on the shape of this and this file in general

src/components/Inventory/Inventory.js
```diff
@@ -57,6 +58,7 @@ const itemTypeCategoryMap = Object.freeze({
   [SCARECROW]: FIELD_TOOLS,
   [SPRINKLER]: FIELD_TOOLS,
   [STONE]: MINED_RESOURCES,
+  [TOOL_UPGRADE]: '',
```
jeremyckahn, line 61: I'm taking this empty string to mean that `TOOL_UPGRADE` is treated as a class of items (in terms of the overall user experience) that has no `farmhand.item`s associated with it. I think it's reasonable to extend the item category system like this for now (but see below), but I'd like to see a clearer distinction between categories with items and categories without. I'd like to see some tweaks here:

  1. A comment explaining why `TOOL_UPGRADE` is different from the other categories.
  2. More of a nit, but consider changing the empty string to a `null` (it's functionally equivalent, but I do something similar in other parts of the codebase to more explicitly indicate something is defined, but not really useful or actionable).

-------------------------------

Not necessary for this PR, but in the future we may want to consider extending the item categories system to support other purchasable categories with separate maps/data structures. In other words, tacking on `TOOL_UPGRADE` to `itemTypeCategoryMap` works for now, but this pattern may not scale well once there's more than one exception case.

src/components/Recipe/Recipe.js
```diff
@@ -22,7 +22,7 @@ import FarmhandContext from '../../Farmhand.context'
 
 import './Recipe.sass'
 
-const IngredientsList = ({
+export const IngredientsList = ({
```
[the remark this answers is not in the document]
jeremyckahn: I think that makes sense!

src/components/Toolbelt/Toolbelt.js
```diff
@@ -5,68 +5,98 @@ import classNames from 'classnames'
...
+
+  return tools
 })
 
+const getToolImage = tool => {
+  if (tool.level === toolLevel.DEFAULT) {
+    return toolImages[tool.id]
+  }
+
+  let id = `${tool.id}-${tool.level.toLowerCase()}`
```
jeremyckahn, line 40: nit: Please change this `let` to `const` (or use the value directly in the lookup below).

src/components/Toolbelt/Toolbelt.js
```diff
@@ -5,68 +5,98 @@ import classNames from 'classnames'
...
-                }),
-                color: 'primary',
-                onClick: () => handleFieldModeSelect(fieldMode),
-                variant: fieldMode === currentFieldMode ? 'contained' : 'text',
+                key: fieldMode,
+                placement: 'top',
+                title: (
+                  <>
+                    <p>{alt}</p>
+                    <p>{levelInfo[toolLevels[type]]}</p>
```
jeremyckahn, line 64: :100:

src/components/Toolbelt/Toolbelt.sass
```diff
@@ -3,3 +3,10 @@
   button
     flex-grow: 1
     margin: 0 0.5em
+
+  .MuiButton-containedPrimary.selected
+    border: 1px solid #000
+    background-color: transparent
+
+    &:hover
+      background-color: #ffffb3
```
jeremyckahn, line 12: I dig these new styles. Nice work here!

src/components/Toolbelt/Toolbelt.test.js
```diff
@@ -12,36 +14,58 @@ jest.mock('../../config', () => ({
   },
 }))
 
+jest.mock('../../utils', () => ({
+  ...jest.requireActual('../../utils'),
+  memoize: jest.fn(callback => {
+    return (...args) => {
+      return callback(...args)
+    }
+  }),
```
jeremyckahn, line 23: This is a great way to test memoized functions. We should use this more often! :clap:

src/components/UpgradePurchase/UpgradePurchase.js
```diff
@@ -0,0 +1,95 @@
...
+import { IngredientsList } from '../Recipe/Recipe'
+
+import { canMakeRecipe, doesInventorySpaceRemain } from '../../utils'
+
+import { craftedItems } from '../../img'
+
+import FarmhandContext from '../../Farmhand.context'
+import './UpgradePurchase.sass'
+
+export class UpgradePurchase extends Component {
```
jeremyckahn, line 18: My preference would be use function-based components where possible (to stay consistent with current React best practices), but I'm curious if there was a specific reason you chose to go with a class-based component here?

src/components/UpgradePurchase/UpgradePurchase.js
```diff
@@ -0,0 +1,95 @@
...
+    const canBeMade =
+      canMakeRecipe(upgrade, inventory, 1) &&
+      doesInventorySpaceRemain({
+        inventory,
+        // Without the Infinity coercion, this would break recipes for unlimited
+        // inventoryLimits.
+        inventoryLimit:
+          (inventoryLimit === -1 ? Infinity : inventoryLimit) +
+          spaceFreedByIngredientsConsumed,
+      })
```
jeremyckahn, line 43: It looks like this was taken from `Recipe.js`: https://github.com/jeremyckahn/farmhand/blob/88cf18b487f351c5eb48d2e669e2efac15fcf53a/src/components/Recipe/Recipe.js#L75-L92

Do you think this could should be refactored out of `Recipe.js` and used in both places? Alternatively, perhaps a TODO to investigate it later?

jeremyckahn (changes requested):
I wasn't able to do a complete review yet, but I've added comments for the files I've had time to go through so far. I'll drop more comments as I have time. Great work so far! :smiley:

src/Farmhand.js
```diff
@@ -390,6 +396,12 @@ export default class Farmhand extends Component {
       todaysPurchases: {},
       todaysRevenue: 0,
       todaysStartingInventory: {},
+      toolLevels: {
+        [toolType.HOE]: toolLevel.DEFAULT,
+        [toolType.SCYTHE]: toolLevel.DEFAULT,
+        [toolType.SHOVEL]: null,
```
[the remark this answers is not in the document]
lstebner: yeah, i went back and forth a bit but landed on this so that things checking `toolLevels` can assume there's an entry for every _possible_ tool, but the `null` means it isn't unlocked yet. i suppose a `toolLevel.LOCKED` could exist? the "unlock" code though is what updates the shovel to the default level

src/components/Inventory/Inventory.js
lstebner: fully agree with what you've said here! i did this as a sort of band-aid to fix some errors and wasn't sure what the best long term solution is. i'll make your #1 and 2 changes, but am happy to make more changes that better set us up for the future if you want

src/components/UpgradePurchase/UpgradePurchase.js
lstebner: no _good_ reason 😄 . i started with a copy of `TierPurchase`, and i just tend to enjoy class components. will update!

src/components/UpgradePurchase/UpgradePurchase.js
lstebner: yup! i think putting it somewhere shared makes sense. utils, probably?

src/components/Inventory/Inventory.js
lstebner: i added a named `UPGRADES` category and things are working here. i think that cleans this up enough for now, but let me know what you think once you get back to this

src/reducers.js
```diff
@@ -1688,7 +1756,14 @@ export const minePlot = (state, x, y) => {
     return state
   }
 
-  const spawnedResources = ResourceFactory.instance().generateResources()
+  if (!doesInventorySpaceRemain(state)) {
+    return showNotification(state, INVENTORY_FULL_NOTIFICATION)
```
jeremyckahn, line 1760: Yesssssss! 👏👏👏

src/strings.js
```diff
@@ -31,3 +31,37 @@ export const HOE_HIDDEN_TEXT =
...
+    IRON: 'Iron - keeps plots watered for three days',
+    SILVER: 'Silver - keeps plots watered for four days',
+    GOLD: 'Gold - keeps plots watered for five days',
+  },
+  HOE: {
+    DEFAULT: 'Basic - destroys crops',
+    BRONZE: 'Bronze - 20% chance to retrieve seeds when destroying crops',
+    IRON: 'Iron - 40% chance to retrieve seeds when destroying crops',
+    SILVER: 'Silver - 60% chance to retrieve seeds when destroying crops',
+    GOLD: 'Gold - 80% chance to retrieve seeds when destroying crops',
```
jeremyckahn, line 62: These are some awesome upgrades! 😄

src/utils/isRandomChance.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Does a dice roll
+ * @param {float} chance float between 0-1 to compare dice roll against
```
jeremyckahn, line 3: `float` here should be `number`.

src/utils/isRandomChance.js
```diff
@@ -0,0 +1,8 @@
...
+ * @returns {bool} true if the dice roll was equal to or lower than the given chance, false otherwise
+ */
+export default function isRandomChance(chance) {
```
jeremyckahn, line 6: This function name a is a bit confusing to me, as it doesn't quite map to what this function is computing. I read `isRandomChance` as "is this thing a random chance?" but it's determining if a given value is greater than a random value. Maybe something like `isRandomNormalValueLessThan` or `isRandomlyLessThan` instead?

jeremyckahn (changes requested):
Another partial review! I haven't had a chance to go over your responses to my previous partial review yet but I at least wanted to power through some more of what's here. Incredible work! There's more I need to work though here but I at least wanted to share the feedback I've got so far.

Sorry it's taking me so long to get through this; I know it's taken you even longer to put this together. I just gotta take a PR like this in chunks and share feedback as I go. 🙂

src/components/UpgradePurchase/UpgradePurchase.js
jeremyckahn: That makes sense to me!

src/utils/isRandomChance.js
lstebner: i know what you mean, and i'm open to changing it, but i am going to attempt to defend the name first. i actually started using this type of method in a random project recently after dealing with mocking `Math.random` in a bunch of tests, very similar to the resource factory stuff. i started to think that not only is mocking `Math.random` a pain, but the lines of code often don't make all that much sense either. for example, i was using the name `diceRoll` as a var to try to help document what was going on in code like:

```javascript
const diceRoll = Math.random()
if (diceRoll < .5) {
    doSomething()
}
```

so, the name `isRandomChance` seemed to nicely solve both of these problems for me. it's easy to mock to return `true` or `false` and it makes the above code much easier to write, and to read.

```javascript
if (isRandomChance(.5)) {
    doSomething()
}
```

so, i sort of purposely avoided exposing the under the hood concept in favor of the way the method reads when used. all that said, i really am open to changing it. i just wanted to explain the whole story i had in mind.

src/components/IngredientsList/IngredientsList.js
```diff
@@ -0,0 +1,48 @@
...
+          <p
+            {...{
+              className: classNames(
+                playerInventoryQuantities[itemId] >= ingredients[itemId]
+                  ? 'in-stock'
+                  : 'out-of-stock'
+              ),
+            }}
+          >
+            {ingredients[itemId]} x {itemsMap[itemId].name} (On hand:{' '}
```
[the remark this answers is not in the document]
lstebner: actually i was overlooking this one, thanks for calling it out!

src/components/UpgradePurchase/UpgradePurchase.js
lstebner: so i went in to actually do this and added a util for getting the total number of ingredients in a recipe, but the `canBeMade` one i backed out of because it ended up looking more confusing with all the args it needs, and it is really just a series of other things that already come from utils. if you still want to break it out we can file an issue for it, but i think it needs further consideration that is more than we need to do for this PR. let me know what you think!

src/Farmhand.js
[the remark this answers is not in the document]
jeremyckahn: Ahhh that makes sense! In that case I think this would benefit from either a comment explaining the special case value or using a new enum value like `toolLevel.LOCKED` (though I'd prefer something a little less ambiguous like `toolLevel.UNAVAILABLE`). The second option doesn't seem like it requires explanation via comments, so that's perhaps the better route to take here. 🙂

src/components/UpgradePurchase/UpgradePurchase.js
jeremyckahn: Nah that's fine. A little bit of practical repetition is fine and certainly better than inscrutable code. Thanks for giving this a try!

src/strings.js
lstebner: glad you think so too! but i haven't fully implemented the hoe and watering can ones just fyi. i was thinking they could be followups to the mining feature, or at least this PR. what do you think?

src/utils/isRandomChance.js
jeremyckahn: I understand what you're saying. I think this utility function makes sense and fits well in the codebase, but the name just doesn't map to what it's doing for me. 😅

I think a much more explicit name like `isRandomNumberLessThan` will yield the same functional benefits while also making the code easier to understand. With `isRandomChance`, I feel like I'll have to keep looking at the code's implementation to know what it does.

src/strings.js
jeremyckahn: I'd be fine with tying that work off into a separate PR. I feel like it's worth reaching to implement these upgrades before releasing the Mining feature, as their omission might feel like a disappointment to players. At the very least, I'd want to make sure there's no in-game indication of a tool upgrade that isn't actually available. In other words, these string shouldn't be exposed until the upgrades are implemented (I'm not sure if that's the case or not right now).

src/strings.js
```diff
@@ -31,3 +31,37 @@ export const HOE_HIDDEN_TEXT =
...
+    GOLD: '**Gold**: Keeps plots watered for five days.',
+  },
+  HOE: {
+    DEFAULT: '**Basic**: Destroys crops.',
+    BRONZE: '**Bronze**: 20% chance to retrieve seeds when destroying crops.',
+    IRON: '**Iron**: 40% chance to retrieve seeds when destroying crops.',
+    SILVER: '**Silver**: 60% chance to retrieve seeds when destroying crops.',
+    GOLD: '**Gold**: 80% chance to retrieve seeds when destroying crops.',
+  },
+}
```
jeremyckahn, line 64: FYI @lstebner, I tweaked these strings a bit in https://github.com/jeremyckahn/farmhand/commit/39bea985894418b7b000b3f5e6afc4beb3a22e61.

Screenshot from 2021-09-29 08-39-05

src/Farmhand.js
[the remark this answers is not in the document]
lstebner: just made the update to add `toolType.UNAVAILABLE` and it reads nicely!

src/strings.js
lstebner: these strings are not exposed anywhere yet, so that part is true for now. i'm fine with getting these in before going live with the mining feature in general. i added them as ideas to the project board for tracking separately, will work on that as soon as we get done with this PR

src/utils/isRandomChance.js
lstebner: aye! `I'll have to keep looking at the code's implementation to know what it does.` i don't want that 😄 . so i've updated to `isRandomNumberLessThan` now.

src/strings.js
lstebner: nice! good idea

src/utils/isRandomChance.js
jeremyckahn: Awesome, thank for being flexible with the name! The code reads much more easily to me now. 😄

src/strings.js
jeremyckahn: Sounds like a great plan to me!

Issue #199, opened 2021-09-03
visual indicator that ore was dug up when using shovel


lstebner:
add some sort of animation to the field when an ore is dug up

PR #212, opened 2021-09-12
Mining animation

jeremyckahn:
### What this PR does

For #199, this PR adds an animation to indicate what the user has just dug up.

A lot of the changes in this PR are part of the ongoing effort to adopt React Testing Library. Rather than updating all of the old test files related to this change, this PR adds `Plot.rtl.test.js` for React Testing Library-specific changes.

Go to https://farmhand-git-jeremyckahn-199mining-animation-jeremyckahn.vercel.app/?enable_MINING=true, unlock the shovel, and dig up some items from the field.

### Additional information

https://user-images.githubusercontent.com/366330/133005441-ed90de71-5498-435d-8c3f-26acea8af377.mp4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@jeremyckahn/farmhand",
-      "version": "1.10.19",
+      "version": "1.10.28",
       "license": "CC BY-NC-SA 4.0",
       "dependencies": {
         "@fortawesome/fontawesome-svg-core": "^1.2.27",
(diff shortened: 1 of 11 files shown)


lstebner (approved):
this is great! i really don't see anything that needs changed here, and i tested it for quite a bit and didn't encounter any issues. if i had to mention _anything_ it would be the variable `plotStartedShoveled` but i don't have any better suggestion, it's just a mouthful hah. 🚀 !

jeremyckahn:
> if i had to mention _anything_ it would be the variable `plotStartedShoveled` but i don't have any better suggestion, it's just a mouthful hah. 🚀 !

Yeah, that is a bit of an awkward name. I changed it to `initialIsShoveledState` in https://github.com/jeremyckahn/farmhand/commit/3ad97086c6a4207053f81b4428f85b2de214b47d, which is actually a bit longer, but at least uses somewhat more standard semantics!

PR #224, opened 2021-12-13
Make piskel-cli optional

jeremyckahn:
### What this PR does

This PR fixes breaking builds in Vercel, which we need for deploys. I'm not really sure what changed, but it seems to be something on Vercel's side. `piskel-cli` is an optional development dependency, and it sub-depends on PhantomJS. PhantomJS has long been a source of pain when installing Farmhand. It tends to not install correctly on a lot of environments, which causes the entire project installation to fail.

This PR should hopefully provide a long term fix by moving `piskel-cli` to the [`optionalDependencies` list](https://betterprogramming.pub/what-are-npms-optional-dependencies-and-when-should-we-use-them-796a6a964e73).

I think this passing build should sufficiently validate this change: https://vercel.com/jeremyckahn/farmhand/HTdUa4tjXEvQAnT9Tq9DkKWqTNVt

You can also do `rm -rf node_modules && npm ci --no-optional && ls node_modules/piskel-cli/` to see something like:

```
ls: cannot access 'node_modules/piskel-cli/': No such file or directory
```

Maybe there's a better way to do this? I'd like to drop PhantomJS altogether, but https://github.com/jordanjwatkins/piskel-cli seems to be pretty dependent upon it and hasn't seen any updates in a few years.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,6 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 14
-      - run: npm ci
+      - run: npm ci --no-optional
       - run: npm test
       - run: npm run build
--- a/.github/workflows/run-release.yml
+++ b/.github/workflows/run-release.yml
@@ -29,7 +29,7 @@ jobs:
           node-version: 14
           registry-url: 'https://registry.npmjs.org'
 
-      - run: npm ci
+      - run: npm ci --no-optional
 
       - name: Tag release
         run: npm version ${{ github.event.inputs.releaseType }}
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ nvm i
 npm ci
 ```
 
-If `npm ci` errors out due to PhantomJS installation errors (this has been seen in some WSL/Linux environments), try `npm_config_tmp=/tmp npm ci` instead. [See this related comment](https://github.com/yarnpkg/yarn/issues/1016#issuecomment-283067214).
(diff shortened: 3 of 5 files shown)


jeremyckahn:
`piskel-cli` is strictly a development tool. It's called by this npm script: https://github.com/jeremyckahn/farmhand/blob/33c478c3729a6eeb8d29a638667fa193527b65fb/package.json#L26

Which invokes this script: https://github.com/jeremyckahn/farmhand/blob/33c478c3729a6eeb8d29a638667fa193527b65fb/update-images.sh#L3-L9

That script calls `piskel-cli` on every `.piskel` file that is either new or changed since the previous Git commit, which updates the associated `.png` image file. It's just a handy bit of automation for generating PNG files from Piskel source files, but it's not necessary in development. I imagine I'm the only person who's actually used it. :sweat_smile: 

>maybe we don't need it at all and it can just be installed one off for anyone who uses those features over the GUI ones?

We could remove it, but I do appreciate the workflows it automates (if even just for me). Moving it to `optionalDependencies` should keep it out more out of the way than it is now, but unfortunately npm `optionalDependencies` are opt-out, not opt-in. It would be ideal if that was the other way around, but `npm ci --no-optional` will at least skip it for anyone who specifically doesn't want it.

I suppose the optimal setup would be to just have `.piskel` source files and only ever generate `.png`s on the fly by the development server and by the build process at deployment time, but that seems a little heavy-handed to me.

lstebner:
thank you for the thorough explanation @jeremyckahn ! that makes sense, and sounds quite handy. I definitely hand-saved many files that I would have used this for if I knew it existed. i'm good with approving based on all this!

lstebner (approved):

PR #232, opened 2021-12-21
Update Trystero to 0.10.0

jeremyckahn:
### What this PR does

This PR updates Trystero, the library used for peer-to-peer to communication, to it's latest version (0.10.0). Trystero has had some significant updates recently: https://github.com/dmotz/trystero/commits/main

I've noticed some flakiness with peer connections lately, so hopefully this will help.

Go online successfully in the game.

Preview URL: https://farmhand-git-jeremyckahn-update-trystero-jeremyckahn.vercel.app/

There isn't a great way to test this without putting it into production, really. I think we just have to ship it and hope for the best. :grimacing:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@jeremyckahn/farmhand",
-      "version": "1.10.30",
+      "version": "1.11.3",
       "license": "CC BY-NC-SA 4.0",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


lstebner:
> There isn't a great way to test this without putting it into production, really. I think we just have to ship it and hope for the best. 

ah the joy of package dependencies!

lstebner (approved):

lstebner:
i was able to successfully go online on the preview URL so that seems like it's good to go to me! as you said, not sure what else we can really do here.

PR #238, opened 2022-01-01
feat:[closes #234] Hide Home Screen option

### LINKED ISSUE #234: Add a setting to hide the Home screen
Players don't have much use for the Home screen after the start of the game, so there should be an option to hide it.

**Acceptance criteria**

  - There is a new toggle option in the Settings view to show the Home screen that is on (enabled) by default. 
  - When the toggle is disabled:
    - The Home screen no longer appears in the view selector.
    - The Home screen is not accessible via the left/right bottom navigation buttons. 
    - The Home screen is not accessible via the "1" keyboard shortcut. All other view keyboard shortcuts are decremented by 1, so Shop goes from "2" to "1," Field goes from "3" to "2," etc.
  - When the toggle is enabled (current, default behavior):
    - The Home screen appears in the view selector as before. 
    - The Home screen is accessible via the left/right bottom navigation buttons.
    - The Home screen is accessible via the "1" keyboard shortcut. Shop is accessible via "2," Field is accessible via "3," etc.
  - The state of the new toggle option should be persisted to the saved game state.

Where the new option should go

View selector

UnquietBones:
### What this PR does

This adds an option to the Settings page to hide the Home screen. It is off by default and changes persist into the saved game state.

When the option is toggled on, HOME is removed as a navigation option and if the user is on the Home page it will move them to the next view (which is currently the SHOP).

When the page is first loaded or reloaded with the Home view hidden, the program will move to the next view in the list as above (is not hardcoded to SHOP).

Manual testing by opening settings and toggling the option.

If you toggle it off and are shifted over to the SHOP, it will not shift you back to HOME if you toggle it off again. Not sure if users will care?

### Additional information

234 Settings Screenshot with new option

### DIFF
--- a/src/Farmhand.js
+++ b/src/Farmhand.js
@@ -236,6 +236,7 @@ const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
  * their respective quantities.
  * @property {boolean} useAlternateEndDayButtonPosition Option to display the
  * Bed button on the left side of the screen.
+ * @property {boolean} hideHomeScreen Option to hide the Home Screen
  * @property {Object.<number>} valueAdjustments
  * @property {string} version Comes from the `version` property in
  * package.json.
@@ -299,7 +300,11 @@ export default class Farmhand extends Component {
 
   get viewList() {
     const { COW_PEN, FIELD, HOME, WORKSHOP, SHOP } = stageFocusType
-    const viewList = [HOME, SHOP, FIELD]
+    const viewList = [SHOP, FIELD]
+
+    if (!this.state.hideHomeScreen) {
+      viewList.unshift(HOME)
+    }
 
     if (this.state.purchasedCowPen) {
       viewList.push(COW_PEN)
@@ -403,6 +408,7 @@ export default class Farmhand extends Component {
         [toolType.WATERING_CAN]: toolLevel.DEFAULT,
       },
       useAlternateEndDayButtonPosition: false,
+      hideHomeScreen: false,
       valueAdjustments: {},
       version: process.env.REACT_APP_VERSION,
     }
--- a/src/Farmhand.js
+++ b/src/Farmhand.js
@@ -225,6 +225,7 @@ const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
  * @property {number} revenue The amount of money the player has generated in
  * @property {string} redirect Transient value used to drive router redirection.
  * @property {string} room What online room the player is in.
+ * @property {boolean} showHomeScreen Option to show the Home Screen
  * @property {boolean} showNotifications
  * @property {farmhand.module:enums.stageFocusType} stageFocus
  * @property {Array.<farmhand.notification>} todaysNotifications
@@ -236,7 +237,6 @@ const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
  * their respective quantities.
  * @property {boolean} useAlternateEndDayButtonPosition Option to display the
  * Bed button on the left side of the screen.
- * @property {boolean} hideHomeScreen Option to hide the Home Screen
  * @property {Object.<number>} valueAdjustments
  * @property {string} version Comes from the `version` property in
  * package.json.
@@ -302,7 +302,7 @@ export default class Farmhand extends Component {
     const { COW_PEN, FIELD, HOME, WORKSHOP, SHOP } = stageFocusType
     const viewList = [SHOP, FIELD]
 
-    if (!this.state.hideHomeScreen) {
+    if (this.state.showHomeScreen) {
       viewList.unshift(HOME)
     }
 
@@ -394,6 +394,7 @@ export default class Farmhand extends Component {
       purchasedCowPen: 0,
       purchasedField: 0,
       purchasedSmelter: 0,
+      showHomeScreen: true,
       showNotifications: true,
       stageFocus: stageFocusType.HOME,
       todaysNotifications: [],
@@ -408,7 +409,6 @@ export default class Farmhand extends Component {
         [toolType.WATERING_CAN]: toolLevel.DEFAULT,
       },
       useAlternateEndDayButtonPosition: false,
-      hideHomeScreen: false,
       valueAdjustments: {},
       version: process.env.REACT_APP_VERSION,
     }
--- a/src/Farmhand.js
+++ b/src/Farmhand.js
@@ -555,6 +555,13 @@ export default class Farmhand extends Component {
           }
         })
       })
+
+      if (
+        !this.state.showHomeScreen &&
+        this.state.stageFocus === stageFocusType.HOME
+      ) {
+        this.focusNextView()
+      }
     } else {
       // Initialize new game
       this.incrementDay(true)
--- a/src/Farmhand.js
+++ b/src/Farmhand.js
@@ -555,13 +555,6 @@ export default class Farmhand extends Component {
           }
         })
       })
-
-      if (
-        !this.state.showHomeScreen &&
-        this.state.stageFocus === stageFocusType.HOME
-      ) {
-        this.focusNextView()
-      }
     } else {
       // Initialize new game
       this.incrementDay(true)
--- a/src/components/SettingsView/SettingsView.js
+++ b/src/components/SettingsView/SettingsView.js
@@ -25,8 +25,10 @@ const SettingsView = ({
   handleSaveButtonClick,
   handleShowNotificationsChange,
   handleUseAlternateEndDayButtonPositionChange,
+  handleHideHomeScreenChange,
   showNotifications,
   useAlternateEndDayButtonPosition,
+  hideHomeScreen
 }) => {
   const [isClearDataDialogOpen, setIsClearDataDialogOpen] = useState(false)
 
@@ -69,6 +71,17 @@ const SettingsView = ({
               />
             }
             label="Show new notifications"
+          />
+          <FormControlLabel
+              control={
+                <Switch
+                    color="primary"
+                    checked={hideHomeScreen}
+                    onChange={handleHideHomeScreenChange}
+                    name="hide-home-screen"
+                />
+              }
+              label="Hide Home Screen"
           />
         </FormGroup>
       </FormControl>
@@ -175,8 +188,10 @@ SettingsView.propTypes = {
   handleSaveButtonClick: func.isRequired,
   handleShowNotificationsChange: func.isRequired,
   handleUseAlternateEndDayButtonPositionChange: func.isRequired,
+  handleHideHomeScreenChange: func.isRequired,
   showNotifications: bool.isRequired,
   useAlternateEndDayButtonPosition: bool.isRequired,
+  hideHomeScreen: bool.isRequired
 }
 
 export default function Consumer(props) {
--- a/src/components/SettingsView/SettingsView.js
+++ b/src/components/SettingsView/SettingsView.js
@@ -25,10 +25,10 @@ const SettingsView = ({
   handleSaveButtonClick,
   handleShowNotificationsChange,
   handleUseAlternateEndDayButtonPositionChange,
-  handleHideHomeScreenChange,
+  handleShowHomeScreenChange,
   showNotifications,
   useAlternateEndDayButtonPosition,
-  hideHomeScreen
+  showHomeScreen,
 }) => {
   const [isClearDataDialogOpen, setIsClearDataDialogOpen] = useState(false)
 
@@ -73,15 +73,15 @@ const SettingsView = ({
             label="Show new notifications"
           />
           <FormControlLabel
-              control={
-                <Switch
-                    color="primary"
-                    checked={hideHomeScreen}
-                    onChange={handleHideHomeScreenChange}
-                    name="hide-home-screen"
-                />
-              }
-              label="Hide Home Screen"
+            control={
+              <Switch
+                color="primary"
+                checked={showHomeScreen}
+                onChange={handleShowHomeScreenChange}
+                name="show-home-screen"
+              />
+            }
+            label="Show the Home Screen"
           />
         </FormGroup>
       </FormControl>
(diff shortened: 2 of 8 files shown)


src/Farmhand.js
```diff
@@ -403,6 +408,7 @@ export default class Farmhand extends Component {
         [toolType.WATERING_CAN]: toolLevel.DEFAULT,
       },
       useAlternateEndDayButtonPosition: false,
+      hideHomeScreen: false,
```
jeremyckahn, line 411: I feel that the semantics for the Settings view should indicate positive/affirmative states ("on" instead of "off", "yes" instead of "no"), as it creates a more consistent user experience. Considering that, would you mind changing `hideHomeScreen: false` to `showHomeScreen: true` and updating the relevant logic in this PR to match?

Additionally, please move this to property to its alphabetical position in the object.

src/event-handlers.js
```diff
@@ -399,6 +399,17 @@ export default {
     this.setState({ useAlternateEndDayButtonPosition })
   },
 
+  handleHideHomeScreenChange(
+      _e,
+      hideHomeScreen
+  ) {
+    if (this.state.stageFocus === 'HOME') {
```
lstebner, line 406: instead of the string 'HOME' here, i think you'll want to use `stageFocusType.HOME` from enums instead. if you look into the `Farmhand.js` file you can see it used there, but just let me know if you don't find it

lstebner:
every time I refresh, I still end up on the home screen even though i have chosen to hide it. i'm actually not sure what the best UX here is, so I'm going to defer to @jeremyckahn to say if that's ok, or if it would be better to put the person on the shop on page load if the home screen has been hidden

src/components/SettingsView/SettingsView.js
```diff
@@ -70,6 +72,17 @@ const SettingsView = ({
...
+              control={
+                <Switch
+                    color="primary"
+                    checked={hideHomeScreen}
+                    onChange={handleHideHomeScreenChange}
+                    name="hide-home-screen"
+                />
+              }
+              label="Hide Home Screen"
+          />
```
jeremyckahn, line 85: As mentioned in my comment in `Farmhand.js`, I feel this should use "show" semantics instead of "hide." Also, it looks like this is using inconsistent indentation with the rest of the file. Running `npx prettier 'src/components/SettingsView/SettingsView.js' --write` should be all we need!

src/constants.js
```diff
@@ -160,6 +160,7 @@ export const PERSISTED_STATE_KEYS = [
   'todaysStartingInventory',
   'toolLevels',
   'useAlternateEndDayButtonPosition',
+  'hideHomeScreen',
```
jeremyckahn, line 163: Yay for persisting this setting! Just for the sake of clarity, this property name should be updated to `showHomeScreen` as well.

src/event-handlers.js
```diff
@@ -399,6 +399,17 @@ export default {
...
+      this.focusNextView()
+    }
+
+    this.setState({ hideHomeScreen })
+  },
```
jeremyckahn, line 411: Though this logic is fairly simple, I think it's worth adding some unit testing for this event handler in case it ever becomes more complex.

jeremyckahn (changes requested):
Great work so far! This is off to a really strong start. I left some comments inline, but I also noticed some interesting behavior: When the new setting is changed from its default state and the game is saved, a refresh brings the user back to the Home screen. They can navigate away and not back, which is good. Was this behavior expected? It's not what I was anticipating but I'm open to letting it stay if you'd prefer to keep this behavior.

>If you toggle it off and are shifted over to the SHOP, it will not shift you back to HOME if you toggle it off again. Not sure if users will care?

I think this behavior is totally fine. Thanks for checking!

jeremyckahn:
> every time I refresh, I still end up on the home screen even though i have chosen to hide it. i'm actually not sure what the best UX here is, so I'm going to defer to @jeremyckahn to say if that's ok, or if it would be better to put the person on the shop on page load if the home screen has been hidden

I noticed this and commented as well (before seeing this comment...whoops!). If @UnquietBones prefers it I'm willing to run with it. We can always change the behavior later if players don't like it!

jeremyckahn:
I'm noticing this in the browser console on page reloads when the new setting is changed from its default value. Even if we stick with the behavior of dropping the player into the Home view when the page is refreshed, we should make sure there are no console errors or warnings.

Console warnings

src/event-handlers.js
UnquietBones: Ack yeah, I should have caught that sorry. 🤦‍♀️ Fixed!

src/Farmhand.js
```diff
@@ -236,6 +236,7 @@ const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
  * their respective quantities.
  * @property {boolean} useAlternateEndDayButtonPosition Option to display the
  * Bed button on the left side of the screen.
+ * @property {boolean} hideHomeScreen Option to hide the Home Screen
```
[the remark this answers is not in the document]
UnquietBones: I didn't notice it was alphabetical, sorry, fixed and I'll keep that in mind going forward! :)

src/Farmhand.js
UnquietBones: Will do and I'll make a note so I don't forget if there are more settings added later. :) Fixed and alphabetized!

src/components/SettingsView/SettingsView.js
UnquietBones: Darnit, sorry IntelliJ hijacks the formatting sometimes, I'd thought I'd fixed them all. Done and done!

src/constants.js
UnquietBones: Done :)

UnquietBones:
> I'm noticing this in the browser console on page reloads when the new setting is changed from its default value. Even if we stick with the behavior of dropping the player into the Home view when the page is refreshed, we should make sure there are no console errors or warnings.

I was getting those earlier, but I thought I'd I cleared them out... apparently not! I'll have to go dig into it and see why they're popping up again.

src/event-handlers.js
UnquietBones: I will get these done tomorrow as I haven't written Javascript unit tests before so I need to go inhale some knowledge before I stab wildly into the dark. 😅

src/components/SettingsView/SettingsView.js
UnquietBones: I think I have IntelliJ setup right now, but if anyone has a CodeStyles I can borrow to make sure, that would be awesome.

src/event-handlers.js
jeremyckahn: Sounds good! Let us know if you have any questions. IMO the project's current testing situation is decent, if a little hit-or-miss. There are some odd aspects of how a few things are set up related to mocks, but I don't think you'll run into much (if any) of that in this PR.

(FWIW, I take full responsibility for the more subpar aspects of Farmhand's testing suite. @lstebner has improved it considerably overall since he joined the project! 🙂)

src/components/SettingsView/SettingsView.js
jeremyckahn: I'm not a IntelliJ user, so I'm not familiar with CodeStyles (I'm guessing you're referring to [this](https://www.jetbrains.com/help/idea/configuring-code-style.html#create-copy)). I think you might be the first one on this project! If you believe a CodeStyles configuration would helpful for other IntelliJ users, please feel free to add one in a new PR. 

In the meantime, we do have some automation around code linting and formatting. Specifically, we have this Git hook: https://github.com/jeremyckahn/farmhand/blob/32648fc0fe3d6e499e2898cda232e0cbd2a68b49/package.json#L118-L120

However, that only runs when you're on the `develop` branch. I made that the case so frequent commits on a local topic branch wouldn't get bogged down by a heavy pre-commit hook and instead let [the CI system](https://github.com/jeremyckahn/farmhand/actions/workflows/ci.yml) do it asynchronously. So, the formatting wasn't automatically cleaned up here.

I'm realizing now that linting and formatting should happen for commits on all branches, but only the test runner part of the hook should be limited to the local `develop` branch. I can make that change in a separate PR (and also clean up a variety of unrelated formatting violations) so that these sorts of things can be automated for all going forward!

jeremyckahn: #240 has been merged and should prevent formatting violations from getting committed going forward. Thank you for helping us find this misconfiguration @UnquietBones!

UnquietBones:
I'm still chasing down the out-of-range bug, but I've added a first pass at some unit tests and a fix for the reload/first load issues where the Home screen was still displaying.

Putting PR on hold until I get those sorted out. Sorry!

src/event-handlers.test.js
```diff
@@ -168,3 +168,48 @@ describe('handleCowSelect', () => {
...
+
+  test('change show home screen setting to True and does not navigate', () => {
+    component.state().stageFocus = stageFocusType.SHOP
+    component.state().showHomeScreen = false
+
+    handlers().handleShowHomeScreenChange(null, true)
+    expect(component.instance().setState).toHaveBeenCalled()
+    expect(component.instance().focusNextView).not.toHaveBeenCalled()
+  })
+})
```
jeremyckahn, line 215: Sorry for just dropping a big diff here, but I spent some time with this and this and this change seems to get things going:

```diff
diff --git a/src/event-handlers.test.js b/src/event-handlers.test.js
index 1d97569..b99b98a 100644
--- a/src/event-handlers.test.js
+++ b/src/event-handlers.test.js
@@ -170,46 +170,48 @@ describe('handleCowSelect', () => {
 })
 
 describe('handleShowHomeScreenChange', () => {
-
   beforeEach(() => {
-    jest.spyOn(component.instance(), 'setState').mockImplementation()
     jest.spyOn(component.instance(), 'focusNextView').mockImplementation()
   })
 
   test('change show home screen setting to False and navigate to the next view', () => {
-    component.state().stageFocus = stageFocusType.HOME
-    component.state().showHomeScreen = true
+    component.setState({
+      stageFocus: stageFocusType.HOME,
+      showHomeScreen: true,
+    })
 
     handlers().handleShowHomeScreenChange(null, false)
-    expect(component.instance().setState).toHaveBeenCalled()
-    // expect(component.showHomeScreen).toBeFalse()
+    expect(component.state().showHomeScreen).toBeFalse()
     expect(component.instance().focusNextView).toHaveBeenCalled()
   })
 
   test('change show home screen setting to False and do not navigate', () => {
-    component.state().stageFocus = stageFocusType.SHOP
-    component.state().showHomeScreen = true
+    component.setState({
+      stageFocus: stageFocusType.SHOP,
+      showHomeScreen: true,
+    })
 
     handlers().handleShowHomeScreenChange(null, false)
-    expect(component.instance().setState).toHaveBeenCalled()
     expect(component.instance().focusNextView).not.toHaveBeenCalled()
   })
 
   test('change show home screen setting to True and does not navigate', () => {
-    component.state().stageFocus = stageFocusType.HOME
-    component.state().showHomeScreen = false
+    component.setState({
+      stageFocus: stageFocusType.HOME,
+      showHomeScreen: false,
+    })
 
     handlers().handleShowHomeScreenChange(null, true)
-    expect(component.instance().setState).toHaveBeenCalled()
     expect(component.instance().focusNextView).not.toHaveBeenCalled()
   })
 
   test('change show home screen setting to True and does not navigate', () => {
-    component.state().stageFocus = stageFocusType.SHOP
-    component.state().showHomeScreen = false
+    component.setState({
+      stageFocus: stageFocusType.SHOP,
+      showHomeScreen: false,
+    })
 
     handlers().handleShowHomeScreenChange(null, true)
-    expect(component.instance().setState).toHaveBeenCalled()
     expect(component.instance().focusNextView).not.toHaveBeenCalled()
   })
 })
```

Basically, I think we need to drop `jest.spyOn(component.instance(), 'setState').mockImplementation()` in the `beforeEach` (because we want to use it as-is), use `setState` to set up desired starting states to test, and drop the `expect(component.instance().setState).toHaveBeenCalled()` assertions.

I think this could use some additional `expect(component.state().showHomeScreen).toBeFalse()`/`toBeTrue()`'s as necessary to round it out, but the diff above should help to get things into a better place to work from at least. Let me know if you have additional questions about this!

UnquietBones: That cleared it all up, thanks! :)

UnquietBones:
Once more into the breach! Should be all sorted out now, the last of the console errors are gone and the tests appear to be running as they should. Please let me know if anything else pops up. 🤔

jeremyckahn (changes requested):
Awesome progress here! I noticed a need for a minor logic tweak (in my own proposed code... :man_facepalming:), and I wanted to get others' thoughts on alternative testing strategy.

src/event-handlers.test.js
```diff
@@ -168,3 +168,53 @@ describe('handleCowSelect', () => {
     expect(component.instance().selectCow).toHaveBeenCalledWith({ id: 'abc' })
   })
 })
+
+describe('handleShowHomeScreenChange', () => {
+  beforeEach(() => {
+    jest.spyOn(component.instance(), 'focusNextView').mockImplementation()
+  })
```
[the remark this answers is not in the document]
lstebner: 💯 for testing against outcomes. that's something I also converted over to just in the last year, and it's made a _huge_ difference for me. I find that I write less tests, but get more rigidity from them.

src/utils.js
```diff
@@ -997,6 +998,10 @@ export const transformStateDataForImport = state => {
     sanitizedState.toolLevels = unlockTool(sanitizedState.toolLevels, tool)
   }
 
+  if (!sanitizedState.showHomeScreen) {
```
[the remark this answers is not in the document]
UnquietBones: Whoops. completely missed this! Fixed! :)

src/event-handlers.test.js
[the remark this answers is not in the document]
UnquietBones: Done! :)

jeremyckahn (approved):
Amazing work! Well done @UnquietBones. Thanks for sticking with this and being open to change suggestions. I'll merge and ship this now!

Issue #236, opened 2021-12-31
Ranged items should not affect action range slider state


jeremyckahn:
Per [u/diiejso on Reddit](https://www.reddit.com/r/incremental_games/comments/rlfw3u/comment/hqpi7f4/?utm_source=reddit&utm_medium=web2x&context=3)

>I found a bug then and figured out how to reproduce it:
>
>    Method 1 (works properly): New game, buy 1 scarecrow and max carrot seeds. Place seeds in field first and the slider is at 1x1 as you stated.
>
>    Method 2 (produce bug): New game, buy 1 scarecrow and max carrot seeds. Place scarecrow first, and then start to place the carrot seeds. The action slider is now at the max size.
>
>    Method 3 (produce bug): New game, buy 1 scarecrow and max carrot seeds. Place some carrots and the action size is 1x1, place a scarecrow, now place some carrots again and the action size is maxed out.
>
>It appears that placing the scarecrow will automatically adjust the action size to the max value.
>
>OS:
>
>    Ubuntu 21.10
>
>Browsers Tested:
>
>    Firefox 95.0.2 snap (64-bit)
>
>    Chromium 96.0.4664.110 (Official Build) snap (64-bit)

Additionally:

> Buying and placing a sprinkler sets your action size to 3x3 and leaves it there, probably related to the scarecrow thing from before.

This occurs because the Scarecrow and Sprinklers set the action range slider state to a specific value each time they are selected (by design). These items should have their own range logic and not affect (or be affected by) the general range slider state.

PR #241, opened 2022-01-05
Decouple static ranges from slider

jeremyckahn:
### What this PR does

This PR implements #236. It's a UX improvement that prevents the player's field range settings from being changed by selecting the Scarecrow or the Sprinkler. It also improves how field ranges are displayed.

Use different items and tools at different ranges, go back and forth between selecting the Scarecrow and Sprinkler items, and see if the interaction feels intuitive.

  - How well do the range colors work? Anything else that should change with the display of the ranges?
  - Does this new UX of the different items having different ranges feel good?

#### Planting a seed
Planting a seed

#### Watering Can selected
Watering Can selected

#### Hoe selected
Hoe selected

#### Scythe selected
Scythe selected

#### Shovel selected
Shovel selected

#### Scarecrow selected
Scarecrow selected

#### Sprinkler selected

Sprinkler selected

### DIFF
--- a/src/components/Field/Field.js
+++ b/src/components/Field/Field.js
@@ -18,7 +18,13 @@ import Plot from '../Plot'
 import QuickSelect from '../QuickSelect'
 import { fieldMode } from '../../enums'
 import tools from '../../data/tools'
-import { doesInventorySpaceRemain, nullArray } from '../../utils'
+import {
+  doesInventorySpaceRemain,
+  farmProductsSold,
+  getLevelEntitlements,
+  levelAchieved,
+  nullArray,
+} from '../../utils'
 
 import './Field.sass'
 
@@ -27,6 +33,7 @@ const {
   FERTILIZE,
   HARVEST,
   MINE,
+  OBSERVE,
   PLANT,
   SET_SCARECROW,
   SET_SPRINKLER,
@@ -49,19 +56,40 @@ if (tools.shovel) {
 }
 
 export const isInHoverRange = ({
+  fieldMode,
   hoveredPlotRangeSize,
   hoveredPlot: { x: hoveredPlotX, y: hoveredPlotY },
+  itemsSold,
   x,
   y,
 }) => {
   // If hoveredPlotX is null, assume that hoveredPlotY is as well.
-  if (hoveredPlotX == null) {
+  // If fieldMode === OBSERVE, nothing is in hover range.
+  if (hoveredPlotX == null || fieldMode === OBSERVE) {
     return false
   }
 
-  const squareSize = 2 * hoveredPlotRangeSize
-  const rangeFloorX = hoveredPlotX - hoveredPlotRangeSize
-  const rangeFloorY = hoveredPlotY - hoveredPlotRangeSize
+  let hoveredPlotRangeSizeToRender = hoveredPlotRangeSize
+
+  switch (fieldMode) {
+    case SET_SPRINKLER:
+      hoveredPlotRangeSizeToRender = getLevelEntitlements(
+        levelAchieved(farmProductsSold(itemsSold))
+      ).sprinklerRange
+
+      break
+
+    case SET_SCARECROW:
+      hoveredPlotRangeSizeToRender = Number.MAX_SAFE_INTEGER
+
+      break
+
+    default:
+  }
+
+  const squareSize = 2 * hoveredPlotRangeSizeToRender
+  const rangeFloorX = hoveredPlotX - hoveredPlotRangeSizeToRender
+  const rangeFloorY = hoveredPlotY - hoveredPlotRangeSizeToRender
   const rangeCeilingX = rangeFloorX + squareSize
   const rangeCeilingY = rangeFloorY + squareSize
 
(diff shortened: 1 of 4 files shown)


src/components/Plot/Plot.sass
```diff
@@ -36,8 +45,11 @@ $not-allowed-color: rgba(255, 0, 0, 0.5)
       cursor: pointer
       border-color: var(--color-yellow)
 
+    &.is-in-hover-range
+      border-color: var(--color-yellow)
+
   .cleanup-mode &.crop,
-    background-color: rgba(255, 255, 0, 0.5)
+    background-color: var(--color-yellow)
```
lstebner, line 52: this comment applies to this whole file and how these CSS vars are used because I was using them this same way up until just before the recent holiday break, but then came to realize that this is still a "sassy" approach to them. CSS vars are most powerful for dynamic values, for static uses a SASS var is ultimately more performant.

so, instead of having these CSS vars like `--color-yellow` and `--color-blue-water`, they could instead be one var like `--plot-background-color`, which gets redefined in certain situations. the value could however come from SASS vars, to really make the most of everything available to us.

```sass
// these might be in a different file
$color_yellow: rgba(255, 255, 0, 0.5)
$color_blue_water: rgba(125, 245, 255, 0.75)

.Plot
    background-color: var(--plot-background-color)

.water-mode &.crop:hover
    --plot-background-color: #{$color_blue_water}

.cleanup-mode &.crop,
    --plot-background-color: #{$color_yellow}
```

this is just an example with one property, though there's a few that I see could be updated. it's also a little bit of a tangent from the intention of this PR so I don't think it _has_ to change to merge, but I wanted to mention it while I was thinking of it. I'm also happy to help update this file if you want!

src/components/Field/Field.js
```diff
@@ -49,19 +56,40 @@ if (tools.shovel) {
...
   hoveredPlotRangeSize,
   hoveredPlot: { x: hoveredPlotX, y: hoveredPlotY },
+  itemsSold,
   x,
   y,
 }) => {
   // If hoveredPlotX is null, assume that hoveredPlotY is as well.
-  if (hoveredPlotX == null) {
+  // If fieldMode === OBSERVE, nothing is in hover range.
+  if (hoveredPlotX == null || fieldMode === OBSERVE) {
```
lstebner, line 68: what do you think about dropping the comments and changing this to a one liner. i think it would be:

`return hoveredPlotX != null && fieldMode !== OBSERVE`

then just reading the method and conditions tells me that fieldMode of 'observe' will never be in hover range.

lstebner (commented):
this is a fantastic change! left some comments, but good call working on this.

src/components/Field/Field.js
jeremyckahn: That would prevent the rest of the code in this function from running, which is the real crux of this function. This initial check just bails out in a few specific cases where the rest of the logic doesn't apply. So, I think this logic needs to stay as-is.

Let me know if I'm understanding your suggestion correctly!

src/components/Plot/Plot.sass
jeremyckahn: I see what you're saying! You have a great point; Sass variables make a lot more sense for static values like this. I made the change in https://github.com/jeremyckahn/farmhand/commit/9e7e84a2b761c9249e613721164d352973bad5ea. Is that what you had in mind?

src/components/Field/Field.test.js
```diff
@@ -153,29 +156,109 @@ describe('water-mode class', () => {
...
+        hoveredPlotRangeSize: 2,
+        hoveredPlot: { x: 1, y: 1 },
+        itemsSold: {},
+        x: 0,
+        y: 0,
+      })
+    ).toBe(true)
+  })
+
+  describe('fieldMode === OBSERVE', () => {
```
lstebner, line 183: can these instead be worded how the player would be thinking. i.e. maybe `describe("when observing the field")` and for the sprinkler one maybe `describe("when placing a sprinkler")`?

src/components/Plot/Plot.sass
```diff
@@ -1,65 +1,70 @@
...
+  &:hover
+    background-color: #{$colorGenericHighlight}
+    cursor: pointer
+
+  &.is-in-hover-range
+    border-color: #{$colorGenericHighlight}
+
   .water-mode &.crop:hover
-    background-color: var(--color-blue-hover)
+    background-color: #{$colorBlueWater}
```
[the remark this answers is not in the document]
jeremyckahn: Whoops, good catch! Silly me. This is fixed in https://github.com/jeremyckahn/farmhand/commit/65abdcd00a27fce9fdccee6c481657d7c73e0553!

src/components/Field/Field.test.js
jeremyckahn: Good idea! this is done in https://github.com/jeremyckahn/farmhand/commit/a23362cfe63e953c00b8a2ba7dff628ee2d72b21.

```
  isInHoverRange
    ✓ indicates when plot is not in hover range
    ✓ indicates when plot is in hover range (1 ms)
    when observing the field
      ✓ indicates when plot is not in hover range
      ✓ indicates when plot is in hover range
    when placing a sprinkler
      ✓ indicates when plot is not in hover range (1 ms)
      ✓ indicates when plot is in hover range
    when placing a scarecrow
      ✓ indicates that all plots are in hover range
```

This is much clearer! :+1:

jeremyckahn (commented):
Thanks for the helpful comments! I think the code is looking a lot better now.

lstebner (approved):
looks great! ship it!

Issue #248, opened 2022-01-27
Cow trading


jeremyckahn:
Allow cows to be traded between online peer players.

PR #249, opened 2022-01-27
#248: Refactor `CowCard`

jeremyckahn:
### What this PR does

This is a pure refactor PR that is meant to set up for the new functionality to come in #248. The goal is to pay down some tech debt in a noisy but straightforward diff so that the new functionality is easier to review in a subsequent PR. There is no new logic here, the code is simply moved from `CowPenContextMenu.js` to `CowCard.js` (tests and common utility functions have also been moved as necessary).

It also adds some `propTypes` that were previously missing.

Play with the cow cards in the menu a bit and ensure there are no regressions.

### Additional information

Additional cleanup and improvements could be made to this code, but for this PR I just wanted to focus on extracting the `CowCard` and `CowCardSubheader` components from the `CowPenContextMenu` file, as they didn't belong there to begin with.


src/Farmhand.js
```diff
@@ -163,7 +163,7 @@ const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
  * @property {Object.<string, boolean>} completedAchievements Keys are
  * achievement ids.
  * @property {farmhand.cow} cowForSale
- * @property {Array.<farmhand.cowBreedingPen>} cowBreedingPen
```
jeremyckahn, line 166: This was incorrect; it was likely a copy/paste error on my part.

src/components/CowCard/CowCard.js
```diff
@@ -0,0 +1,384 @@
...
+
+CowCardSubheader.propTypes = {
+  cow: object.isRequired,
+  cowBreedingPen: object.isRequired,
+  cowInventory: array.isRequired,
+  handleCowAutomaticHugChange: func,
+  handleCowBreedChange: func,
+  huggingMachinesRemain: bool.isRequired,
+  isCowPurchased: bool.isRequired,
+}
```
jeremyckahn, line 208: :arrow_up: Actual new code!

src/components/CowCard/CowCard.js
```diff
@@ -0,0 +1,384 @@
...
+  handleCowBreedChange: func,
+  handleCowHugClick: func,
+  handleCowNameInputChange: func,
+  handleCowPurchaseClick: func,
+  handleCowSellClick: func,
+  inventory: array.isRequired,
+  isSelected: bool,
+  money: number.isRequired,
+  purchasedCowPen: number.isRequired,
+}
```
jeremyckahn, line 384: :arrow_up: Actual new code!

jeremyckahn:
Oops, I forgot to move the styles as well. Working on that now!

src/components/CowCard/CowCard.js
lstebner: since you're moving things around, i would recommend getting every component into its own file. it can feel like a lot of overhead when creating them, but it's much easier to sort through down the road. i would look at the top level component more like a package, which has helpers and other components inside to create that thing. in this specific case it seems like `CowCard` is the top level one, and this component could be located at `CowCard/Subheader/Subheader.js` and there could be a `CowHeader/helpers.js` for things like `getCowMapById`, or that could be in `CowHeader/helpers/getCowMapById.js`, styles can be shared or local as well as needed. there's really no right or wrong way of course, but i think there are very few downsides to the extra files.

src/components/CowCard/CowCard.sass
```diff
@@ -0,0 +1,73 @@
...
+      border-radius: 50%
+      border: solid 1px #000
+      height: $indicator-size
+      margin-bottom: 0
+      margin-right: 0.5em
+      margin-top: 0
+      width: $indicator-size
+
+      &.blue
+        background: #8ff0f9
```
lstebner, line 37: would these colors get updated if the cow colors ever changed? they seem like they should be put into variables somewhere for maintainability

src/components/CowCard/CowCard.test.js
```diff
@@ -0,0 +1,386 @@
...
+    describe('cow is purchased', () => {
+      test('renders hearts', () => {
+        const { container } = render(
+          <CowCardSubheader {...{ ...baseProps, isCowPurchased: true }} />
+        )
+
+        expect(container.querySelectorAll('.heart')).toHaveLength(10)
+      })
+
+      test('renders full hearts that match cow happiness', () => {
```
lstebner, line 347: i would break this test up into a couple pieces. have expects -> render -> expects can cause for some whack a mole when things break. i would at least make two pieces here, since you could simply insert another `test` line after the first two expects, but i would also reconsider if they actually fit together. i see a "it renders all 10 hearts regardless of current happiness level", and also a "it renders full hearts to represent current happiness". i almost suggested combining the expects when i saw the last 2 because it seems like expecting .heart.is-full to have length of 10 covers both "are there 10 hearts and are they full?" but i think your intention was different. so creating the separate `test`'s will better express this.

to take it a step further, you could go with a loop that increments from 1-10 and then just write 2 `test`'s inside, one that makes sure there's always 10 hearts, and one that makes sure there are N full hearts.

jeremyckahn: > i would break this test up into a couple pieces. have expects -> render -> expects can cause for some whack a mole when things break. i would at least make two pieces here, since you could simply insert another test line after the first two expects

This is a great idea! Done in https://github.com/jeremyckahn/farmhand/commit/5f4fe0c5206fbbba36dd00ec4b978f8bde0d6e69. I also removed the two `expect(container.querySelectorAll('.heart')).toHaveLength(10)`s in this test as they seemed redundant given the expectation in the `'renders hearts'` test above.

> to take it a step further, you could go with a loop that increments from 1-10 and then just write 2 `test`'s inside, one that makes sure there's always 10 hearts, and one that makes sure there are N full hearts.

I was thinking about that too, but I generally try to avoid dynamic tests as they introduce their own complexity. They can be an appropriate solution, but in this case I don't think the value that dynamic tests would give us here would justify the complexity cost.

src/components/CowCard/Bloodline/Bloodline.sass
```diff
@@ -0,0 +1,38 @@
+@import ../../../styles/variables.sass
+
+.CowCard
```
lstebner, line 3: is it important to leave this wrapping selector?

lstebner (commented):
thanks for making all those changes! everything is looking great, I just came up with one more question i'm not sure about

src/components/CowCard/Bloodline/Bloodline.sass
jeremyckahn: I think so! The idea is to scope these subpackage styles so that other components can potentially have a `.Bloodline` class that doesn't conflict. Admittedly this is more of a relevant concern for the new `.Subheader` class, but I figured it would be good to stick to a convention here.

lstebner: gotcha. the issue I have with it is that Bloodline is now tied to only being usable inside CowCard, but I do also understand what you're attempting to prevent from happening. this one specific css file isn't a huge deal, but maybe implementing [CSS modules](https://css-tricks.com/css-modules-part-1-need/) is worth doing at some point to have a simpler way of scoping styles in the future.

lstebner (approved):

src/components/CowCard/Bloodline/Bloodline.sass
jeremyckahn: > Bloodline is now tied to only being usable inside CowCard

Yep, that's the intention here. My thinking is that if `Bloodline` was meant to be more generally usable, it should moved to its own component directory outside of the `CowCard` directory. And in that case, we would simply remove this nesting.

> but maybe implementing [CSS modules](https://css-tricks.com/css-modules-part-1-need/) is worth doing at some point to have a simpler way of scoping styles in the future.

I would be open to that! I haven't used them at all myself, but I know that it is The New Way of organizing CSS.

PR #258, opened 2022-03-03
Implement Cow trading

jeremyckahn:
### What this PR does

This is a really big PR... sorry. 😅

This PR implements a cow trading feature for online players (#248). With this feature, you can offer any of your cows for trade and initiate a trade with any other online player that is doing the same. 

When the trade is completed successfully, the game is saved for the player who initiated the trade (but not the other player to prevent game data corruption).

There are a few other supporting features to make this a reasonably complete package:

  - The ability to hide/show custom names from other players (foreign cows). Custom foreign cow names are hidden by default as an abuse countermeasure.
  - A new stat to track cow trades.
  - A redesigned "Active Players" modal.

1. In browser A, buy at at least one cow and offer it for trade
2. In browser B, buy at at least one cow and offer it for trade
3. In either browser, initiate the trade

The feature seems to work pretty well in my testing, but you'll notice that this PR doesn't add many unit tests. Development of this feature has dragged on for over a month, and to be honest I kind of just want to ship it and get it out to players... and move onto something else. :sweat_smile: 

If there's a part of this PR that seems like it really calls some automated testing before it should be shipped, let me know and I can get it done. I've just used up all my motivation to do it on my own volition at this point. :upside_down_face: 

Overall, this isn't a perfect PR. However, it makes the game materially better for players and it doesn't introduce any instability I'm aware of. I'd love to polish this up more and write more tests, but at some point you just gotta ship and accept a bit of technical debt in order to make some progress. That said, I think this is ready to ship! 🙂

### Additional information

image


jeremyckahn:
This is ready for another review!

I made some fixes and snuck in another minor feature via https://github.com/jeremyckahn/farmhand/commit/c933ed7bc5b0cbb3445733a252305628f8cf7c81: Cows received from others will now show the original owner's name!

image

src/Farmhand.js
```diff
@@ -1117,10 +1240,16 @@ export default class Farmhand extends Component {
       viewTitle,
     } = this
 
+    const blockInput =
```
lstebner, line 1243: not a must change, but this could be a getter instead. though if it will never applicable anywhere else it doesn't matter.

src/trystero-handlers.js
```diff
@@ -0,0 +1,250 @@
...
+          },
+        },
+      }
+    },
+    async () => {
+      if (!wasTradeSuccessful) return
+
+      // Allow any pending state updates to complete so that an invalid state
+      // is not persisted
+      await sleep(750)
```
lstebner, line 129: what's the reason it can be lagging behind so much? i feel obliged to ask if there's anything else that can be done here, though it's mostly just to raise the question. like, is there some sort of event or callback approach instead?

lstebner:
two things that came to mind while i was reviewing that apply to this as a whole... 

what do you think about adding a task to the backlog for a visual indicator on the cow itself when you have it marked available for trade. i'm thinking maybe like a little ribbon? like you see at county fairs, if you've been.. hah.

also, i'm wondering if "rescind" is the most common term to use here. is that what pokemon uses? i know you've based a lot of this on that, and i'm not super familiar.

lstebner (approved):
i had some very minor thoughts compared to what's been added here. i wouldn't have any concerns if you shipped this based on my playtesting, but leaving these few thoughts as well.

jeremyckahn:
> what do you think about adding a task to the backlog for a visual indicator on the cow itself when you have it marked available for trade. i'm thinking maybe like a little ribbon? like you see at county fairs, if you've been.. hah.

I like that idea! I've filed https://github.com/jeremyckahn/farmhand/issues/260 to track this.

> also, i'm wondering if "rescind" is the most common term to use here. is that what pokemon uses? i know you've based a lot of this on that, and i'm not super familiar.

That's a good point. I don't love "rescind" either. Thinking about it again, I think "withdraw" works better, so I'll go with that!

jeremyckahn (commented):
Thanks for the helpful review @lstebner! I made the changes you suggested that I was able to solve for. I'm going to merge and FINALLY ship this! 🚢🐮

Issue #269, opened 2022-04-02
Replace Enzyme with React Testing Library


jeremyckahn:
As part of a larger effort to update Farmhand's various dependencies, we need to remove Enzyme as it does not support the latest versions of React.

  - https://github.com/enzymejs/enzyme/issues/2429

PR #270, opened 2022-04-04
Cow selection testing

jeremyckahn:
### What this PR does

This PR starts on the long road that is #269. It replaces Enzyme-based tests that deal with cow selection with a React Testing Library-based approach that tests the outcomes from a user's perspective.

Additionally, this PR makes some project structure changes. It introduces `src/game-logic/`, a directory to house code that drives the overall game ("business logic," if you will) rather than a single given component. Reducers and event handlers have been moved into this directory, and it will likely contain many more tests as we shift to unifying on React Testing Library.

Run the tests (`npm run test`) and ensure the game boots.

I don't consider myself an expert on React Testing Library and would appreciate any feedback on how I can improve what I've added here!


src/components/CowPen/CowPen.js
```diff
@@ -209,7 +209,7 @@ export class Cow extends Component {
               {...{
                 src: cowImage,
               }}
-              alt="Cow"
+              alt={getCowDisplayName(cow, id, allowCustomPeerCowNames)}
```
jeremyckahn, line 212: I love that React Testing Library nudge us towards a more accessible UI! 😁

src/game-logic/README.md
```diff
@@ -0,0 +1,3 @@
+# Game Logic
+
+This directory contains general game logic that extends beyond the scope of a single component.
```
lstebner, line 3: i really like the idea of this directory! i don't know how much i love the move of event-handlers.js into here though. i think it's up to you ultimately as it is a design decision and there is no right or wrong, but i am not a fan of the single generic file containing many loosely related things. i would prefer to take a look at what all those event handlers are and see if there is another way to organize them. i also don't know everything in that file right now, but "event handlers" in a web app to me screams "DOM events". 

i think the tl;dr> is just.. I would like to think about how to organize this directory long term. 😄

src/game-logic/cows/cow-selection.test.js
```diff
@@ -0,0 +1,113 @@
...
+
+import { farmhandStub } from '../../test-utils/stubs/farmhandStub'
+import { generateCow, getCowDisplayName } from '../../utils'
+import { stageFocusType } from '../../enums'
+
+describe('cow selection', () => {
+  const cowId1 = 'foo'
+  const cowId2 = 'bar'
+  const cowStub1 = generateCow({ id: cowId1 })
+  const cowStub2 = generateCow({ id: cowId2 })
```
lstebner, line 12: can this be moved into the `beforeEach`? the tests in this file look safe enough but i find it's best to have things like this reset between each spec to avoid accidentally changing something in one test that sticks to the next

src/game-logic/cows/cow-selection.test.js
```diff
@@ -0,0 +1,113 @@
...
+    cow1 = null
+    cow2 = null
+
+    await farmhandStub({
+      cowInventory: [cowStub1, cowStub2],
+      stageFocus: stageFocusType.COW_PEN,
+      purchasedCowPen: 1,
+
+      // Usually this gets initialized during the boot process, but it needs to
+      // be stubbed to prevent a spurious crash
```
lstebner, line 40: this comment makes me want to fix that root problem, is it possible?

src/game-logic/cows/cow-selection.test.js
```diff
@@ -0,0 +1,113 @@
...
+  afterEach(() => {
+    // Allow all cow movement timers to expire
+    act(() => {
+      jest.advanceTimersByTime(2000)
+    })
+  })
+
+  test('no cows are selected by default', () => {
+    expect(cow1.classList).not.toContain('is-selected')
+    expect(cow2.classList).not.toContain('is-selected')
```
lstebner, line 58: i'm looking forward to when we have some text to select! i think it will reduce this to 1 expect which makes me happy 😸

src/game-logic/cows/cow-selection.test.js
```diff
@@ -0,0 +1,113 @@
...
+
+    act(() => {
+      // Prevent async update warning from Popper.js (used by Material UI
+      // tooltips). See "Case 2" here:
+      // https://davidwcai.medium.com/react-testing-library-and-the-not-wrapped-in-act-errors-491a5629193b
+      jest.advanceTimersByTime(500)
+    })
+
+    expect(cow1.classList).toContain('is-selected')
+    expect(cow2.classList).not.toContain('is-selected')
```
lstebner, line 72: i don't feel this line needs to be here. i get what you're thinking, but it feels a little misleading since in reality CowPen will have N number of items, and just testing it checking these 2 statically isn't the same to me. i could get behind an approach that uses a loop to check that of all rendered cows no more than 1 is ever selected, but being as your next test is the deselect version, the only expectation that matters to me is what happens to the one we clicked on in this test (so this note applies to that test as well). if this didn't make sense let me know cause i feel like i used a lot of words, hah

src/game-logic/cows/cow-selection.test.js
```diff
@@ -0,0 +1,113 @@
...
+  })
+
+  test('selecting another cow deselects the first one', () => {
+    userEvent.click(cow1)
+
+    act(() => {
+      jest.advanceTimersByTime(500)
```
lstebner, line 79: this is a long time, does it really take that long?

lstebner (approved):
so nice to see this conversion happening! i left some comments on things but I don't really need to block merge for any of them given that this is an ongoing effort. the biggest thing is what i mentioned in discord around using some hidden accessible text instead of classNames for the selected state. so feel free to address anything i mentioned in this PR or the followups. great job!

src/game-logic/README.md
jeremyckahn: That makes sense to me! I just pushed https://github.com/jeremyckahn/farmhand/commit/00a40f440e354b4e402c7d5c1f7f081c0092d22e, which moves the UI event and peer request handlers to `src/handlers/` and renames them to be a little clearer:

```
src/handlers/
├── peer-events.js
├── README.md
├── ui-events.js
└── ui-events.test.js
```

What do you think of this structure?

jeremyckahn (commented):
Thanks for the helpful feedback @lstebner! This is looking a lot better after implementing your suggestions. 🙂

PR #272, opened 2022-04-08
Lstebner/271 disable upgrade buttons

### LINKED ISSUE #271: Tool upgrades can be purchased without meeting ingredient requirements
Per [this bug report on Discord](https://discord.com/channels/714539345050075176/822068921653788683/961988509424549890):

> you can buy tool upgrades in the forge even if you dont have the materials for them

lstebner:
### What this PR does

fixes #271 by adding a 'disabled' attribute to the button in the `UpgradePurchase` component similar to how the others work

Unlock the recipes for the tool upgrades but then try purchasing them when you don't actually have enough materials. on the current main branch it will happily let you purchase! with this change the buttons will be disabled and will not allow the purchase to be made. the logic for if the player has the available resources was already there, it just wasn't being used to disable the button.

here's a save that has the shovel unlocked and a few recipes: 
[farmhand-save-early-access-to-upgrades.json.zip](https://github.com/jeremyckahn/farmhand/files/8453806/farmhand-save-early-access-to-upgrades.json.zip)

Disabling buttons in this type of way isn't actually accessible. It ends up "removing" them from the DOM tree and then screen readers don't know they're there. Long term, disabling buttons should be visual only, and we have to add logic to the click handlers to prevent the action when it's not allowed. i went with the disabled route for this PR to match how the other areas of the UI are currently working and patch the hole for the majority to start. accessibility can be a separate effort since we know there are other issues here and there.

### Additional information

video of the bug: https://www.dropbox.com/s/14ebdad8ssvqadw/early%20access%20to%20upgrades.mov?dl=0

and the fix with disabled buttons:

image

### DIFF
--- a/src/components/UpgradePurchase/UpgradePurchase.js
+++ b/src/components/UpgradePurchase/UpgradePurchase.js
@@ -65,7 +65,12 @@ export function UpgradePurchase({
       />
 
       <CardActions>
-        <Button color="primary" onClick={handleUpgrade} variant="contained">
+        <Button
+          color="primary"
+          onClick={handleUpgrade}
+          variant="contained"
+          disabled={!canBeMade}
+        >
           Upgrade Tool
         </Button>
       </CardActions>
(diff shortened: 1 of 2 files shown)


jeremyckahn (approved):
Thanks for the quick fix @lstebner! The game save you provided made testing this super easy, so thank you for that as well.

Good point about the accessibility issue. I've filed #273 to track that for when we can get to it!

Feel free to merge and release! :+1:

PR #275, opened 2022-04-09
convert Achievement test from enzyme to rtl

lstebner:
### What this PR does

converts the Achievement test to use react testing library as part of the #269 effort. i also converted a proptype from an object to a shape because when i went to write the test i wasn't sure what achievement should look like without digging into the use.

tests need to pass

n/a

### Additional information

n/a

### DIFF
--- a/src/components/Achievement/Achievement.test.js
+++ b/src/components/Achievement/Achievement.test.js
@@ -1,16 +1,47 @@
 import React from 'react'
-import { shallow } from 'enzyme'
+import { render, screen } from '@testing-library/react'
+
+import FarmhandContext from '../../Farmhand.context'
 
 import Achievement from './Achievement'
 
-let component
+describe('<Achievement />', () => {
+  let achievementObject
 
-beforeEach(() => {
-  component = shallow(
-    <Achievement {...{ achievement: {}, completedAchievements: {} }} />
-  )
-})
+  beforeEach(() => {
+    achievementObject = {
+      description: 'the best achievement',
+      id: 'achievement-1',
+      name: 'achievement one',
+      rewardDescription: 'the greatest reward',
+    }
 
-test('renders', () => {
-  expect(component).toHaveLength(1)
+    const farmhandContextValue = {
+      gameState: {},
+      handlers: {},
+    }
+
+    render(
+      <FarmhandContext.Provider value={farmhandContextValue}>
+        <Achievement
+          achievement={achievementObject}
+          completedAchievements={{}}
+        />
+      </FarmhandContext.Provider>
+    )
+  })
+
+  test('renders the name of the achievement', () => {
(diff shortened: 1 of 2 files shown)


src/components/Achievement/Achievement.test.js
```diff
@@ -1,16 +1,47 @@
...
+      handlers: {},
+    }
+
+    render(
+      <FarmhandContext.Provider value={farmhandContextValue}>
+        <Achievement
+          achievement={achievementObject}
+          completedAchievements={{}}
+        />
+      </FarmhandContext.Provider>
```
jeremyckahn, line 30: This is a really smart pattern for testing components that require game state context. I'll use this for other test cases as well!

src/components/Achievement/Achievement.test.js
```diff
@@ -1,16 +1,47 @@
...
+
+  test('renders the achievement description', () => {
+    expect(screen.getByText(achievementObject.description)).toBeInTheDocument()
+  })
 
-test('renders', () => {
-  expect(component).toHaveLength(1)
+  test('renders the reward description', () => {
+    expect(
+      screen.getByText(new RegExp(achievementObject.rewardDescription))
```
jeremyckahn, line 44: How come a `RegExp` was needed here, but not for the other assertions?

jeremyckahn (approved):
Thank for adding these tests! I definitely took some shortcuts with some of the propType definitions so thank you for improving that as well! 🙏

src/components/Achievement/Achievement.test.js
lstebner: it's because there is more text inside the element (if i recall it's "Reward:\n${rewardDescription}") so the regex is required to tell it that it's ok it's not the entire text of the element (even though the description string isn't broken apart). i actually only found out because i started with the string matcher on this one

jeremyckahn: Ah gotcha. This is something about React Testing Library I don't love. When I run into this I use `exact: false`:

```diff
diff --git a/src/components/Achievement/Achievement.test.js b/src/components/Achievement/Achievement.test.js
index 0d0877a..6853ad2 100644
--- a/src/components/Achievement/Achievement.test.js
+++ b/src/components/Achievement/Achievement.test.js
@@ -41,7 +41,7 @@ describe('<Achievement />', () => {
 
   test('renders the reward description', () => {
     expect(
-      screen.getByText(new RegExp(achievementObject.rewardDescription))
+      screen.getByText(achievementObject.rewardDescription, { exact: false })
     ).toBeInTheDocument()
   })
 })
```

I'm not sure if there's a benefit to one over the other, but that's worked out for me so far!

lstebner: oh, neat! i haven't discovered that but i like that more. will update next time i'm in the area

PR #278, opened 2022-05-06
refactor: convert AccountingView spec from enzyme to testing-library

lstebner:
### What this PR does

Convert spec file from enzyme to using testing-library.

if unit tests pass, it's good to go!

This isn't quite 100% coverage, but it's pretty close. There are currently two components in the `AccountingView.js` file and I think if the other was split out the coverage would be easier to achieve.

### Additional information

Screen Shot 2022-05-06 at 9 34 01 AM
Screen Shot 2022-05-06 at 9 34 40 AM

### DIFF
--- a/src/components/AccountingView/AccountingView.test.js
+++ b/src/components/AccountingView/AccountingView.test.js
@@ -45,7 +45,7 @@ describe('<AccountView />', () => {
       expect(screen.getByText('Loan Balance')).toBeInTheDocument()
     })
 
-    test('it displays the load input', () => {
+    test('it displays the loan input', () => {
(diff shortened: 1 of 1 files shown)


jeremyckahn (approved):
Awesome work! Thanks for adding these new tests. LGTM!

> This isn't quite 100% coverage, but it's pretty close. There are currently two components in the `AccountingView.js` file and I think if the other was split out the coverage would be easier to achieve.

Good call out. I'm not too concerned with <100% coverage. We have more with this PR than without it, and we're testing the critical functionality. That's what really matters! 

Nothing about the code is set in stone, so feel free to make any refactors that improve testability going forward!

Issue #268, opened 2022-03-31
Fix hugging machine exploit


jeremyckahn:
Reported by a player on Discord:

> I found a bug/exploit with hugging machine and cow feed, when you get X% more for all sales their sell price increases but the buy prices stay the same. I am now able to buy hugging machines for 500 and sell for 625 instantly.

PR #279, opened 2022-05-09
Issue #268 - Prevent cow shop items from being re-sold for profit

MarquisDeMizzle:
### What this PR does
Resolves issue #268, preventing cow shop items from being resold for profit

Selling cow feed or hugging machines

### Additional information
image

### DIFF
--- a/src/components/CowPenContextMenu/CowPenContextMenu.js
+++ b/src/components/CowPenContextMenu/CowPenContextMenu.js
@@ -21,7 +21,7 @@ import {
   nullArray,
 } from '../../utils'
 import { PURCHASEABLE_COW_PENS } from '../../constants'
-import { cowFeed, huggingMachine } from '../../data/items'
+import shopInventory from '../../data/shop-inventory-cow'
 
 import CowCard from '../CowCard'
 
@@ -197,24 +197,17 @@ export const CowPenContextMenu = ({
       </TabPanel>
       <TabPanel value={currentTab} index={2}>
         <ul className="card-list">
-          <li>
-            <Item
-              {...{
-                item: cowFeed,
-                isPurchaseView: true,
-                showQuantity: true,
-              }}
-            />
-          </li>
-          <li>
-            <Item
-              {...{
-                item: huggingMachine,
-                isPurchaseView: true,
-                showQuantity: true,
-              }}
-            />
-          </li>
+          {shopInventory.map(i => (
+            <li>
+              <Item
+                {...{
+                  item: i,
+                  isPurchaseView: true,
+                  showQuantity: true,
+                }}
+              />
+            </li>
+          ))}
         </ul>
       </TabPanel>
     </div>
--- a/src/components/CowPenContextMenu/CowPenContextMenu.js
+++ b/src/components/CowPenContextMenu/CowPenContextMenu.js
@@ -21,7 +21,7 @@ import {
   nullArray,
 } from '../../utils'
 import { PURCHASEABLE_COW_PENS } from '../../constants'
-import shopInventory from '../../data/shop-inventory-cow'
+import cowShopInventory from '../../data/shop-inventory-cow'
 
 import CowCard from '../CowCard'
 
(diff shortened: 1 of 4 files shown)


src/components/CowPenContextMenu/CowPenContextMenu.js
```diff
@@ -197,24 +197,17 @@ export const CowPenContextMenu = ({
...
-          <li>
-            <Item
-              {...{
-                item: huggingMachine,
-                isPurchaseView: true,
-                showQuantity: true,
-              }}
-            />
-          </li>
+          {shopInventory.map(i => (
```
jeremyckahn, line 200: I think it's worth keeping variable names descriptive even in a limited context like this. So, let's change `i` to `item` here. As a bonus, this will let us simplify line 204 below to `item,`.

src/components/CowPenContextMenu/CowPenContextMenu.js
```diff
@@ -197,24 +197,17 @@ export const CowPenContextMenu = ({
...
+            <li>
```
jeremyckahn, line 201: `li` needs a `key` prop.

src/data/shop-inventory-cow.js
```diff
@@ -0,0 +1,5 @@
+import { cowFeed, huggingMachine } from './items'
+
+const inventory = [cowFeed, huggingMachine]
+
+export default inventory
```
jeremyckahn, line 5: This is a really nice pattern! I hadn't thought of adding more `ship-inventory-*.js` files like this. Great idea! 😄

src/components/CowPenContextMenu/CowPenContextMenu.js
```diff
@@ -21,7 +21,7 @@ import {
   nullArray,
 } from '../../utils'
 import { PURCHASEABLE_COW_PENS } from '../../constants'
-import { cowFeed, huggingMachine } from '../../data/items'
+import shopInventory from '../../data/shop-inventory-cow'
```
jeremyckahn, line 24: What do you think about renaming `shopInventory` to something like `cowShopInventory` to match `utils.js`?

src/utils.js
```diff
@@ -76,11 +77,16 @@ const Jimp = configureJimp({
...
-const shopInventoryMap = shopInventory.reduce((acc, item) => {
+const purchasableItemMap = cowShopInventory.reduce((acc, item) => {
   acc[item.id] = item
   return acc
 }, {})
 
+shopInventory.reduce((acc, item) => {
+  acc[item.id] = item
+  return acc
+}, purchasableItemMap)
```
jeremyckahn, line 88: It took me a moment to work out what this modified logic is doing. It wasn't immediately obvious to me that `purchasableItemMap` is being built up across these two `reduce` operations, so that suggests to me that it's worth making this logic little clearer. What do you think about something along the lines of this?

```js
const purchasableItemMap = [...cowShopInventory, ...shopInventory].reduce(
  (acc, item) => {
    acc[item.id] = item
    return acc
  },
  {}
)
```

The idea here is to keep all of the data structure logic in one (hopefully straightforward) block.

jeremyckahn (changes requested):
Awesome work @MarquisDeMizzle! I tested this out and indeed the bug is gone. Thank you for the much-needed fix!

I had a number of mostly stylistic comments. Let me know if you want to discuss anything I brought up.

And welcome to Farmhand! 🐮

MarquisDeMizzle:
All your comments should be addressed.

src/components/CowPenContextMenu/CowPenContextMenu.js
```diff
@@ -197,24 +197,17 @@ export const CowPenContextMenu = ({
...
-            <Item
-              {...{
-                item: huggingMachine,
-                isPurchaseView: true,
-                showQuantity: true,
-              }}
-            />
-          </li>
+          {cowShopInventory.map((item, i) => (
+            <li key={i}>
```
jeremyckahn, line 201: `key={i}` should work just fine here, but it seems easy enough to avoid an index-based key so I'll tweak that after merging this.

jeremyckahn (approved):
Awesome work @MarquisDeMizzle! Thanks again for fixing this, and for being receptive to feedback. I'll merge this now, make the `key` tweak I mentioned in my inline comment, and then release this fix to Prod.

src/components/CowPenContextMenu/CowPenContextMenu.js
jeremyckahn: This has been changed to `item.id` in https://github.com/jeremyckahn/farmhand/commit/0384b149f453486c2b38dcc0ffd8557d36513ddc.

PR #280, opened 2022-05-14
Break Apart Reducers: Act 1

jeremyckahn:
### What this PR does

This PR is step one of a larger effort to move to a one-reducer-per-file project organization. This yields many benefits:

  - Smaller files prevent getting "lost" in the monolithic `reducers/index.js` and `reducers/index.test.js` files
  - Tests run faster (fewer irrelevant tests are run in "watch" mode, and file test run loads can be better spread across processor cores)
  - Reducers become fuzzy-findable via their matching file names

Do a high-level smoke test of the game and ensure that nothing crashes.

`src/game-logic/reducers` is going to get quite large:

<details>

```
$ tree src/game-logic/reducers/
src/game-logic/reducers/
├── addItemToInventory.js
├── addItemToInventory.test.js
├── applyCrows.js
├── applyPrecipitation.js
├── applyPrecipitation.test.js
├── clearPlot.js
├── clearPlot.test.js
├── decrementItemFromInventory.js
├── decrementItemFromInventory.test.js
├── fertilizePlot.js
├── fertilizePlot.test.js
├── harvestPlot.js
├── harvestPlot.test.js
├── helpers.js
├── incrementPlotContentAge.js
├── incrementPlotContentAge.test.js
├── index.js
├── index.test.js
├── minePlot.js
├── minePlot.test.js
├── modifyFieldPlotAt.js
├── plantInPlot.js
├── plantInPlot.test.js
├── processField.js
├── processSprinklers.js
├── processSprinklers.test.js
├── processWeather.js
├── processWeather.test.js
├── removeFieldPlotAt.js
├── resetWasShoveled.js
├── resetWasShoveled.test.js
├── setScarecrow.js
├── setScarecrow.test.js
├── setSprinkler.js
├── setSprinkler.test.js
├── showNotification.js
├── waterAllPlots.js
├── waterAllPlots.test.js
└── waterField.js

0 directories, 39 files
```

</details>

But this is still better than the current state of things.

### Additional information

No new logic was introduced in this PR, as it is a reasonably pure refactor. Therefore, I would suggest not spending much time reviewing the reducer logic.

Various `TODO`s have been added throughout for cleanup needs that were surfaced by this refactor.

There will be additional PRs that look similar to this until the conversion is done. My plan is to incrementally move reducers to isolated files, rather than making the whole change all at once and having a _completely_ unmanageable PR. :sweat_smile:


src/game-logic/reducers/addItemToInventory.js
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Only adds as many items as there is room in the inventory for unless
+ * allowInventoryOverage is true.
+ * @param {farmhand.state} state
+ * @param {farmhand.item} item
+ * @param {number} [howMany=1]
+ * @param {boolean} [allowInventoryOverage=false]
+ * @returns {farmhand.state}
+ */
+export const addItemToInventory = (
```
lstebner, line 12: what do you think about making these default exports? I still tend to use the `function` keyword though I have no strong opinions one way or another on that specifically (but I will use it in my example hah). i think making these the default export makes sense though since they are the only export, i.e. `export default function addItemToInventory(...`

the `index.js` can also take advantage of this with a one line import/export like `export { default as addItemToInventory } from "./addItemToInventory"`.

src/game-logic/reducers/addItemToInventory.test.js
```diff
@@ -0,0 +1,107 @@
+import { addItemToInventory } from './addItemToInventory'
```
lstebner, line 1: continuing the default export thread, these would end up not needing the destructure syntax. a minor difference, but destructuring makes me think there are other exports there as well, and even if there were i think it's still nice to have a default one available

lstebner (commented):
left a comment that applies to every file so I wanted to submit it for your thoughts @jeremyckahn and then based on that finish my review. though, there's not much to see since you didn't actually modify behavior here. nice job moving these all! it's going to be much easier to maintain going forward

src/game-logic/reducers/addItemToInventory.js
jeremyckahn: It was a conscious choice to avoid default exports here, as I'm seeing them more and more as an antipattern. [This article](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad) makes a case against them. The point I find compelling in that article is the "Refactoring" section:

> Default exports make large-scale refactoring impossible since each importing site can name default import differently (including typos).

```js
// in file exports.js
export default function () {...}// in file import1.js

import doSomething from "./exports.js"// in file import2.js

import doSmth from "./exports.js"
```

So, relying solely on named exports seems like a good syntactical guard against human error and the maintenance challenges that can result.

I realize that default exports are currently used throughout the codebase, and updating them all would be at least [a modest effort](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/jeremyckahn/farmhand%24+export+default&patternType=literal). Rather than going back to fix them all, I could draft an ADR to define the project standard for new code to adhere to. We could even use [this ESLint rule](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-default-export.md) to enforce style once we have removed them all.

How would you feel about setting this code standard for the project?

src/game-logic/reducers/addItemToInventory.test.js
jeremyckahn: The import destructuring is an unfortunate bit of syntactical overhead with named exports, for sure. Personally, I think it's balanced out by never needing to think about whether an import is default or named, as you can always assume the latter.

src/game-logic/reducers/addItemToInventory.js
lstebner: thanks for linking the article! the tree-shaking seems like a really great reason too. i've never read this, but i'm definitely ok with setting this as a standard for this project. i'll have to ponder a bit more before i go to work and change our stance, but you might have me sold 😄

src/game-logic/reducers/addItemToInventory.test.js
lstebner: yeah that's true. my comment here was moot after your first response anyways. i'm gonna resolve these

src/game-logic/reducers/applyCrows.js
```diff
@@ -0,0 +1,45 @@
...
+import { CROW_CHANCE } from '../../constants'
+import { CROW_ATTACKED } from '../../templates'
+
+import { fieldHasScarecrow, updateField } from './helpers'
+
+/**
+ * @param {farmhand.state} state
+ * @returns {farmhand.state}
+ */
+export const applyCrows = state => {
```
lstebner, line 13: just noticing there seems to be no spec for this one. wanted to flag it in case it got missed, but if none existed it doesn't have to block this PR

src/game-logic/reducers/modifyFieldPlotAt.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @param {farmhand.state} state
+ * @param {number} x
+ * @param {number} y
+ * @param {Function(?farmhand.plotContent)} modifierFn
+ * @returns {farmhand.state}
+ */
+export const modifyFieldPlotAt = (state, x, y, modifierFn) => {
```
lstebner, line 8: flagging this one as missing tests as well

src/game-logic/reducers/processField.js
```diff
@@ -0,0 +1,28 @@
...
+  incrementPlotContentAge,
+  resetWasWatered,
+  resetWasShoveled,
+]
+
+/**
+ * @param {farmhand.state} state
+ * @returns {farmhand.state}
+ */
+export const processField = state => ({
```
lstebner, line 23: missing tests 🏴

src/game-logic/reducers/showNotification.js
```diff
@@ -0,0 +1,35 @@
...
+// many formal parameters.
+/**
+ * @param {farmhand.state} state
+ * @param {string} message
+ * @param {string} [severity] Corresponds to the `severity` prop here:
+ * https://material-ui.com/api/alert/
+ * @returns {farmhand.state}
+ * @see https://material-ui.com/api/alert/
+ */
+export const showNotification = (
```
lstebner, line 11: tests 🏴

src/game-logic/reducers/waterField.js
```diff
@@ -0,0 +1,10 @@
+import { setWasWatered, updateField } from './helpers'
+
+/**
+ * @param {farmhand.state} state
+ * @returns {farmhand.state}
+ */
+export const waterField = state => ({
```
lstebner, line 7: tests 🏴

lstebner (commented):
i didn't think it was worth looking at the internals of each util in this PR so the most useful thing i could think of doing was to review the files added and making sure the names didn't look to have any typos (all seems good there). CI being green feels like the main indicator, but i did notice a few utils don't have test files. i assume they never existed, but i flagged them for you just in case something got missed. we could file a task to get them added if it's all known as i don't think it needs to hold up this PR. up to you though!

src/game-logic/reducers/showNotification.js
jeremyckahn: It turns out that this actually had tests! I moved them into their own file in https://github.com/jeremyckahn/farmhand/commit/c337899e6c0ae332648d56a457bcc2c734fe3173. Thanks for calling this out!

src/game-logic/reducers/waterField.js
jeremyckahn: Added a TODO for this in https://github.com/jeremyckahn/farmhand/commit/b28490644c45a589008d909fc85add26f51d46df.

src/game-logic/reducers/modifyFieldPlotAt.js
jeremyckahn: Added a TODO for this in https://github.com/jeremyckahn/farmhand/commit/b28490644c45a589008d909fc85add26f51d46df.

src/game-logic/reducers/applyCrows.js
jeremyckahn: Added a TODO for this in https://github.com/jeremyckahn/farmhand/commit/b28490644c45a589008d909fc85add26f51d46df.

jeremyckahn (commented):
Good idea to flag the missing tests @lstebner! I consider adding tests to be out of scope for this PR, but it's worth adding to TODOs to make sure they get added down the line. That's done now!

src/game-logic/reducers/addItemToInventory.js
jeremyckahn: I've added an ADR for this decision in https://github.com/jeremyckahn/farmhand/commit/6bdc2cd3b2c11522e0f95f3e1fa6f9f0c4df0b53!

lstebner (approved):

PR #282, opened 2022-05-22
Use `mprocs` for `dev` script

jeremyckahn:
### What this PR does

This PR reimplements the `dev` script to use [`mprocs`](https://github.com/pvolok/mprocs) instead of `concurrently`. This make the different sub-scripts interactive and easier to make sense of since the output isn't all mixed together.

I also made some README updates.

I'm not sure that it can be on any machine other than mine because of https://github.com/vercel/vercel/discussions/4925. 😕 However, this is generally true of the `dev` script right now.

### Additional information

https://user-images.githubusercontent.com/366330/169719779-76af6e58-e9ab-4a5e-86c8-81c8c4f87fca.mp4


lstebner (approved):
looks sane to me!

Issue #281, opened 2022-05-17
Define smoke test process


jeremyckahn:
Define a smoke test process that any player can perform to ensure stability. Tasks should be relatively high-level and avoid being overly prescriptive to allow for a degree of spontaneity during testing. This spontaneity should hopefully result in wider test coverage.

lstebner:
was thinking about this so figured I might as well leave my thoughts here rather than just in my head.

- starting from a brand new game
    - [ ] can purchase seeds from shop
    - [ ] can plant a seed
    - [ ] can plant seeds with different range settings
    - [ ] can water seeds
    - [ ] empty plots don't get watered
    - [ ] can hoe seeds
    - [ ] can end the day
    - [ ] reloading the page resumes the game in progress
    - [ ] can purchase storage upgrades
    - [ ] can't store more than storage allows
    - [ ] can purchase cowpen
    - [ ] can purchase cows
    - [ ] can offer cow for trade
    - [ ] can harvest grown crops
    - [ ] can sell grown crops
    - [ ] recipes unlock from selling crops
    - [ ] can make recipes
    - [ ] can sell crafted recipes
    - [ ] can hug cows
    - [ ] can level up

late game / mining features probably need a save to load to test efficiently
    - [ ] combine auto harvests crops
    - [ ] shovel unlocks at level 6
    - [ ] can purchase smelter
    - [ ] can dig empty plots with the shovel
    - [ ] plots with crops don't get shoveled
    - [ ] can sell mined items
    - [ ] can craft ores
    - [ ] can craft tool upgrades
    - [ ] <insert some tests for each upgraded tool to make sure they do what they're supposed to>
    - [ ] cannot have more cows than cowpen allows
    - [ ] can breed cows
    - [ ] cow offspring have correct traits
    - [ ] can breed rainbow cow
    - [ ] can pay off loan


that's everything off the top of my head! there's probably more stuff even, but I went by memory here :smile:

PR #285, opened 2022-05-24
Smoke test issue templates

jeremyckahn:
### What this PR does

This PR adds a GitHub Issue template for performing manual smoke tests. The idea is to give users an interactive checklist so they can track their progress and share it publicly.

We'll have to merge this to `develop` for any of the links referenced in this PR to work, unfortunately.

I don't consider this checklist complete or exhaustive (it's mostly just what @lstebner posted in [this comment](https://github.com/jeremyckahn/farmhand/issues/281#issuecomment-1134138883)), it's just the first iteration. I just wanted to get something going quickly!


README.md
```diff
@@ -95,6 +95,10 @@ The system design for Farmhand's multiplayer functionality has been [detailed in
 
 Automation is done with [GitHub Actions](.github/workflows). All changes are tested and built upon Git push. Merges to `main` automatically deploy to Production (the `gh-pages` branch) upon a successful test run and build.
 
+### Smoke Testing
+
+All are welcome to help with smoke testing Farmhand to ensure the game is stable! All you need is to be logged into GitHub and open a [new Smoke Test Report issue](https://github.com/jeremyckahn/farmhand/issues/new?template=smoke-test-report.md). This issue will have interactive checkboxes that you can use to track your progress. Once smoke testing is complete, please close the issue.
```
lstebner, line 100: is any of the list critical/blocker for deploying? i was thinking of smoketesting as a pre-deploy thing, so maybe even just a line here about when/why you would do a smoketest would explain.

src/test-utils/saves/lots-of-money.json
```diff
@@ -0,0 +1,3 @@
+{
+  "money": 1000000
```
lstebner, line 2: nice, so simple!

lstebner (approved):
left a thought but it doesn't have to block if you don't have an answer right now. this is a great step! especially the quick access via template

README.md
jeremyckahn: Good thinking! I added a note about this in https://github.com/jeremyckahn/farmhand/commit/97c416e09cac4898336dcba29bfab89e0a702273.

> is any of the list critical/blocker for deploying?

I don't want to be too prescriptive with this process at the outset, as we don't even know if it will be used all that much. I think keeping it a bit looser will make folks more inclined to participate, and we'll be able to evaluate issues on a case-by-case basis. Let's improve this process as we go!

PR #290, opened 2022-06-08
refactor: Move `Farmhand` component to `components` directory

jeremyckahn:
### What this PR does

This PR moves the `Farmhand` component and its related files to the `components` directory. Now `<Farmhand />` is treated just like any other component. This is a preliminary step for breaking up the rather rather large `Farmhand.js` file into smaller helper files. Now that the `Farmhand` component is in a subdirectory rather than `src`, we can divide it into many files without worrying about cluttering up `src`.

Boot the game and ensure it doesn't crash, and ensure the tests pass.

### Additional information

Similarly to the splitting of reducer files, I plan to move many of the private helpers in `Farmhand.js` into their own files. This will help us get away from Enzyme (#269) and have a generally more manageable project structure.


lstebner (approved):
🎉 easy one! (to review). nice job :)

PR #294, opened 2022-06-20
refactor: Modularize `<Farmhand />` component

jeremyckahn:
### What this PR does

This PR does some refactoring of the `<Farmhand />` component and also converts the bulk of its Enzyme-based unit tests to React Testing Library-based integration tests. There's a few Enzyme tests that remain, but I'm opening this PR now to unblock @lstebner so that he can work on converting other component tests with the helpers introduced here.

There are a variety of refactors in this PR, but it's all in service of #269.

Ensure the game boots and that the tests pass.


src/components/Farmhand/Farmhand.js
```diff
@@ -271,19 +270,21 @@ const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
...
    * @member farmhand.Farmhand#state
    * @type {farmhand.state}
    */
   state = this.createInitialState()
 
+  static defaultProps = {
+    localforage: localforage.createInstance({
+      name: 'farmhand',
+      description: 'Persisted game data for Farmhand',
+    }),
```
jeremyckahn, line 283: This change effectively lets us dependency-inject LocalForage, as can be seen in some of the new tests in this PR. There are a few benefits to this:

  - We can drop the `initialState` construct, which is a feature that was designed specifically for testing and therefore not ideal.
  - We can hydrate the app with testing state by leveraging standard features of the game (loading a save file).

src/components/Farmhand/Farmhand.js
```diff
@@ -314,12 +315,7 @@ export default class Farmhand extends Component {
   get playerInventoryQuantities() {
     const { inventory } = this.state
 
-    return itemIds.reduce((acc, itemId) => {
-      const itemInInventory = inventory.find(({ id }) => id === itemId)
-      acc[itemId] = itemInInventory ? itemInInventory.quantity : 0
-
-      return acc
-    }, {})
```
jeremyckahn, line 322: This getter's logic was moved to an external helper so that it can be tested in isolation.

src/components/Farmhand/Farmhand.test.js
```diff
@@ -45,182 +36,61 @@ jest.mock('../..//constants', () => ({
...
-    test('filters out non-plantable items', () => {
       expect(plantableCropInventory).toEqual([sampleCropSeedsItem1])
     })
   })
 })
 
-describe('getters', () => {
-  describe('playerInventoryQuantities', () => {
-    test('computes a map of item IDs to their quantity in the inventory', () => {
-      component.setState({
```
jeremyckahn, line 148: Most of the Enzyme tests in this file have been replaced by React Testing Library tests that are fundamentally different in design, but should cover all of the same use cases in a more effective way. These new tests are in separate files, since they are concerned with broader game logic rather than component-level units of functionality.

src/game-logic/cows/cow-selection.test.js
```diff
@@ -19,12 +20,20 @@ describe('cow selection', () => {
...
-      stageFocus: stageFocusType.COW_PEN,
       purchasedCowPen: 1,
     })
 
+    await farmhandStub({
+      localforage: {
+        getItem: () => Promise.resolve(loadedState),
+        setItem: (_key, data) => Promise.resolve(data),
+      },
+    })
```
jeremyckahn, line 33: This is the new preferred way to hydrate the game with state for testing.

src/test-utils/ui.js
```diff
@@ -0,0 +1,24 @@
+import { screen } from '@testing-library/react'
```
jeremyckahn, line 1: I imagine that this file will be filled out more over time with many more React Testing Library helpers.

jeremyckahn (commented):
I left some comments that should hopefully make reviewing this a little easier!

src/components/Farmhand/Farmhand.js
lstebner: it looks like the getter is only used in one place; what do you think about just calling the helper there and removing the getter?

src/components/Farmhand/helpers/getPlayerInventoryQuantities.js
```diff
@@ -0,0 +1,16 @@
+import { itemsMap } from '../../../data/maps'
+
+const itemIds = Object.keys(itemsMap)
+
+/**
+ * @param {Array.<{ id: farmhand.item, quantity: number }>} inventory
+ * @returns {Object.<string, number>}
+ */
+export const getPlayerInventoryQuantities = inventory => {
```
lstebner, line 9: does this need the word "Player" in it? seems like any inventory could come through and work with this method (though maybe today the player's inventory is the only one that exists)

src/components/Farmhand/helpers/getPlayerInventoryQuantities.js
```diff
@@ -0,0 +1,16 @@
...
+  return itemIds.reduce((acc, itemId) => {
+    const itemInInventory = inventory.find(({ id }) => id === itemId)
+    acc[itemId] = itemInInventory ? itemInInventory.quantity : 0
+
+    return acc
+  }, {})
```
lstebner, line 15: i think this would actually be more efficient as two separate operations. first one being a loop over `itemIds` to set `acc[itemId] = 0` and then a 2nd separate loop over `inventory` setting to the quantity. the way it is now, each find will loop over some portion of `inventory` that is going to be redundant.

src/game-logic/cows/cow-selection.test.js
lstebner: as in, do this instead of creating a farmhand context provider? and does this mean every test has to be async now?

src/shell/navigation.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+describe('navigation', () => {
+  test('cycles forwards through the standard views', async () => {
+    await farmhandStub()
+    const [header] = document.getElementsByTagName('header')
+
+    for (const viewName of ['Shop', 'Field', 'Workshop', 'Home']) {
+      await nextView()
+      within(header).getByText(viewName)
+    }
```
lstebner, line 15: there is no `expect` in any of these tests in this file

src/shell/notifications.test.js
```diff
@@ -0,0 +1,88 @@
...
+  test('pending notifications are shown when a day starts', async () => {
+    jest.spyOn(Math, 'random').mockReturnValue(0)
+
+    await farmhandStub()
+    await endDay()
+
+    // STORM_MESSAGE message is a sample pending day notification that is known
+    // to show upon starting a new day when Math.random() is locked to 0.
+    await screen.findByText(STORM_MESSAGE)
+  })
```
lstebner, line 20: no `expect` in these tests either

src/shell/notifications.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+describe('notifications', () => {
+  test('pending notifications are shown when a day starts', async () => {
+    jest.spyOn(Math, 'random').mockReturnValue(0)
+
+    await farmhandStub()
+    await endDay()
+
+    // STORM_MESSAGE message is a sample pending day notification that is known
+    // to show upon starting a new day when Math.random() is locked to 0.
```
lstebner, line 18: would be great to capture this in a variable/function. like `STORM_CHANCE_VALUE = 0` used for the `mockReturnValue` above, or a `isStormingToday` util that can be mocked

src/shell/notifications.test.js
```diff
@@ -0,0 +1,88 @@
...
+    const carrotInput = within(carrotItem).getByDisplayValue('1')
+    userEvent.type(carrotInput, '10')
+    const carrotSellButton = within(carrotItem).getByText('Sell')
+    userEvent.click(carrotSellButton)
+    const notification = await screen.findByRole('alert')
+
+    within(notification).getByText('You learned a new recipe:', {
+      exact: false,
+    })
+    within(notification).getByText('Carrot Soup')
```
lstebner, line 44: i think only one of these is necessary

src/shell/notifications.test.js
```diff
@@ -0,0 +1,88 @@
...
+    const carrotSellButton = within(carrotItem).getByText('Sell')
+
+    userEvent.type(cornInput, '6')
+    userEvent.click(cornSellButton)
+
+    userEvent.type(spinachInput, '30')
+    userEvent.click(spinachSellButton)
+
+    userEvent.type(carrotInput, '10')
+    userEvent.click(carrotSellButton)
```
lstebner, line 82: this ends up looking like a test that is checking to make sure recipe unlock logic is working rather than simply the notifications. maybe that was your goal, but it stands out to me as a bit overly complex and more of an integration rather than a unit test. it makes me wonder if there's a util that exists (or should exist) that can be mocked such as `getTodaysNotifications` instead?

src/test-utils/stubs/saveFileStubFactory.js
```diff
@@ -0,0 +1,11 @@
+import { computeStateForNextDay } from '../../game-logic/reducers'
+
+export const saveFileStubFactory = ({ dayCount = 1, ...restOverrides }) =>
```
lstebner, line 3: this feels more like `saveData` rather than `saveFile` 🤔

src/game-logic/cows/cow-selection.test.js
lstebner: also can a note be added to the README about how to work with this?

src/components/Farmhand/helpers/getPlayerInventoryQuantities.js
jeremyckahn: Good catch! This is optimized in https://github.com/jeremyckahn/farmhand/commit/3ae18d7efd8ad51e11d10dc266c30b5d9d14b4d5. It's faster AND easier to follow! :clap:

jeremyckahn (commented):
I finally had a chance to catch up with the review here! @lstebner you had some really good suggestions all around. A few threads warrant a bit more discussion, but I think this is already looking better! 😄

jeremyckahn:
Oh whoops, I just noticed that there were a few threads I didn't follow up on. Sometimes I get lost in GitHub's PR UI. I'll address the remaining comments when I can!

src/shell/navigation.test.js
lstebner: I personally prefer every test have an expect (and in my world, only one expect 😄) but I do know what you mean about how they would still work as tests this way. I guess I like the explicit expect for consistency's sake, but won't hold up the PR for this if it's your style not to include them.

src/shell/notifications.test.js
lstebner: I will give that a read though I'm definitely not anti-integration test. I think the part that stood out to me as convoluted was that you need to understand recipe ingredients to understand this test, rather than just the fact that it's testing how multiple notifications work together. I'll give it some more thought on a fresh coffee though (and after reading that article, thanks for sharing!)

src/shell/notifications.test.js
jeremyckahn: Added in https://github.com/jeremyckahn/farmhand/commit/37e407d8117b4993d2e17cf968c54ee9cc6504aa!

src/shell/navigation.test.js
jeremyckahn: I've actually been having a similar conversation about this exact thing at work this week, and my coworkers also prefer to have explicit `expect`ations, as they are clearer and easier to follow. So, it seems my opinion is unpopular. 😂

I'm fine with explicit `expect`ations. I've added them in https://github.com/jeremyckahn/farmhand/commit/37e407d8117b4993d2e17cf968c54ee9cc6504aa!

jeremyckahn (commented):
I've followed up with everything (I think for real this time). Thanks for the helpful suggestions so far!

src/shell/notifications.test.js
lstebner: a comment to me always says there's room for a code change to better explain what's going on, but i don't really want to be _too_ fussy over this. reading the comment though made me think maybe there's room for a method similar to what I called out but local to this spec. that way when looking at this test you can read it like:

```
unlockSummerSalad()
unlockCarrotSoup()

// expect some notifications...
```

src/components/Farmhand/Farmhand.js
```diff
@@ -271,19 +270,21 @@ const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
...
-    description: 'Persisted game data for Farmhand',
-  })
-
   /*!
    * @member farmhand.Farmhand#state
    * @type {farmhand.state}
    */
   state = this.createInitialState()
 
+  static defaultProps = {
```
lstebner, line 279: I tend to prefer the `Farmhand.deafultProps` at the end of the file approach just for sake of consistency since it's the only way to set a static property for the functional components. just a thought. if you prefer this way it's ultimately the same

src/game-logic/reducers/applyPrecipitation.js
```diff
@@ -24,7 +24,7 @@ export const applyPrecipitation = state => {
   let scarecrowsConsumedByReplanting = 0
   let notification
 
-  if (Math.random() < STORM_CHANCE) {
+  if (shouldStormToday()) {
```
lstebner, line 27: nice!

lstebner (approved):
i left a followup comment but it's nothing major so i'm going to also approve so i don't hold this up any more than i already have! 😄

src/components/Farmhand/Farmhand.js
jeremyckahn: My preference is actually the opposite, as I think it's worth leveraging language-level features wherever it seems reasonable (`static` properties, in this case). To me it helps to indicate intent. That being said, I'm fine with a mix of `static` properties and `SomeObject.staticProperty` forms in the code base, so feel free to use it when you need to define a static property!

> it's the only way to set a static property for the functional components

That's a really good point. I actually didn't realize that functional components supported `defaultProps`. Personally, I'd avoid defining `defaultProps` for functional components and instead use default function parameters like we have here: https://github.com/jeremyckahn/farmhand/blob/734284da8b6224f6889bb6612fadb6d999b3004c/src/components/Plot/Plot.js#L77-L81

It solves the same problem, but this form allows the code to read more top-to-bottom.

In any case, I think that both forms work just fine and can be freely mixed. 🙂

lstebner: what you pointed out is fine for defaults, but doesn't work for `propTypes` (which is what i'm now realizing i had in mind when I left the original comment.) either way though this is definitely just a preference choice 😄

jeremyckahn: Ah yes, that's a good point. `propTypes` would have to be defined separately for functional components!

PR #296, opened 2022-06-22
Lstebner/rtl tests take two

lstebner:
this is a recreation of PR #295 because I messed up the branch there beyond repair. it was much simpler to just whip up a new PR. my bad 😓 

### What this PR does

updates some enzyme based tests to use testing-library

- update tests for Home component
- update tests for Workshop component
- broke out helper for `getAvailableUpgrades` from workshop and added test for it too

once I had all these new tests in place I decided to refactor the Workshop component a bit. I broke it apart into some separate smaller components and simplified some of the logic, but I'm open to suggestions/feedback on this if it doesn't seem improved, or there's further improvement that comes to mind.

- tests pass
- it's worth also taking a look at the Workshop to make sure the kitchen and forge recipes seem to still be arranged as expected

happy to refactor anything additional that stands out here

### DIFF
--- a/src/components/Workshop/ForgeTabPanel.js
+++ b/src/components/Workshop/ForgeTabPanel.js
@@ -19,6 +19,7 @@ import { getUpgradesAvailable } from './getUpgradesAvailable'
 
 export function ForgeTabPanel({
   currentTab,
+  index,
   learnedForgeRecipes,
   setCurrentTab,
   toolLevels,
@@ -29,7 +30,7 @@ export function ForgeTabPanel({
   })
 
   return (
-    <TabPanel value={currentTab} index={1}>
+    <TabPanel value={currentTab} index={index}>
       <RecipeList
         learnedRecipes={learnedForgeRecipes}
         allRecipes={recipeCategories[recipeType.FORGE]}
@@ -71,6 +72,7 @@ export function ForgeTabPanel({
 
 ForgeTabPanel.propTypes = {
   currentTab: PropTypes.number.isRequired,
+  index: PropTypes.number.isRequired,
   learnedForgeRecipes: PropTypes.array.isRequired,
   setCurrentTab: PropTypes.func.isRequired,
   toolLevels: PropTypes.object.isRequired,
--- a/src/components/Workshop/ForgeTabPanel.js
+++ b/src/components/Workshop/ForgeTabPanel.js
@@ -1,0 +1,77 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import Card from '@material-ui/core/Card'
+import CardContent from '@material-ui/core/CardContent'
+import Divider from '@material-ui/core/Divider'
+import ReactMarkdown from 'react-markdown'
+
+import { recipeType } from '../../enums'
+
+import { recipeCategories } from '../../data/maps'
+
+import UpgradePurchase from '../UpgradePurchase'
+
+import { TabPanel } from './TabPanel'
+import { RecipeList } from './RecipeList'
+
+import { getUpgradesAvailable } from './getUpgradesAvailable'
+
+export function ForgeTabPanel({
+  currentTab,
+  learnedForgeRecipes,
+  setCurrentTab,
+  toolLevels,
+}) {
+  const upgradesAvailable = getUpgradesAvailable({
+    toolLevels,
+    learnedForgeRecipes,
+  })
+
+  return (
+    <TabPanel value={currentTab} index={1}>
+      <RecipeList
+        learnedRecipes={learnedForgeRecipes}
+        allRecipes={recipeCategories[recipeType.FORGE]}
+      />
+      {upgradesAvailable.length ? (
+        <>
+          <Divider />
+          <ul className="card-list">
+            <li>
+              <h4>Tool Upgrades</h4>
+            </li>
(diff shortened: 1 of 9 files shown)


src/components/Workshop/ForgeTabPanel.js
```diff
@@ -0,0 +1,77 @@
...
+  setCurrentTab,
+  toolLevels,
+}) {
+  const upgradesAvailable = getUpgradesAvailable({
+    toolLevels,
+    learnedForgeRecipes,
+  })
+
+  return (
+    <TabPanel value={currentTab} index={1}>
```
jeremyckahn, line 32: `index={1}` seems to assume a lot about how this component will be used. Do you think it would be worth making `index` a required prop of this component so that the consumer is in charge of tab order?

src/components/Workshop/KitchenTabPanel.js
```diff
@@ -0,0 +1,51 @@
...
+import ReactMarkdown from 'react-markdown'
+
+import { recipeType } from '../../enums'
+
+import { recipeCategories } from '../../data/maps'
+
+import { TabPanel } from './TabPanel'
+import { RecipeList } from './RecipeList'
+
+export function KitchenTabPanel({
```
jeremyckahn, line 16: +1 to extracting this as well!

src/components/Workshop/KitchenTabPanel.js
```diff
@@ -0,0 +1,51 @@
...
+  currentTab,
+  setCurrentTab,
+  learnedKitchenRecipes,
+}) {
+  return (
+    <TabPanel value={currentTab} index={0}>
```
jeremyckahn, line 22: Similar note about deferring tab order to the consumer of this component.

src/components/Workshop/Workshop.js
```diff
@@ -1,97 +1,32 @@
...
+import './Workshop.sass'
 
 const Workshop = ({ learnedRecipes, purchasedSmelter, toolLevels }) => {
   const [currentTab, setCurrentTab] = useState(0)
 
-  const {
-    learnedKitchenRecipes,
-    learnedForgeRecipes,
-  } = getLearnedRecipeCategories(learnedRecipes)
+  const learnedKitchenRecipes = Object.keys(learnedRecipes).filter(
```
jeremyckahn, line 24: I love all the changes in this file. Great work!

src/components/Workshop/Workshop.test.js
```diff
@@ -1,20 +1,149 @@
...
+      <FarmhandContext.Provider value={{ gameState, handlers: {} }}>
+        <Workshop />
+      </FarmhandContext.Provider>
+    )
+  }
+
+  describe('kitchen', () => {
+    let numLearnedRecipes = 0
+    beforeEach(() => {
+      numLearnedRecipes = Object.keys(gameState.learnedRecipes).length
```
jeremyckahn, line 67: Should this line go below the `gameState.learnedRecipes` on line 69? It seems to be using stale data here.

src/components/Workshop/getUpgradesAvailable.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param {array} learnedForgeRecipes - list of learned forge recipes from farmhand state
+ * @param {object} toolLevels - the current level of each tool
+ * @returns {array} a list of all applicable upgrades
+ */
+export function getUpgradesAvailable({ learnedForgeRecipes, toolLevels }) {
+  let upgradesAvailable = []
+  const learnedRecipeIds = learnedForgeRecipes.map(r => r.id)
+
+  for (let type of Object.keys(toolUpgrades)) {
+    let upgrade = toolUpgrades[type][toolLevels[type]]
```
jeremyckahn, line 15: I realize this is preexisting, but it looks like this `let` can be a `const`.

src/components/Workshop/getUpgradesAvailable.test.js
```diff
@@ -0,0 +1,32 @@
+import { toolLevel, toolType } from '../../enums'
+
+import { getUpgradesAvailable } from './getUpgradesAvailable'
+
+describe('getUpgradesAvailable', () => {
```
jeremyckahn, line 5: I'm so glad there's a unit test for this now! 😁

jeremyckahn (changes requested):
Awesome work! This is so much cleaner now. I left a few minor comments and found a propType error, but this is a significant improvement!

src/components/Workshop/Workshop.test.js
lstebner: good catch!

src/components/Workshop/ForgeTabPanel.js
lstebner: oh yeah, great suggestion!

lstebner:
> I haven't finished reviewing this PR's code yet, but i did pull it down for testing. When I purchase the Smelter and navigate to the Workshop, I see this propType warning in the JS console:

thanks for noticing this! i forgot the prop needed to be an integer not a boolean, easy fix though 😄

jeremyckahn (approved):
Thanks for making all the changes! I confirmed that the kitchen and forge recipes are in the correct order. Awesome work! LGTM!

PR #298, opened 2022-06-29
Lstebner/more rtl tests

lstebner:
### What this PR does

- convert TierPurchase tests to use testing-library instead of enzyme
- convert component from class to functional

- CI needs to be passing
- test out tier purchase functionality in preview environment

code coverage says this one area isn't covered but I test if options are enabled and disabled in various scenarios so i think it actually is 🧐 

image

### Additional information

image

image

### DIFF
--- a/src/components/TierPurchase/TierPurchase.js
+++ b/src/components/TierPurchase/TierPurchase.js
@@ -16,7 +16,7 @@ export class TierPurchase extends Component {
   constructor(props) {
     super(props)
     this.state = {
-      selectedTier: '',
+      selectedTier: props.purchasedTier > 0 ? `${props.purchasedTier}` : '',
     }
 
     this.tierValues = [...props.tiers.entries()]
--- a/src/components/TierPurchase/TierPurchase.js
+++ b/src/components/TierPurchase/TierPurchase.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { useState } from 'react'
 import Button from '@material-ui/core/Button'
 import Select from '@material-ui/core/Select'
 import MenuItem from '@material-ui/core/MenuItem'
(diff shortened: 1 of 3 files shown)


src/components/TierPurchase/TierPurchase.js
```diff
@@ -12,142 +12,104 @@ import { func, instanceOf, node, number, string } from 'prop-types'
 import FarmhandContext from '../Farmhand/Farmhand.context'
 import './TierPurchase.sass'
 
-export class TierPurchase extends Component {
-  state = {
-    selectedTier: '',
-  }
+export function TierPurchase({
```
jeremyckahn, line 15: YAY! Thank you for converting this to a functional component. I'd like to convert all of Farmhand's components to functional components!

src/components/TierPurchase/TierPurchase.js
```diff
@@ -12,142 +12,104 @@ import { func, instanceOf, node, number, string } from 'prop-types'
...
     return (
       !!selectedTier &&
-      !this.hasPurchasedTier(selectedTierNumber) &&
+      !hasPurchasedTier(selectedTierNumber) &&
       money >= selectedTier.price
     )
   }
 
-  hasPurchasedTier = tierLevel => tierLevel <= this.props.purchasedTier
+  const onClickBuy = () => {
```
jeremyckahn, line 53: Would you be okay with changing this back to `handleTierPurchase`, or perhaps `handleBuyClick`? I've found that naming event handlers like `handleXEvent` helps to keep a distinction between prop and function names. 

This article does a pretty good job of defining a concrete style for this sort of thing: https://jaketrent.com/post/naming-event-handlers-react. I just found it, but if this seems reasonable to you I can add it to the documentation so we can start building up a style guide!

src/components/TierPurchase/TierPurchase.js
```diff
@@ -12,142 +12,104 @@ import { func, instanceOf, node, number, string } from 'prop-types'
...
-
-  componentDidMount() {
-    this.updateSelectedTier()
-  }
-
-  componentDidUpdate(prevProps) {
-    if (this.props.money !== prevProps.money) {
-      this.updateSelectedTier()
-    }
+  const onSelectChange = ({ target: { value } }) => {
```
jeremyckahn, line 61: I realize that given my event handler name comment above, this component is self-inconsistent. Sorry about that... 😅

I'd rename this to `handleSelectChange`, but this is preexisting so I don't consider it a necessary change for this PR.

jeremyckahn (changes requested):
Awesome work @lstebner! Thanks so much for doing this. There's really only one thing I wanted to address, but this PR is in pretty good shape.

As for code coverage... I never really look at it. 😅 So, I'm not too worried about that.

It might be worth incorporating coverage checks into our CI workflow so we can track it more easily (#299). I can take care of that sometime (unless that's something you'd want to do yourself).

src/components/TierPurchase/TierPurchase.js
jeremyckahn: > i actually only thought to rename this because `handleTierPurchase` is the prop and forced my hand to make a change.

Oh whoops... it looks like there's a number of preexisting issues with this file. 😅 I don't expect anyone else to clean up my mess, so don't stress about this. I think that to conform to the article, we'd need to expose an `onBuyClick` prop for this `TierPurchase` component. Changing this component's API would be blowing out the scope of this PR though, which I don't think is necessary.

Based on this part of the article:

> For the function names, I follow the exact same pattern, but I replace `on` with `handle*`, as in `handleClick`. Together, it’d look like:
>
> <MyComponent onClick={handleClick} />

I think this internal handler would need to be `handleBuyClick`. Does that seem right?

src/components/TierPurchase/TierPurchase.js
jeremyckahn: My understanding of [the article](https://jaketrent.com/post/naming-event-handlers-react) is that handlers exposed as props (the public API) should be named `on*`, and private internal handlers should be named `handle*`. Since this is the latter, `handleSelectChange` seems to conform to what the article proposes. I think I have that right?

jeremyckahn (commented):
I hope what I'm saying makes sense. I had to do a few double takes, so let me know if my suggestions aren't straightforward (or if i got something wrong)!

src/components/TierPurchase/TierPurchase.js
lstebner: none of this was used in _that_ many places so i went ahead and did the prop change. let me know though if i still don't have the method names correct!

src/components/TierPurchase/TierPurchase.js
```diff
@@ -12,147 +12,109 @@ import { func, instanceOf, node, number, string } from 'prop-types'
...
+  const handleBuyClick = () => {
+    const canAfford = tiers.get(selectedTierNumber).price <= money
 
-    const nextTierNumber = selectedTierNumber + 1
-    const nextTierToPurchase = this.props.tiers.get(nextTierNumber)
-
-    if (nextTierToPurchase && money >= nextTierToPurchase.price) {
-      this.setState({ selectedTier: String(nextTierNumber) })
+    if (canAfford) {
+      onBuyClick(selectedTierNumber)
```
jeremyckahn, line 57: Perfect! 🚀

jeremyckahn (approved):
Awesome work! Thanks for making the prop name changes as well. This is a big improvement to the code base! LGTM!

PR #301, opened 2022-07-08
feat: License code under GPL-2.0

jeremyckahn:
### What this PR does

This PR changes the license of Farmhand's code to be GPL-2.0. This aligns the project's licensing with its ideals: Open, transparent, and free.

You can review the license text if you'd like, but it was copy/pasted from [the source](https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt).

This _technically_ opens the door to people using Farmhand's code to develop a commercial, for-profit product. In Farmhand's case, this is more of a theoretical concern rather than a practical one, as it would be _extremely_ difficult to make commercially valuable product of the code given the requirements of the GPL. The game art remains under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/legalcode), so the game would have to be completely reskinned even in such a scenario. So, I'm not too worried about it. 🙂


lstebner (approved):
🚀

PR #303, opened 2022-07-10
fix: Do not overlap End Day button with QuickSelect

jeremyckahn:
### What this PR does

It reveals my hubris. 🙃

It also fixes a minor big introduced by #302 wherein the QuickSelect overlapped the End Day button on smaller screens in landscape orientation.

Play the game on a smaller screen like an iPhone SE (emulated is fine) and observe that a full tool bar does not overlap the End Day button.

I'm going to stop merging PRs without review...

### Additional information

Before this change:
Screen Shot 2022-07-10 at 18 27 20

After this change:
Screen Shot 2022-07-10 at 18 27 45

### DIFF
--- a/src/components/Field/Field.sass
+++ b/src/components/Field/Field.sass
@@ -45,7 +45,7 @@ $appBarOffset: 135px
 
       @media (orientation: landscape)
         flex-direction: row
-        margin-right: 4.5em
+        margin-right: $field-space-for-right-side-controls
 
   &[data-purchased-field="1"]
     .react-transform-element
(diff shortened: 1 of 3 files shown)


lstebner:
this is fine for the fix but it would be nice if the inventory was centered with the field here 
image

lstebner (approved):

jeremyckahn:
> this is fine for the fix but it would be nice if the inventory was centered with the field here

Great idea! This is done in https://github.com/jeremyckahn/farmhand/commit/41824c6e61eed360d404b844faec9a5731246c0e.

Screen Shot 2022-07-10 at 19 48 14

PR #305, opened 2022-07-13
feat: Native app builds

jeremyckahn:
### What this PR does

This PR automates the publishing of Electron build of Farmhand for Linux, MacOS, and Windows. This is an initial pass at this functionality and there will likely be some issues to work out. However, nothing here should affect the stability of https://www.farmhand.life/ !

I think we just have to merge this, run a release, and hope for the best. 😬

There's no way to truly test this without merging, running a release and seeing what happens. The joy of CI/CD! 😂


public/electron.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    icon: path.join(__dirname, './app-icons/Icon-512x512.png'),
+  })
+
+  // and load the index.html of the app.
+  // win.loadFile("index.html");
+  win.loadURL(
+    isDev
+      ? 'http://localhost:3000'
+      : `file://${path.join(__dirname, '../build/index.html')}`
```
lstebner, line 25: very minor comment but i find things like this much easier to read if stored into a variable instead of done inside the function call. 

```suggestion
  const hostUrl = isDev ? 'http://localhost:3000' : `file://${path.join(__dirname, '../build/index.html')}`
  
  win.loadURL(hostUrl)
```

lstebner (approved):
looks like it's working so i'm approving, but have one minor readability comment 😄

PR #307, opened 2022-07-14
feat: Universal MacOS builds

jeremyckahn:
### What this PR does

This PR makes the Farmhand app for MacOS perform better for Apple Silicon users by distributing a [universal binary](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary).

Run `npm run build:native` and verify that the build artifact at ` build/native/mac-universal/Farmhand.app` works on a Mac. You will likely see harmless errors in the build output related to missing GitHub tokens, but that is expected.

The generated app is excessively large (like 1.75GB) and I have no idea why that is. I think it's normal, just... wow.

### Additional information

I've run this in the CI job and it seems to work: https://github.com/jeremyckahn/farmhand/runs/7342659160?check_suite_focus=true

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,11 @@
       "output": "build/native"
     },
     "mac": {
-      "target": "dmg",
+      "target": {
+        "target": "dmg",
+        "arch": "universal"
+      },
+      "mergeASARs": false,
       "icon": "./public/app-icons/Icon-512x512.png"
     },
     "win": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -145,7 +145,11 @@
       "output": "build/native"
     },
     "mac": {
-      "target": "dmg",
+      "target": {
+        "target": "dmg",
+        "arch": "universal"
+      },
+      "mergeASARs": false,
```
jeremyckahn, line 152: This fix came from: https://github.com/electron-userland/electron-builder/issues/6735#issuecomment-1075372551

lstebner (approved):

Issue #308, opened 2022-07-14
Add greater zoom-out levels


jeremyckahn:
From [this Reddit post](https://www.reddit.com/r/incremental_games/comments/vycmpy/comment/ig4wk6d/?utm_source=reddit&utm_medium=web2x&context=3):

> Could you provide more zoom levels? I desperately want to zoom out at least twice more so I can see the entire field without having to scroll. On my bigger monitors I only see the top 5 rows.

PR #309, opened 2022-07-15
fix: [#308] limit field size on wider screens

jeremyckahn:
### What this PR does

This PR addressed the need described in #308 by limiting the field size for wider screens.

Open the field in a variety of screen sizes and ensure that the field is generally fully visible and accessible. It won't always be perfect, but it should at least fit on the screen reasonably well.

Enabling greater zoom-out was much harder than I anticipated due to issues similar to https://github.com/prc5/react-zoom-pan-pinch/issues/84. We have a workaround in place to deal with this here: https://github.com/jeremyckahn/farmhand/blob/4f8970728c3958ccc1a1f4406b035d74a9608c53/src/components/Field/Field.js#L376-L382

Ideally we could simply remove the `if (scale >= 1)` check, but that crashes the game. 🙁

### Additional information

Before this PR:

image

After this PR:

image

### DIFF
--- a/src/components/Field/Field.sass
+++ b/src/components/Field/Field.sass
@@ -9,6 +9,7 @@ $appBarOffset: 135px
     margin-bottom: 10.5em
   @media (orientation: landscape)
     margin-bottom: 5em
+    max-width: 120vh
 
   @media (min-width: #{$break-medium-phone})
     margin-top: 5em
--- a/src/components/Field/Field.sass
+++ b/src/components/Field/Field.sass
@@ -2,7 +2,7 @@
 
 $appBarOffset: 135px
 $bottomControlsOffset: 260px
-$horizontalQuickSelectOffest: 80px;
+$horizontalQuickSelectOffest: 80px
 $obscuringPortraitUiVerticalOffset: $appBarOffset + $horizontalQuickSelectOffest
 $obscuringLandscapeUiVerticalOffset: $appBarOffset + $bottomControlsOffset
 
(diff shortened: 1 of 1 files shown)


lstebner:
this does seem improved even if it isnt perfect so i say merge it!

lstebner (approved):

lstebner:
actually i just noticed this 
image

lstebner:
oops that last screenshot actually applies to prod as well so it isn't related to this change it seems like

jeremyckahn:
> oops that last screenshot actually applies to prod as well so it isn't related to this change it seems like

I think that's acceptable. Portrait mode primarily applies to mobile players, and in that case I think it's better for the field to be larger (even if some panning is required). I'm open to revisiting that later. In any case, I want to keep this PR focused on landscape screens in case we need to tweak this new CSS further.

[As discussed on Discord](https://discord.com/channels/714539345050075176/847884323330850847/997341663779377244), this needed some more work and I have just pushed https://github.com/jeremyckahn/farmhand/commit/9dbfa68c537c938fc63812d333d9fdc5bbf28eee to tighten up the width calculations. Let me know how this looks for you @lstebner!

lstebner (approved):

src/components/Field/Field.sass
```diff
@@ -1,17 +1,47 @@
 @import ../../styles/variables.sass
 
 $appBarOffset: 135px
+$bottomControlsOffset: 260px
+$horizontalQuickSelectOffest: 80px;
```
lstebner, line 5: looks like you don't need this semi-colon here

jeremyckahn: Thanks for spotting this! Fixed in https://github.com/jeremyckahn/farmhand/commit/e3f32839fec032811d0e808dfc43ab30da4fbb14.

PR #320, opened 2022-08-25
refactor: Remove `MINING` feature flag

jeremyckahn:
### What this PR does

This PR removes the `MINING` feature flag, which is no longer used.

Ensure that the mining-related functionality still works.

~I'm not actually sure how any of this was working with the flag still in place. It's not clear how the flag is being enabled in Production. :grimacing:~ I figured it out: https://github.com/jeremyckahn/farmhand/commit/97ca7eef988c28d84f480a5709806af43250fd56


lstebner (approved):
looks good to me!

PR #321, opened 2022-08-30
feat: Use relay servers for improved peer connectivity

jeremyckahn:
### What this PR does

This PR improves peer-to-peer connectivity by using a [public relay server](https://www.metered.ca/tools/openrelay/) when direct connections fail. This address the problem of peers that never connect successfully.

This video explains some of the networking specifics better than I can: https://www.youtube.com/watch?v=4dLJmZOcWFc

It depends on whether you experience peer connection issues without this change. Personally, I was unable to connect to peers when on a mobile network. After testing this in the preview environment, my mobile device was able to connect successfully.

This change depends on a third-party service that we don't control (https://www.metered.ca/tools/openrelay/). Ideally we'd host our own server to provide the relay functionality, but that would cost money. I'm open to self-hosting the relay server and WebTorrent pairing server in the future to increase control of our stack.

### Additional information

It's worth being aware of Open Relay's [terms and conditions](https://www.metered.ca/tools/openrelay/terms-and-conditions):

> THE TURN SERVER CAN AND WILL GO DOWN WITHOUT NOTICE AND WE DO NOT PROVIDE ANY GUARANTEES OF ANY KIND REGARDING ITS WORKINGS.

I believe that our peer connections will gracefully fall back to attempting direct peer connections as they currently do, in this case.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -38,3 +38,26 @@ for (const key of searchParams.keys()) {
     features[matches[1]] = true
   }
 }
+
+export const rtcConfig = {
+  iceServers: [
+    {
+      urls: 'stun:openrelay.metered.ca:80',
+    },
+    {
+      urls: 'turn:openrelay.metered.ca:80',
+      username: 'openrelayproject',
+      credential: 'openrelayproject',
+    },
+    {
+      urls: 'turn:openrelay.metered.ca:443',
+      username: 'openrelayproject',
+      credential: 'openrelayproject',
+    },
+    {
+      urls: 'turn:openrelay.metered.ca:443?transport=tcp',
+      username: 'openrelayproject',
+      credential: 'openrelayproject',
+    },
+  ],
+}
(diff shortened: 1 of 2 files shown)


src/config.js
```diff
@@ -38,3 +38,26 @@ for (const key of searchParams.keys()) {
...
+
+export const rtcConfig = {
+  iceServers: [
+    {
+      urls: 'stun:openrelay.metered.ca:80',
+    },
+    {
+      urls: 'turn:openrelay.metered.ca:80',
+      username: 'openrelayproject',
+      credential: 'openrelayproject',
```
jeremyckahn, line 50: Typically I wouldn't put credentials directly in the code like this, but this is effectively public information per [Open Relay's documentation](https://www.metered.ca/tools/openrelay/#-how-to-use).

lstebner (approved):
i didn't do any exhaustive testing but the code looks fine, and it sounds like you were able to test it. if you'd like to pair up on some more testing though hit me up on discord and i'll help out

jeremyckahn:
> i didn't do any exhaustive testing but the code looks fine, and it sounds like you were able to test it. if you'd like to pair up on some more testing though hit me up on discord and i'll help out

Thanks for the offer! Based on my own testing I think this should be in pretty good shape. It'll either connect or it won't, and I observed it connecting when it didn't before the change. That said, I'll merge and release this now.

Thanks for the review! 😄

PR #331, opened 2022-10-06
feat: Halloween Harvest achievement

jeremyckahn:
### What this PR does

This PR implements the "Halloween Harvest" achievement. When the player plays Farmhand in October, they will receive 150 Jack-o'-lanterns.

Open the game in October and confirm that the player is given 150 Jack-o'-lanterns.

### Additional information

Achievement screenshot

### DIFF
--- a/src/data/achievements.js
+++ b/src/data/achievements.js
@@ -104,6 +104,16 @@ const achievements = [
     reward: state => addItemToInventory(state, cowFeed, reward, true),
   }))(),
 
+  ((reward = 150) => ({
+    id: 'play-during-october',
+    name: 'Halloween Harvest',
+    description: 'Play Farmhand in October and get the gift of the season.',
+    rewardDescription: `${reward} units of ${itemsMap.jackolantern.name}`,
+    condition: () => new Date().getMonth() === 9,
+    reward: state =>
+      addItemToInventory(state, itemsMap.jackolantern, reward, true),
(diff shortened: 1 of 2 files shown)


src/data/achievements.js
```diff
@@ -104,6 +104,18 @@ const achievements = [
     reward: state => addItemToInventory(state, cowFeed, reward, true),
   }))(),
 
+  ((reward = 150) => ({
+    id: 'play-during-october',
+    name: 'Halloween Harvest',
+    description: 'Play Farmhand in October and get the gift of the season.',
+    rewardDescription: `${reward} units of ${itemsMap.jackolantern.name}`,
+    condition: () =>
+      // The environment check is necessary to ensure consistent test results.
```
lstebner, line 113: this makes me wonder if we should add a `getDate` or `getMonth` utility so we can mock it for tests. what do you think?

jeremyckahn: I think the standard `Date` constructor works well enough for our needs, but there's definitely value in mocking it to avoid special case logic like this. I've mocked out `getMonth` in https://github.com/jeremyckahn/farmhand/commit/b3ba38569db539fc013ad3d5f387db57bda96dae. What do you think about that approach?

lstebner: oops for some reason i left a PR level comment instead of in this thread but i'm good with that!

PR #334, opened 2022-10-09
fix quantity inputs allowing unacceptable values

lstebner:
### What this PR does

fixes a bug where the QuantityInput would start to count up while pressing the decrement button. it doesn't really do much other than move away from using `--value` in favor of a more functional approach which seems to fix the problem.

load up a fresh game and sell all but one of the jack-o-lanterns that are currently being put in the inventory (since it's october). then press 'decrement' and notice it does not start counting up like production currently is doing. after selling the jack-o-lanterns some fertilizer will have appeared, so press increment and decrement on that and make sure it is still functioning as expecting with the wrapping behavior.

### Additional information

this bug was reported in discord #bug-reports channel

### DIFF
--- a/src/components/QuantityInput/QuantityInput.js
+++ b/src/components/QuantityInput/QuantityInput.js
@@ -66,41 +66,61 @@ const QuantityInput = ({
   maxQuantity,
   setQuantity,
   value,
-}) => (
-  <div className="QuantityInput">
-    <QuantityTextInput
-      {...{ handleSubmit, handleUpdateNumber, maxQuantity, value }}
-    />
-    <span className="quantity">
-      /{' '}
-      <AnimatedNumber {...{ number: maxQuantity, formatter: integerString }} />
-    </span>
-    <div className="number-nudger-container">
-      <Fab
-        disabled={!value}
-        {...{
-          'aria-label': 'Increment',
-          color: 'primary',
-          onClick: () => setQuantity(++value > maxQuantity ? 1 : value),
-          size: 'small',
-        }}
-      >
-        <KeyboardArrowUp />
-      </Fab>
-      <Fab
-        disabled={!value}
-        {...{
-          'aria-label': 'Decrement',
-          color: 'primary',
-          onClick: () => setQuantity(--value === 0 ? maxQuantity : value),
-          size: 'small',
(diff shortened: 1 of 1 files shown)


src/components/QuantityInput/QuantityInput.js
```diff
@@ -66,41 +66,61 @@ const QuantityInput = ({
...
-      /{' '}
-      <AnimatedNumber {...{ number: maxQuantity, formatter: integerString }} />
-    </span>
-    <div className="number-nudger-container">
-      <Fab
-        disabled={!value}
-        {...{
-          'aria-label': 'Increment',
-          color: 'primary',
-          onClick: () => setQuantity(++value > maxQuantity ? 1 : value),
```
jeremyckahn, line 84: I just realized how poorly I initially implemented this. These event handlers are mutating `value`, so it's no wonder we're experiencing broken behavior. I have no idea why I though this would be a good idea. 🤦‍♂️

jeremyckahn (approved):
Thank you for the quick fix @lstebner! Functionally this seems good to go. I saw a few minor stylistic issues, but once those are addressed feel free to merge and run a `patch` release!

src/components/QuantityInput/QuantityInput.js
```diff
@@ -66,41 +66,61 @@ const QuantityInput = ({
...
+}) => {
+  const decrementQuantity = () => {
+    var newValue = value - 1
+    if (newValue === 0) {
+      newValue = maxQuantity
+    }
+    setQuantity(newValue)
+  }
+
+  const incrementValue = () => {
```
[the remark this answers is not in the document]
lstebner: this was an accident, thanks for noticing! definitely agree they should match

src/components/QuantityInput/QuantityInput.js
lstebner: all good! kinda interesting one honestly. i've always kind of avoided this syntax cause it's less common and tricky to read, but i've never seen it actually cause a bug until now

jeremyckahn: It's a remnant from my C++ days in college. I should know better by now! 😂

PR #335, opened 2022-10-18
chore: Remove MacOS support

jeremyckahn:
### What this PR does

This PR removes MacOS app support. This is being done because Electron Builder (the toolchain we are using to generate native apps) [cannot enable app auto-updating without code signing](https://www.electron.build/auto-update):

> macOS application must be [signed](https://www.electron.build/code-signing) in order for auto updating to work.
> ...
> ### Where to Buy Code Signing Certificate
>
> ... Please note — Gatekeeper only recognises [Apple digital certificates](http://stackoverflow.com/questions/11833481/non-apple-issued-code-signing-certificate-can-it-work-with-mac-os-10-8-gatekeep).

[Additionally](https://stackoverflow.com/a/24464807):

> you need to ... pay $99 per year for the Apple developer program so you can get an Apple certificate.

We'll have to run a release and ensure that the "Release New Version" workflow succeeds without generating MacOS artifacts.

This is a bummer of a change, but I think it's important to keep Farmhand fully operational with zero costs. That can't be achieved with MacOS, so unfortunately we won't be able to serve those players as well as those on other platforms. The good news is that Farmhand can be installed as a PWA on any platform, and that will provide a native app experience for all!

### DIFF
--- a/.github/workflows/run-release.yml
+++ b/.github/workflows/run-release.yml
@@ -116,7 +116,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest]
     steps:
       - uses: actions/checkout@v2.3.1
         with:
--- a/README.md
+++ b/README.md
@@ -8,8 +8,10 @@
 
 **[Play Farmhand in your browser!](https://www.farmhand.life/)**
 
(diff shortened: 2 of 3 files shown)


jeremyckahn:
We talked about this [on Discord](https://discord.com/channels/714539345050075176/847884323330850847/1031966018358218762). @lstebner has an active Apple Developer account we can use (🙏) to sign the MacOS binaries, so we can consider broken MacOS auto-updates to be a bug with a fix.

#336 has been opened to address this.

PR #348, opened 2022-11-30
feat: Add Farmhand Go! Steam link

jeremyckahn:
### What this PR does

This PR adds a link to Farmhand Go!'s Steam store page.

Make sure the link works and that the copy reads well.


lstebner (approved):
reads well, and link works! good idea to add this

PR #350, opened 2022-12-01
feat: Show holiday header during December

jeremyckahn:
### What this PR does

This PR shows an alternate header on the Home screen during the month of December.

Ensure that the alternate header appears when playing during December and not during any other month. You can simulate another month by running something like the following in your browser's console:

```js
Date.prototype.getMonth = () => 0
```

And then navigating away from the Home page and then back to it.

### Additional information

Desktop view: 
Desktop screenshot

Mobile view:
Mobile screnshot


jeremyckahn:
Closed in favor of #351.

PR #355, opened 2022-12-05
feat: Improve peer re-connection

jeremyckahn:
### What this PR does

This PR incorporates https://github.com/dmotz/trystero/pull/30 to improve peer re-connection when playing online. Typically when an online player puts their computer to sleep and wakes it up again later, the player will not be able to reconnect to peers until they refresh the page. This is an attempt to enable players to reconnect automatically without a page refresh.

You'll need two computers to test this. On both computers, join the same Farmhand room. On computer A, turn off the WiFi. On computer B, the peer should disappear within about a minute. Turn computer A's WiFi back on. Within 1-2 minutes, computer A should successfully reconnect to computer B again. 

https://github.com/dmotz/trystero/pull/30 has not yet been accepted by Trystero's developer, so this is still experimental. However, I think it should be fairly safe to put into Production. If this ends up causing issues then rolling back (reverting this PR) should be quick, safe, and easy.

### Additional information

This does not improve re-connection for iOS Safari. See [this comment](https://github.com/dmotz/trystero/pull/30#issuecomment-1336566515) for more info. I'm still investigating how to improve things for iOS players.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@jeremyckahn/farmhand",
-      "version": "1.15.11",
+      "version": "1.15.13",
       "license": "GPL-2.0-or-later",
       "dependencies": {
         "@fortawesome/fontawesome-svg-core": "^1.2.27",
@@ -47,7 +47,7 @@
         "redis": "^3.0.2",
         "shifty": "^2.15.2",
         "source-map-explorer": "^2.3.1",
-        "trystero": "^0.11.1",
+        "trystero": "github:jeremyckahn/trystero#bugfix/29__clean-up-peers",
         "typeface-francois-one": "0.0.71",
         "typeface-public-sans": "^1.1.4",
         "uuid": "^3.4.0",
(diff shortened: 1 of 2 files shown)


jeremyckahn:
Looks like CI is failing. I'll address this!

lstebner:
tested this and I did seem the connection failed message while wifi was off and then it stopped when my wifi was turned back on. on my other computer I saw the "active player" count drop when computer B was offline, then go back up by 1 when back online. however, one thing I don't understand is why i see a count of 6 for active players, but i don't see 6 when i open the menu.

on computer B i actually only see one. on computer A i saw a list of 4, including computer B, and then when going offline that one went away and never came back when it came back online. i'm not really sure if this indicates a bug or just a misunderstanding of how things work on my part. happy to test some more when we're both around tomorrow!

image

jeremyckahn:
Thanks for testing this @lstebner! There's a few things going on here.

> one thing I don't understand is why i see a count of 6 for active players, but i don't see 6 when i open the menu.

The "Active players" count display in the menu represents how many people have polled the `/api/get-market-data` endpoint within the last 25 seconds: https://github.com/jeremyckahn/farmhand/blob/15d97be44cbed0cf618d602d5adcd8aeaf8726f6/api/get-market-data.js#L38-L56

This strategy is quite imperfect and frequently shows a different number than the amount of peers that are connected. This is because it represents different data. I'd like to change how this works in the future, but it's where we are today. It's set up like this because the REST API-based market system predates the client-side peer connection system. And because the peer connection system is so fragile, I've kept it in place because it's a (somewhat) more accurate indicator of how many people are playing online. Ideally, the peer connections would be stable enough where we can simply display how many peers are connected and not deal with this API-side player count system, which as we can see is inherently flawed.

Which brings us to this PR! :) This PR is not a silver bullet, as it doesn't fully solve peer connection fragility. However, it should at least help to improve things and save players from having to reload the page in order to reconnect to peers. There's still more to do here and I plan to iterate on this issue with future PRs.

> on computer B i actually only see one. on computer A i saw a list of 4, including computer B, and then when going offline that one went away and never came back when it came back online.

My guess is that your computer B (the one that went offline and then came back) isn't seeing the other peers because they aren't running this fix, or not enough time was given to reconnect (I've seen it take upwards of a minute to reconnect). As long as your two computers are reconnecting after some time, I think that shows that this fix is working. Unfortunately I think this is one of those changes where we just have to ship it and hope for the best because it needs to be running on everyone else's machine to fully take effect. 😕

PR #357, opened 2022-12-28
crow logic updates

lstebner:
### What this PR does

this change alters crow logic so that there is a maximum number of crows that can be spawned and those crows then pick from plots that currently have crops at random to destroy them. previously, all plots were subject to an attack based on a random chance which could result in a lot of crops being destroyed at once.

this change also updates the notification after crows destroy plots to mention the number of plots destroyed in a single notification instead of one per plot destroyed which would often result in a flood of notifications that were hard to read before they would disappear.

Plant crops and then advance days. If there is no scarecrow, you should see some crops destroyed and the new notification. There is also a random chance no crops will be destroyed over night. When a scarecrow is present, no crops should ever be destroyed.

If it's preferred to go back to a notification per destroyed crop, or to bundle them up in a different way (such as a notification per crop _type_ destroyed) I can make that change. I wasn't sure how much it really mattered to know about every single one since they kinda disappear faster than they can be read when there's a lot anyways.

There is currently no spec for the `applyCrows` logic so I will work on adding that as well.

### Additional information

Screen Shot 2022-12-28 at 10 06 51 AM

Screen Shot 2022-12-28 at 10 07 08 AM

### DIFF
--- a/src/game-logic/reducers/applyCrows.js
+++ b/src/game-logic/reducers/applyCrows.js
@@ -1,5 +1,5 @@
 import { doesPlotContainCrop } from '../../utils'
-import { MAX_CROWS } from '../../constants'
+import { CROW_CHANCE, MAX_CROWS } from '../../constants'
 import { CROWS_DESTROYED } from '../../templates'
 
 import { modifyFieldPlotAt } from './modifyFieldPlotAt'
@@ -14,8 +14,8 @@ import { fieldHasScarecrow } from './helpers'
 export const applyCrows = state => {
   const { field, purchasedField } = state
 
-  if (fieldHasScarecrow(field)) {
-    return field
+  if (fieldHasScarecrow(field) || Math.random() > CROW_CHANCE) {
+    return state
   }
 
   const newDayNotifications = [...state.newDayNotifications]
--- a/src/game-logic/reducers/applyCrows.js
+++ b/src/game-logic/reducers/applyCrows.js
@@ -1,4 +1,4 @@
-import { doesPlotContainCrop } from '../../utils'
+import { doesPlotContainCrop, isRandomNumberLessThan } from '../../utils'
 import { CROW_CHANCE, MAX_CROWS } from '../../constants'
 import { CROWS_DESTROYED } from '../../templates'
 
@@ -14,7 +14,7 @@ import { fieldHasScarecrow } from './helpers'
 export const applyCrows = state => {
   const { field, purchasedField } = state
 
-  if (fieldHasScarecrow(field) || Math.random() > CROW_CHANCE) {
+  if (fieldHasScarecrow(field) || !isRandomNumberLessThan(CROW_CHANCE)) {
     return state
   }
 
--- a/src/game-logic/reducers/applyCrows.js
+++ b/src/game-logic/reducers/applyCrows.js
@@ -1,10 +1,10 @@
-import { itemType } from '../../enums'
-import { itemsMap } from '../../data/maps'
-import { getPlotContentType } from '../../utils'
-import { CROW_CHANCE } from '../../constants'
-import { CROW_ATTACKED } from '../../templates'
+import { doesPlotContainCrop } from '../../utils'
+import { MAX_CROWS } from '../../constants'
+import { CROWS_DESTROYED } from '../../templates'
 
-import { fieldHasScarecrow, updateField } from './helpers'
+import { modifyFieldPlotAt } from './modifyFieldPlotAt'
+
+import { fieldHasScarecrow } from './helpers'
 
 // TODO: Add tests for this reducer.
 /**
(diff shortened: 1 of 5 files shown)


src/templates.js
```diff
@@ -19,6 +19,11 @@ import { itemsMap } from './data/maps'
  */
 export const CROW_ATTACKED = (_, crop) => `Oh no a crow ate your ${crop.name}!`
 
+export const CROWS_DESTROYED = (_, numCropsDestroyed) =>
```
jeremyckahn, line 22: nit: This template needs JSDoc annotations to be consistent with the other ones.

src/game-logic/reducers/applyCrows.js
```diff
@@ -1,39 +1,53 @@
...
 // TODO: Add tests for this reducer.
 /**
  * @param {farmhand.state} state
  * @returns {farmhand.state}
  */
 export const applyCrows = state => {
-  const { field } = state
+  const { field, purchasedField } = state
+
+  if (fieldHasScarecrow(field) || Math.random() > CROW_CHANCE) {
```
jeremyckahn, line 17: Would it make sense to use `isRandomNumberLessThan()` instead of the `Math.random()` check here?

src/game-logic/reducers/applyCrows.js
```diff
@@ -1,39 +1,53 @@
...
+  const plotsWithCrops = []
 
-  const updatedField = fieldHasScarecrow(field)
-    ? field
-    : updateField(field, plotContent => {
-        if (!plotContent || getPlotContentType(plotContent) !== itemType.CROP) {
-          return plotContent
-        }
+  field.forEach((row, y) =>
+    row.forEach((plotContents, x) => {
```
jeremyckahn, line 27: I think we can use [the `forRange` reducer](https://github.com/jeremyckahn/farmhand/blob/15d97be44cbed0cf618d602d5adcd8aeaf8726f6/src/game-logic/reducers/forRange.js#L10-L30) here to DRY this nested loop up a bit.

(Tangentially, I'm noticing that `forRange`'s JSDoc annotations are innacurate. I'll fix that in a seperate PR, but that shouldn't prevent it from being usable here.)

src/game-logic/reducers/applyCrows.js
```diff
@@ -1,39 +1,53 @@
...
+      if (doesPlotContainCrop(plotContents)) {
+        plotsWithCrops.push({ x, y })
+      }
+    })
+  )
 
-        const destroyCrop = Math.random() <= CROW_CHANCE
+  const numCrows = Math.min(
+    plotsWithCrops.length,
+    Math.floor(Math.random() * (purchasedField + 1) * MAX_CROWS)
```
jeremyckahn, line 36: I like how the crow maximum here scales! 👍️

jeremyckahn (changes requested):
Great work @lstebner! The logic looks good and the gameplay change feels right to me. I made a few suggestions related to code reuse, and as you mentioned this could use some unit test coverage. Once that's taken care of, this should be ready to ship!

src/game-logic/reducers/applyCrows.js
jeremyckahn: PR for JSDoc update: https://github.com/jeremyckahn/farmhand/pull/358

jeremyckahn:
@lstebner if you can review and approve #358, it can be merged and it should fix the CI breakage we're seeing with this PR.

src/game-logic/reducers/applyCrows.js
lstebner: yeah, i think so, though it's going to be `!isRandomNumberLessThan(CROW_CHANCE)` which for some reason feels awkward to me. i'm going to make the change for now but curious if you think that makes sense, or maybe the value for `CROW_CHANCE` should be changed to `.6` instead of `.4` though that doesn't seem right to me either.

src/game-logic/reducers/applyCrows.js
lstebner: approved! 

re: `forRange`... i see what you're thinking here but it feels a little strange because of the radius arg. i guess i could just use a very large number, or do something like use the field width? i could also see a util being added like `forEachPlot` that is similar to `forRange` but loops over every plot and calls a function with similar args as `forRange` does. sorta thinking out loud, but i'm open to whatever you think is best here.

lstebner: i updated to use `forRange` but it still feels a bit awkward to me. mainly because that reducer takes `state` as an arg and doesn't really simplify getting access to the plot contents. using it still requires the caller to know how to dig into `state.field` along with the requirement to return an updated `state` value. i think something like this would still be ideal, but if this is the only usecase for it then it's a little tricky to justify

```js

jeremyckahn: Thanks for giving `forRange` a try. I agree that using it here looks weird!

I really like the `forEachPlot` helper you've got sketched out here. I think it would make sense to add it as a utility function and use it here. On the other hand, that would be a pretty textbook case of premature refactoring. I think the simplest solution for this PR is to revert the use of `forRange` and put it back the way you had it. If you'd like to implement `forEachPlot`, I think it would be small and useful enough to justify expanding the scope of this PR a bit. Whichever solution you feel like going with is fine with me.

Sorry for going back and forth on this! I hadn't considered how using `forRange` here would actually look. 😅

jeremyckahn:
@lstebner now that #358 is merged, you'll want to either rebase off of `develop` or merge it into your branch to get CI green again.

src/game-logic/reducers/applyCrows.js
lstebner: updated this again! i ended up making `forEachPlot` but just local in this file. if it's ever needed elsewhere it can be pulled out to a util but i think just keeping it local is the way to go for now

lstebner:
> @lstebner now that #358 is merged, you'll want to either rebase off of `develop` or merge it into your branch to get CI green again.

i did rebase but there's also a spec failing because of the changes to this reducer that i need to look in to. thanks for that quick fix though!

jeremyckahn:
> > @lstebner now that #358 is merged, you'll want to either rebase off of `develop` or merge it into your branch to get CI green again.
> 
> i did rebase but there's also a spec failing because of the changes to this reducer that i need to look in to. thanks for that quick fix though!

Oh! I thought it was the same issue as before (a broken `npm install`). Carry on! 😅

src/game-logic/reducers/applyCrows.js
jeremyckahn: > it's going to be !isRandomNumberLessThan(CROW_CHANCE) which for some reason feels awkward to me.

I don't find it overly awkward, personally. In general, I try to follow [this principle](https://libquotes.com/eric-s-raymond/quote/lbt5w7w):

> Smart data structures and dumb code works a lot better than the other way around.

In terms of this situation, it means that the data (or at least Farmhand's `CONSTANT` values) should express the rules of the game as clearly as possible, even at the some expense of straightforwardness in the logic. There are exceptions to this of course, but I think it results in a more cohesive system overall.

That isn't a strongly-held opinion, it's just how I think about this stuff. I trust your judgment, so feel free to change `0.4` to `0.6` and drop the negation if you think that makes more sense. The more important thing here is using consistent solutions for similar problems (using `isRandomNumberLessThan` for dice rolls). 🙂

src/game-logic/reducers/applyCrows.js
jeremyckahn: I think that's a great solution! Good thinking.

src/game-logic/reducers/applyCrows.js
lstebner: that all makes sense. i decided to go with `isRandomNumberLessThan(1 - CROW_CHANCE)` ultimately. let me know what you think!

jeremyckahn (approved):
The new tests look great! Thanks for cleaning up the old ones as well.

This is a big improvement to how crows work in Farmhand, and the code is much better with this change. Thanks @lstebner! LGTM!

src/game-logic/reducers/applyCrows.js
jeremyckahn: Oh whoops, I just saw this. I think that's a good approach! :+1:

PR #362, opened 2022-12-30
chore: Use Create React App 5

jeremyckahn:
### What this PR does

This PR updates Farmhand to use Create React App 5. This required a number of changes throughout to get things working smoothly so that's in here as well.

Make sure the game boots and doesn't throw any errors. Also, run the main scripts (`start`, `test`, and `build`) to make sure there are no regressions.


src/components/CowPen/CowPen.test.js
```diff
@@ -40,19 +42,19 @@ describe('Cow', () => {
 
   describe('movement', () => {
     test('schedules a position change at boot', () => {
-      expect(setTimeout).toHaveBeenCalledWith(
+      expect(global.setTimeout).toHaveBeenCalledWith(
```
lstebner, line 45: even though you use `global` for the spy do you actually have to change these references? i think they are references to the global versions by default, but i'm not positive

lstebner (approved):
one question but i'm cool with it either way so going to approve

src/components/CowPen/CowPen.test.js
jeremyckahn: Yeah, this change was necessary to fix some newly-breaking tests. It seems that newer versions of Jest have changed how timers work. Accessing via `global` is the only way to spy on them as necessary for these tests (that I've found so far).

PR #363, opened 2022-12-31
[closes #118] add garlic and garlic recipes

### LINKED ISSUE #118: add crop: Garlic
this is just an idea, but i think garlic would be great for various recipes, and a fast growing/cheap crop.

lstebner:
### What this PR does

adds garlic crop, art, and two associated recipes

garlic was added as a level 34 unlock since there was nothing else there at the moment. however, jalapeno and potato are needed to unlock all the recipes.

### Additional information

Screen Shot 2022-12-31 at 11 22 37 AM
Screen Shot 2022-12-31 at 11 22 14 AM
Screen Shot 2022-12-31 at 11 22 08 AM
Screen Shot 2022-12-31 at 4 15 28 PM


src/data/crops/garlic.js
```diff
@@ -0,0 +1,29 @@
...
+export const garlicSeed = crop({
+  cropType: cropType.GARLIC,
+  cropTimetable: {
+    [SEED]: 2,
+    [GROWING]: 3,
+  },
+  growsInto: 'garlic',
+  id: 'garlic-seed',
+  name: 'Garlic Bulb',
+  tier: 3,
```
jeremyckahn, line 19: If we change this to be unlocked at a later level like 34 (see my comment in `levels.js`), it would make sense to increase `tier` to 5, as it would be the third tier 5 seed. Having about three items per tier (unlocked in order) seems to be a good balance so far.

src/data/levels.js
```diff
@@ -29,6 +29,10 @@ levels[6] = {
   unlocksTool: toolType.SHOVEL,
 }
 
+levels[7] = {
+  unlocksShopItem: items.garlicSeed.id,
+}
```
jeremyckahn, line 34: What do you think about keeping the level 7 reward slot open for now and moving this up to something like level 34? I think having gaps in the static level rewards brings a few advantages:

  - It makes for a more unique experience for each player earlier in the game, as [levels that don't have a predefined reward just give the player a random seed](https://github.com/jeremyckahn/farmhand/blob/5c87bb6e604cb00c9ed2dc289ec606e5fa862b1c/src/utils.js#L887-L892).
  - It adds to the late game experience which (hopefully) keeps players coming back more over the long term.

jeremyckahn (changes requested):
Thanks for getting to this longstanding ticket, @lstebner! The code looks good, but I have some suggestions for how we might reconfigure the level rewards and tier value to serve the late game experience more.

As for the new art, we've got a good start here. The new images could use some outlining to contrast them a bit against their background during gameplay. You struck a good balance with this in the Asparagus images, so that would be a good source of (literal!) prior art. Let me know if you'd like a hand with updating the new images to match the established style.

src/data/levels.js
lstebner: ahh yes i totally forgot about the random seed reward. i'm fine with changing it to a level 34 reward!

jeremyckahn (approved):
The code and config looks great! Thanks for making those changes and for updating the Garlic art. I think the Garlic Fries art could use an outline as well, but I won't hold up this PR over it. Nice work, LGTM!

PR #366, opened 2023-01-04
[closes #259] added scarecrow to inventory at initial state of the game

### LINKED ISSUE #259: Give new players a Scarecrow
Suggested by @lstebner [on Discord](https://discord.com/channels/714539345050075176/722286738344640552/950144300757434428):

> it might be nice to give players a scarecrow in their initial inventory. i often go to the shop and just buy up as many seeds as i can, then get pretty bummed when i realize i need a scarecrow and can't afford it. it would ease the learning curve a little so that it becomes an item you replace, not acquire then lose

karlerik-simon:
### What this PR does
It adds a scarecrow at the beginning of the game by including an item object at inventory of initial state

By running a new game, and the scarecrow will be seen in the inventory. You can also see the scarecrow at the side menu in the fields tab.

Currently the scarecrow is added but the $500 at the start still remains. Should it be reduced by $160 to account for the scarecrow?

Should addItemToInventory be used instead of "hardcoding" it into the initial state?

### DIFF
--- a/src/components/Farmhand/Farmhand.js
+++ b/src/components/Farmhand/Farmhand.js
@@ -101,6 +101,8 @@ import {
 } from '../../strings'
 import { endpoints, rtcConfig, trackerUrls } from '../../config'
 
+import { scarecrow } from '../../data/items'
+
 import { getInventoryQuantities } from './helpers/getInventoryQuantities'
 import FarmhandContext from './Farmhand.context'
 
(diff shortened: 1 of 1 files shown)


lstebner (approved):
this seems good to me though @jeremyckahn should definitely have the final say. nice work @karlerik-simon!

jeremyckahn (approved):
Excellent work @karlerik-simon! Congrats on your first Farmhand PR. We appreciate the help!

> Currently the scarecrow is added but the $500 at the start still remains. Should it be reduced by $160 to account for the scarecrow?

No, I think it makes sense to gift the player a bonus Scarecrow to get their farm started.

> Should addItemToInventory be used instead of "hardcoding" it into the initial state?

No, I think the solution you have here is as simple as it needs to be and should be maintainable.

LGTM! I'll merge and deploy this as a patch release.

PR #372, opened 2023-01-10
feat: [closes #369] add sweet potato

lstebner:
### What this PR does

adds sweet potato crop, sweet potato pie recipe, and sweet potato fries recipes. also adds vegetable oil, onion rings, and fried tofu

### How this change can be validated

sweet potato is a high level crop so need to get enough sales to be level 35 or higher. to unlock the recipes you need sales of at least 100 sweet potatoes for fries, and 200+ for the pie along with 250 wheat and 75 butter (same as pumpkin pie).

you can do this legitimately or enter this in the console (the 120,000 pumpkins are just to get to level 35):

```js
farmhand.setState({ itemsSold: Object.assign(farmhand.state.itemsSold, { pumpkin: 120000, 'sweet-potato': 200, wheat: 250, butter: 75 }) })
```

sell any one item through the UI after doing that to trigger the recipe logic and unlock the recipes.

### Questions or concerns about this change

As usual I am open to any suggestions/improvements to the art, ingredients, crop tier, growing schedule, etc.

### Additional information

note: sweet potato is not grown from seeds, but it is grown from rooted segments called "slips".

Screen Shot 2023-01-09 at 10 31 57 PM

Screen Shot 2023-01-09 at 10 38 17 PM

Screen Shot 2023-01-09 at 10 58 34 PM

Screen Shot 2023-01-10 at 4 04 32 AM
Screen Shot 2023-01-10 at 4 06 21 AM
Screen Shot 2023-01-10 at 3 57 33 AM

PR #377, opened 2023-01-11
feat: [closes #373] add olive crop and olive oil

### LINKED ISSUE #373: Olives and olive oil
Farmhand should have an Olive crop and a related Olive Oil recipe.

lstebner:
### What this PR does

adds olive crop, olive oil, and garlic bread

### Additional information

Screen Shot 2023-01-10 at 5 26 21 AM

Screen Shot 2023-01-10 at 5 26 27 AM

Screen Shot 2023-01-10 at 5 26 15 AM

Screen Shot 2023-01-11 at 8 19 48 PM


src/data/recipes.js
```diff
@@ -56,6 +56,20 @@ export const butter = itemify({
...
 
+/**
+ * @property farmhand.module:recipes.oliveOil
+ * @type {farmhand.recipe}
+ */
+export const oliveOil = itemify({
+  id: 'olive-oil',
+  name: 'Olive Oil',
+  ingredients: {
+    [items.olive.id]: 15,
```
jeremyckahn, line 67: Apparently it takes [an incredible amount of olives to produce olive oil](https://oliveemotion.com/en/how-many-olives-are-needed-to-make-one-liter-of-olive-oil/):

> To obtain 1 liter of extra virgin olive oil, 5 to 6 kg of olive are required, approximately 5200 to 8000 olives.

That might be a little extreme for what we're going for, but what do you think about upping the challenge here to be slightly closer to reality? Maybe ~250 olives?

src/data/recipes.js
```diff
@@ -56,6 +56,20 @@ export const butter = itemify({
...
+  },
+  condition: state => state.itemsSold[items.olive.id] >= 150,
```
jeremyckahn, line 69: If we significantly increase the amount of olives need to produce oil, we should scale up the condition logic to be consistent with the other recipes. Perhaps requiring 400-500 olives to be sold?

jeremyckahn (commented):
Great work here! I just pushed https://github.com/jeremyckahn/farmhand/commit/c7670e477498931e56e8507791413492851d98e2, which centers the `olive-growing` image:

Olive Growing preview

The code and art all look great, but I wanted to get your thoughts on tweaking the Olive Oil values before we merge.

src/data/recipes.js
lstebner: oh wow, considering most recipes that use olive oil probably only need one i'm cool with upping it. it's going to be worth an insane amount based on that so we should probably also change the name to "extra virgin olive oil" haha. 

this also makes me wonder if the vegetable oil should have used more soybeans.... i've merged that but maybe we should alter it before releasing the update. thoughts?

src/data/recipes.js
lstebner: yep that makes sense to me!

src/data/recipes.js
jeremyckahn: These both sounds like good ideas to me! #372 is merged but not yet released, so tweaking it should be no problem.

lstebner: looks like producing vegetable oil from soybeans is even more extreme haha. 7-8kg per liter is what i found, and an average bean weighs like .15g. granted a soybean plant produces more than just one bean, it's still dozens of thousands. our soybean crop does grow quickly though, so maybe like 300 or 400? 400 sounds like a lot but it's less than fully filling the largest field twice so maybe that's ok?

lstebner: i split the difference and went 350 soybeans for veggie oil; and went with your suggestion of 250 for olive oil

lstebner:
@jeremyckahn updated recipe ingredient amounts and rebased to resolve a conflict from my sweet potato branch merging. please re-review when you have time!

src/data/recipes.js
```diff
@@ -330,9 +344,9 @@ export const vegetableOil = itemify({
   id: 'vegetable-oil',
   name: 'Vegetable Oil',
   ingredients: {
-    [items.soybean.id]: 30,
+    [items.soybean.id]: 350,
```
jeremyckahn, line 347: 💯

src/data/recipes.js
jeremyckahn: That all seems balanced to me. Thanks for making the change!

jeremyckahn (approved):
Excellent work! Thanks for making all the tweaks. LGTM!

jeremyckahn (approved):
LGTM! Great work to everyone involved. I'll merge and ship this now!

PR #378, opened 2023-01-11
add sunflower crop and sunbutter recipe

lstebner:
### What this PR does

adds sunflower crop and sunbutter recipe

### Additional information

Screen Shot 2023-01-10 at 6 17 40 AM

Screen Shot 2023-01-10 at 6 17 26 AM

Screen Shot 2023-01-10 at 6 31 01 AM


src/data/recipes.js
```diff
@@ -56,6 +56,20 @@ export const butter = itemify({
...
 
+/**
+ * @property farmhand.module:recipes.sunButter
+ * @type {farmhand.recipe}
+ */
+export const sunButter = itemify({
+  id: 'sun-butter',
+  name: 'SunButter',
+  ingredients: {
+    [items.sunflower.id]: 8,
```
jeremyckahn, line 67: What do you think about increasing this to roughly 25-50 or so? I imagine that sunflower seed butter requires a lot of seeds in real life (though I couldn't find any hard figures), and it would be good to make this late-game content more challenging as well.

jeremyckahn (approved):
I love sun better! 😄 Good idea to add it. I wanted to propose increasing the sun butter requirements, but otherwise LGTM!

src/data/recipes.js
lstebner: >  it would be good to make this late-game content more challenging as well.

i definitely agree with this! i don't think sunbutter requires quite as extreme of an amount as the oils do since (i believe) the whole kernel becomes the butter. and apparently one sunflower can produce up to 2k of them!? i think 25ish fits the game well though.

src/data/recipes.js
```diff
@@ -56,6 +56,20 @@ export const butter = itemify({
   recipeType: recipeType.KITCHEN,
 })
 
+/**
+ * @property farmhand.module:recipes.sunButter
+ * @type {farmhand.recipe}
+ */
+export const sunButter = itemify({
+  id: 'sun-butter',
+  name: 'SunButter',
```
jeremyckahn, line 65: Looks like we need a space here:

```suggestion
  name: 'Sun Butter',
```

src/data/recipes.js
jeremyckahn: 25 seems like a good number to go with!

jeremyckahn (approved):
Nice work! These new items should brighten up the game a bit. 🌻

LGTM!

PR #386, opened 2023-01-28
fix: [fixes #385] Ensure valid breeding pen references

### LINKED ISSUE #385: Cow breeding pen broken by trading (potentially)
A user reported that their save file got corrupted and it caused the game to crash when navigating to the Cows page.

> i dont exactly recall what happened but i believe i clicked on offer and had two cows in breeding pen and the next time i loaded the game i was facing this error. Also for some reason i have the legendary rainbow cow in my shed and it says "Original owner: Cheerful Dove". 

My theory is that the player offered a cow for trade that was in their breeding pen, a connected peer tried to make a trade, but then the trade sequence wasn’t able to complete for some reason (likely one of the players closed the game or lost connection before it finished).

What’s puzzling about this is that the UI prevents the user from offering a cow for trade that’s already in the breeding pen, and likewise prevents a cow from being moved to the breeding pen that’s offered for trade. So, it’s unclear how this corrupt state was achieved.

Corrupt save file data: https://gist.github.com/jeremyckahn/69b9ee85c07d93bce035086b2271586d

jeremyckahn:
### What this PR does

This PR is a **speculative** fix for #385. I wasn't able to reproduce the bug that was reported, but after spending some time with the code I believe I've identified a possible race condition related to cow trading. This race condition could lead to the breeding pen data referencing cows that don't exist in the inventory, thus corrupting the data. This PR introduces a check to ensure that such an invalid reference is never made.

Ensure that there are no regressions related to moving cows into or out of the breeding pen or cow trading.

### Additional information

I plan to close #385 by merging this. I couldn't reproduce it and there's not much value in leaving the ticket open since we can't prove a negative. If this issue resurfaces, we can always reopen the ticket.

### DIFF
--- a/src/components/CowCard/Subheader/Subheader.js
+++ b/src/components/CowCard/Subheader/Subheader.js
@@ -57,11 +57,10 @@ const Subheader = ({
 
   const mateId = cowBreedingPen.cowId1 ?? cowBreedingPen.cowId2
   const mate = getCowMapById(cowInventory)[mateId]
+  const isEligibleToBreed = cow.gender !== mate?.gender
 
-  const isEligibleToBreed =
-    (!mate || cow.gender !== mate.gender) && !isThisCowOfferedForTrade
-
-  const canBeMovedToBreedingPen = isRoomInBreedingPen && isEligibleToBreed
+  const canBeMovedToBreedingPen =
+    isRoomInBreedingPen && isEligibleToBreed && !isThisCowOfferedForTrade
 
   const disableBreedingControlTooltip =
     !canBeMovedToBreedingPen && !isInBreedingPen
--- a/src/components/CowCard/Subheader/Subheader.js
+++ b/src/components/CowCard/Subheader/Subheader.js
@@ -46,28 +46,22 @@ const Subheader = ({
   handleCowAutomaticHugChange,
   handleCowBreedChange,
   huggingMachinesRemain,
-  id,
+  id, // Player ID, not cow ID
   isCowPurchased,
 }) => {
   const numberOfFullHearts = cow.happiness * 10
   const isInBreedingPen = isCowInBreedingPen(cow, cowBreedingPen)
-  const isBreedingPenFull =
-    cowBreedingPen.cowId1 !== null && cowBreedingPen.cowId2 !== null
+  const isRoomInBreedingPen =
+    cowBreedingPen.cowId1 === null || cowBreedingPen.cowId2 === null
+  const isThisCowOfferedForTrade = cowIdOfferedForTrade === cow.id
 
-  const isCowOfferedForTrade = !!cowInventory.find(
-    ({ id }) => id === cowIdOfferedForTrade
-  )
+  const mateId = cowBreedingPen.cowId1 ?? cowBreedingPen.cowId2
+  const mate = getCowMapById(cowInventory)[mateId]
 
-  let canBeMovedToBreedingPen = !isBreedingPenFull && !isCowOfferedForTrade
+  const isEligibleToBreed =
+    (!mate || cow.gender !== mate.gender) && !isThisCowOfferedForTrade
 
-  if (canBeMovedToBreedingPen) {
-    const potentialMateId = cowBreedingPen.cowId2 ?? cowBreedingPen.cowId1
-
-    if (potentialMateId !== null) {
-      canBeMovedToBreedingPen =
-        cow.gender !== getCowMapById(cowInventory)[potentialMateId].gender
-    }
-  }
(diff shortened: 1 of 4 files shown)


src/components/CowCard/Subheader/Subheader.js
```diff
@@ -46,28 +46,21 @@ const Subheader = ({
...
   const isInBreedingPen = isCowInBreedingPen(cow, cowBreedingPen)
-  const isBreedingPenFull =
-    cowBreedingPen.cowId1 !== null && cowBreedingPen.cowId2 !== null
+  const isRoomInBreedingPen =
+    cowBreedingPen.cowId1 === null || cowBreedingPen.cowId2 === null
+  const isThisCowOfferedForTrade = cowIdOfferedForTrade === cow.id
 
-  const isCowOfferedForTrade = !!cowInventory.find(
-    ({ id }) => id === cowIdOfferedForTrade
-  )
```
jeremyckahn, line 59: This wasn't causing the issue that was reported, but I believe this logic is wrong and led to a broken and confusing UX. This deleted code is checking to see if _any_ cow is in the breeding pen, which is irrelevant. It should check if _the cow being rendered_ is in the breeding pen, which it is now with this change. I made a number of other changes in this file to simplify the logic, as it was hard to follow previously.

src/game-logic/reducers/changeCowBreedingPenResident.js
```diff
@@ -3,26 +3,30 @@ import { COW_GESTATION_PERIOD_DAYS } from '../../constants'
...
   let newCowBreedingPen = { ...cowBreedingPen }
 
-  if (doAdd) {
-    if (isPenFull || isCowInPen) {
+  const isCowInInventory = !!cowInventory.find(({ id }) => {
+    return id === cow.id
+  })
+
+  if (isAdding) {
+    if (!isCowInInventory || isBreedingPenFull || isCowInBreedingPen) {
```
jeremyckahn, line 22: This change is the crux of this PR. Previously this code assumed that `cow` was in the cow inventory, but there is no guarantee of that being the case in the context of this reducer. Now it will simply bail out if the cow being moved to the breeding pen is not in the inventory.

src/handlers/peer-events.js
```diff
@@ -91,7 +91,7 @@ export const handleCowTradeRequest = async (
 
       if (!peerMetadata) {
         console.error(`No data for peer ${updatedCowOffered.ownerId}`)
-        return
+        return null
```
jeremyckahn, line 94: Trivial change for consistency.

src/components/CowCard/Subheader/Subheader.js
```diff
@@ -46,28 +46,21 @@ const Subheader = ({
   handleCowAutomaticHugChange,
   handleCowBreedChange,
   huggingMachinesRemain,
-  id,
+  id, // Player ID, not cow ID
```
lstebner, line 49: maybe rename this variable? if not everywhere, just here, i.e. `id: playerId`

src/components/CowCard/Subheader/Subheader.js
lstebner: `isCowOfferedForTrade` looks like it could have been causing the bug if `id` has always been "player id" in reality. cause it would potentially never find the cow in the pen based off that. right?

src/game-logic/reducers/changeCowBreedingPenResident.js
lstebner: 3 conditions is my tipping point for a var or a function, but i also won't hold up merge over this. a `cowCanBeAdded` would sum things up nicely though, and would also allow us to remove a level of nesting. i.e. `if (isAdding && cowCanBeAdded)`

lstebner (approved):
have a couple of comments but nothing major. i'm going to approve so you can change what you want, or leave it as is, and then merge. happy to take another look though if you want!

lstebner (approved):

src/game-logic/reducers/changeCowBreedingPenResident.js
jeremyckahn: This is a great call out! I've simplified this all in https://github.com/jeremyckahn/farmhand/commit/d3d0a0cc66aed8d96248c859ef5381093250d5b2. I also de-nested and cleaned up a bit of other related logic here.

I think your philosophy of using three condition variables as an indicator to refactor into a function is a really good one and I'll keep that in mind going forward!

jeremyckahn (commented):
Thanks for the helpful review @lstebner! Since this was previously approved and I feel good about the latest changes, I'm going to merge and release this now in order to (hopefully!) fix #385.

Issue #387, opened 2023-01-29
Rename `id` to `playerId`


jeremyckahn:
> I think it would make sense to rename `id` to `playerId` globally, but that's a very wide-ranging change that would be best saved for a focused followup PR.

_Originally posted by @jeremyckahn in https://github.com/jeremyckahn/farmhand/pull/386#discussion_r1090029705_

SujayDas1999:
Hi @jeremyckahn 
If this is not being worked upon yet, I would like to take on this issue :)

Henrichy:
I agree that renaming id to playerId would improve readability and reduce ambiguity. Given the scope of the change, it might be best to address it in a separate, focused PR to keep this one easier to review.

jeremyckahn:
@Henrichy agreed. This actually seems like a tedious task that would be best left to an AI to automate away, so I'll see if I can I do that.

(I forgot about this issue until now! 😅)

Henrichy:
So it's best I leave it right

Henrichy:
I can still assist if you don't mind

Henrichy:
Ok I'll be expecting your feedback

PR #390, opened 2023-01-29
Commit for testing

tarunsamanta2k20:
### Additional information

#387


jeremyckahn:
It looks like this change is just a global substitution of `id` for `playerId`. This won't suffice for #387, as `id` is used for many identifiers beyond just the player's. It is also used for IDs of cows, items, recipes and many other data types. We'll need to scope the changes to just rename `id` to `playerId` in the context of the top-level game state object's `id` variable. Additionally, we will need to migrate old game saves to the new format in [`transformStateDataForImport`](https://github.com/jeremyckahn/farmhand/blob/8d5c4acd68e97a550cbb93189a9b4392259196d1/src/utils.js#L1027-L1094).

PR #393, opened 2023-02-02
feat: Grapes

jeremyckahn:
### What this PR does

This PR adds Grape crops. There are ten types of Grapes, and they grow randomly from the one Grape Seed item. In order to support this, the crop system was extended to generically support varieties.

It's easiest to set yourself up with a high level, lots of money, and no inventory limit to test this:

```js
farmhand.setState({
  itemsSold: { carrot: 200000 },
  money: 10000000,
  inventoryLimit: -1
}) 
```

From there buy Grape Seeds from the Shop and go through a few harvest cycles. It would be good to keep an eye on the price events to see that the various Grape types show up there appropriately (though I've tested that and it seemed stable to me).

Grapes are unlocked at level 42. Would it be worth introducing this more exotic item earlier in the game by swapping it for another, lower-tiered item?

### Additional information

<details>
<summary>Gameplay screenshots</summary>

Planted grape seeds

Growing grapes

Mature grapes

Grapes in the inventory

</details>

The addition of support for crop varieties touched a _lot_ of code, so this is a rather large PR. Subsequent PRs adding new crops with varieties should not be nearly so large.

In particular, this PR updates a lot of tests. Broadly speaking, Farmhand's unit tests are overly dependent on mock item data. We probably shouldn't be doing this at all going forward. It creates needless abstraction, and much of the game's item data is stable enough to build tests on top of. It seems worth migrating tests to use real item data as we update them, and that long-term effort is started here.

This PR also introduces the use [JSDoc import types](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#import-types). While won't be migrating to TypeScript any time in the foreseeable future, we can still enjoy the editor conveniences of rich type system. This won't prevent us from misusing types, as that's a problem that only the complete adoption of a type system can provide. JSDoc import types are just an iterative step towards a better codebase, and it will also be an ongoing improvement effort to enhance files with as we touch them. To be clear: This isn't a new development requirement, just a handy thing we can take advantage of if we feel like it as we go.

<details>
<summary>Here's what we get in an editor that's configured to take advantage of TypeScript (even though these are JavaScript files)</summary>

[JSDoc import type demo](https://user-images.githubusercontent.com/366330/216369215-4bf963a8-0dee-4a0c-a84a-2601d537724e.webm)

</details>


src/data/crops/grape.js
```diff
@@ -0,0 +1,126 @@
...
+    'grape-chardonnay',
+    'grape-sauvignon-blanc',
+    'grape-pinot-blanc',
+    'grape-muscat',
+    'grape-riesling',
+    'grape-merlot',
+    'grape-cabernet-sauvignon',
+    'grape-syrah',
+    'grape-tempranillo',
+    'grape-nebbiolo',
```
lstebner, line 28: this is really neat to think of how many varieties can exist, but i can't help but wonder if this is too many to add. i feel like 3 or 4 would be a pretty good amount, but it's ultimately up to you

lstebner (approved):
> In particular, this PR updates a lot of tests. Broadly speaking, Farmhand's unit tests are overly dependent on mock item data. We probably shouldn't be doing this at all going forward. It creates needless abstraction, and much of the game's item data is stable enough to build tests on top of. It seems worth migrating tests to use real item data as we update them, and that long-term effort is started here.

strongly agree! 

honestly everything here looks reasonable and solid. my only concern is that this is too many varieties, but no issues with the code and i think it's ultimately up to you on the number of varieties. i hope we can utilize this with more crops in the future! or maybe even some of the pre-existing ones like corn, and carrots.

as far as if we should add this earlier in the game level wise i don't necessarily think we need to. i still think we should give players the ability to level up quicker so more of them get to see more content, but i can't think of any specific advantage of having grapes earlier than something else. at least, not until we have kegs in the workshop.

src/data/crops/grape.js
jeremyckahn: That's a fair point. 10 varieties is a little overwhelming now that you mention it! I decided to halve the varieties to 5 total in https://github.com/jeremyckahn/farmhand/commit/58a5d9a2fc2c7b366bc2cda7bf726dc94e5c7466. The unused varieties are commented out for now so we can easily add them back in down the line. 🙂

jeremyckahn:
> my only concern is that this is too many varieties, but no issues with the code and i think it's ultimately up to you on the number of varieties.

Agreed. I've cut out half of them for now and this is what remains:

<details>
<summary>Screenshot</summary>

Final grape variety list

</details>

> i still think we should give players the ability to level up quicker so more of them get to see more content, but i can't think of any specific advantage of having grapes earlier than something else. at least, not until we have kegs in the workshop.

:100:! That all makes sense to me.

PR #397, opened 2023-03-14
feat: [closes #383] Random number generator seeding

### LINKED ISSUE #383: RNG seeding
The random number generator should be [seedable](https://en.wikipedia.org/wiki/Random_seed). This will enable consistent behavior for speed runs.

We could consider using this library to implement this functionality with: https://github.com/davidbau/seedrandom

jeremyckahn:
### What this PR does

This PR implements [random number generation seeding](https://en.wikipedia.org/wiki/Random_seed). The seed state is stored as a URL query parameter rather than the `farmhand.state` object so that seeded URLs can be shared.

This is a feature designed for speedrunners. With a guaranteed sequence of random numbers, Farmhand's behavior can become completely predictable, thus enabling consistent speedrun experiences.

In a Private/Incognito browser session, open the preview URL with an embedded seed query param: https://farmhand-git-feature-383rng-seed-jeremyckahn.vercel.app/?seed=test

Observe that the value of Carrot Seed in the Shop is $21.10. Feel free to test other seed values, this is just a baseline.

Then in another Private/Incognito browser session, open the preview URL without an embedded seed query param: https://farmhand-git-feature-383rng-seed-jeremyckahn.vercel.app/

Observe that the value of Carrot Seed in the Shop is a random value for each new Private/Incognito session.

### Additional information

Random seed UI


src/common/services/randomNumber.js
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * @type {Function?}
+   */
+  seededRandom = null
+
+  constructor() {
+    // The availability of window.location needs to be checked before accessing
+    // its .search property. This code runs in both a browser and Node.js
+    // context, and window.location is not defined in Node.js environments.
+    const initialSeed = new URLSearchParams(window.location?.search).get('seed')
```
lstebner, line 14: you don't need to check for the existence of `window` though? that's what i more commonly see 🤔

src/components/SettingsView/RandomSeedInput.js
```diff
@@ -0,0 +1,52 @@
+import React, { useContext, useState } from 'react'
+import { string } from 'prop-types'
+import window from 'global/window'
+import TextField from '@material-ui/core/TextField'
+
+import FarmhandContext from '../Farmhand/Farmhand.context'
+
+import './RandomSeedInput.sass'
+
+export const RandomSeedInput = ({ search = window.location.search }) => {
```
lstebner, line 10: does this one have to worry about `window` not existing?

src/components/SettingsView/RandomSeedInput.js
```diff
@@ -0,0 +1,52 @@
...
+   * @param {React.SyntheticEvent<HTMLFormElement>} e
+   */
+  const handleSubmit = e => {
+    e.preventDefault()
+
+    handleRNGSeedChange(seed)
+  }
+
+  return (
+    <form className="RandomSeedInput" onSubmit={handleSubmit}>
```
lstebner, line 36: pascal case className?! :sus:

src/handlers/ui-events.js
```diff
@@ -469,4 +471,30 @@ export default {
...
   },
+
+  /**
+   * @param {string} seed
+   */
+  handleRNGSeedChange(seed) {
+    const { origin, pathname, search, hash } = window.location
+    const queryParams = new URLSearchParams(search)
+
+    if (seed === '') {
```
lstebner, line 482: what do you think about doing `seed.trim()` here? or does it make sense to allow a seed that's empty spaces?

src/react-app-env.d.ts
```diff
@@ -0,0 +1,7 @@
+/// <reference types="react-scripts" />
+
+// TODO: Contribute type definitions for 'global' package to
```
lstebner, line 3: have you filed an issue for this? i have a general rule of thumb to link issues to TODO notes to attempt to prevent them from being lost

src/utils/isRandomNumberLessThan.js
```diff
@@ -1,8 +1,11 @@
+import { random } from '../common/utils'
+
+// TODO: Migrate this to src/common/services/randomNumber.js
```
lstebner, line 3: same question here about having an issue

lstebner (commented):
overall i like this approach! just a few questions/comments for you

src/react-app-env.d.ts
jeremyckahn: Good idea! Done in https://github.com/jeremyckahn/farmhand/commit/24f0a6ae84d5261217b8e24dae63995b92ea1d4e.

src/utils/isRandomNumberLessThan.js
jeremyckahn: :heavy_check_mark: https://github.com/jeremyckahn/farmhand/commit/24f0a6ae84d5261217b8e24dae63995b92ea1d4e

src/components/SettingsView/RandomSeedInput.js
jeremyckahn: Hah that's the standard Create React App format for top-level component CSS classes. 🙂

https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/cra-template/template/src/App.js#L6-L7

src/handlers/ui-events.js
jeremyckahn: Oh great catch! I think trimming here is a smart idea. Fixed in https://github.com/jeremyckahn/farmhand/commit/933a857e8b0d28d7e0b61e0d2f922fbbed856a1b!

src/components/SettingsView/RandomSeedInput.js
jeremyckahn: Nope, it's imported up on line 3. It's not strictly necessary, but I've found using the `global/window` module to be a helpful practice, especially for test mocks.

src/common/services/randomNumber.js
jeremyckahn: Not with the `global/window` module being imported on line 2. 🙂

jeremyckahn (commented):
Thanks for the helpful review @lstebner! I think we should be set now. It's getting late for me so I'll let this sit and wait to merge and release until tomorrow morning. Feel free to add any comments in the meantime!

src/components/SettingsView/RandomSeedInput.js
lstebner: huh, i don't think i knew that. i've been doing all css modules work for the last few years now, but always did snake case before that. today i learned

src/components/SettingsView/RandomSeedInput.js
lstebner: ohhh, i see what you did now, i like it. nice one

lstebner (approved):

PR #409, opened 2023-04-13
refactor: [closes #400] Consolidate random number logic

### LINKED ISSUE #400: Migrate `isRandomNumberLessThan` to `src/common/services/randomNumber.js`
_Originally posted by @lstebner in https://github.com/jeremyckahn/farmhand/pull/397#discussion_r1139084327_

jeremyckahn:
### What this PR does

This housekeeping PR moves `isRandomNumberLessThan` to `RandomNumberService`.

Ensure that mining and crow functionality is unchanged.


lstebner (approved):
i don't see why this PR would have changed anything related to unlocking the shovel, but i couldn't get it to unlock using the preview URL. here you can see i'm level 6 and have no shovel still - 
image

did i miss something? 🤔 i'm going to approve nonetheless cause i think this is probably not a real issue, but curious if you can figure out what happened to me here.

jeremyckahn:
> i don't see why this PR would have changed anything related to unlocking the shovel, but i couldn't get it to unlock using the preview URL.

Good catch! I looked into this, and it looks like this is a quirk of how we tend to override the game state for the sake of testing. Personally (and I'll bet you do something similar), I just set state similarly to this to immediately reach specific levels:

```js
window.farmhand.setState({ itemsSold: { carrot: 10000 } })
```

And then trigger [`processLevelUp`](https://github.com/jeremyckahn/farmhand/blob/7077d0231e6acaf326e4a1b9d6ee1f0eb5defb3c/src/game-logic/reducers/processLevelUp.js#L21-L70) by selling an item. This is usually good enough, but it won't unlock all the level up rewards properly. In the case of the Shovel, you'll need to do something like this:

```js
window.farmhand.setState({ inventory: [{ id: 'carrot', quantity: 50 }], itemsSold: { carrot: 2499 } })
```

2,499 is one sale below level 6. From here, just sell a Carrot and you'll get the Shovel as you normally would.

This took me longer to figure out than I care to admit. 😅 In any case, this PR seems ready to go so I'll merge it now. Thanks for the review @lstebner!

Issue #380, opened 2023-01-12
add kegs


lstebner:
this idea came up in a [discussion in discord](https://discord.com/channels/714539345050075176/1062561823733325834/1062861056927600911) and needs fleshed out a bit more. i am summarizing the idea here though so it doesn't get lost.

"kegs" could go into a new area of the Workshop and allow for various aged beverages to be made. unlike the recipes that currently exist in the workshop, kegs could be "filled" with an item and then produce a beverage after some number of days (which could vary per beverage). for example, by putting some grapes into one wine could be produced after a few weeks. the produced beverage would have a substantially higher value than the original crop used. 

idea: the player could buy an infinite number of kegs. they could increase in price each like storage does and the kegs screen of the workshop would have a list of all the kegs. each can be filled separately and be on its own timeline. 

alternate idea to infinite kegs: the player purchases a "cellar" that can both be expanded (similarly to the field) in the number of kegs it has, and the keg quality can be upgraded to allow for faster aging.

there should also be some achievements associated to using kegs. i'm sure there's more to consider here so please add any ideas or thoughts you have!

jeremyckahn:
This is a great idea! I think the cellar route might be a bit more scalable than kegs. Treating cellar space similarly to Field or Cow Pen size seems like it would work well and be consistent with the rest of the game. The UI/UX for managing beverages would be simpler if they were treated in bulk, but also the CPU resources needed for unbounded kegs might become unmanageable over time and cause performance issues.

(This is also why we've avoided unbounded Field expansion: The bigger the Field gets, the larger the memory needs and save file becomes. Farmhand needs to run well on lower-powered devices, so it's worth taking that into account when designing new features like this.)

If it's sufficiently complex, we may also want to consider building this feature's UI as a new page (similarly to the Field or Cow Pen), rather than as a tab in the Workshop.

PR #410, opened 2023-04-25
feat: [closes #408] Salt

### LINKED ISSUE #408: Salt
To support fermentation recipes, we will need a salt item. Salt Rocks should obtainable through mining and convertible into Cooking Salt.

jeremyckahn:
### What this PR does

For #408 (and in service of #380), this PR adds Salt Rock mining and a Salt recipe. Additionally, it:

  - Adds Salt as an ingredient to some preexisting recipes
  - Refactors resource factory code to have a shared (pseudo) interface
  - Tweaks Stone spawning logic to make Stones spawn slightly less frequently
    - This is done to balance against the additional spawning of Salt Rocks

To quickly give yourself the Shovel and a lot of money, first set yourself just below level 6:

```js
window.farmhand.setState({
  itemsSold: { carrot: 2499 },
  inventory: [{ id: "carrot", quantity: 50 }],
  money: 1000000,
})
```

And then sell a Carrot to unlock the Shovel.

From there, go to the Field and mine until you get a Salt Rock. To unlock the Salt recipe, sell 30 Salt Rocks. Salt should also be shown as an ingredient for:

  - French Onion Soup
  - Hot Sauce
  - Garlic Fries
  - Sweet Potato Fries
  - Onion Rings

I chose not to make Salt an unlock condition of the recipes that now use it as an ingredient. If that wasn't the case, some players would see their recipes mysteriously disappear until they sell some Salt. Additionally, letting the Salt-using recipes be unlocked earlier indicates that Salt exists, which is a mid-game item. So, hopefully this all balances out by giving players more goals to work towards.

It's worth mentioning that the addition of Salt as an ingredient will make recipes that now use it more valuable.

### Additional information

Salt Rock

Salt Recipe


src/factories/StoneFactory.js
```diff
@@ -1,20 +1,32 @@
...
     let resources = []
 
-    resources.push(this.spawnStone())
+    if (randomNumberService.isRandomNumberLessThan(STONE_SPAWN_CHANCE)) {
+      resources.push(this.spawnStone())
+    }
+
+    if (randomNumberService.isRandomNumberLessThan(SALT_ROCK_SPAWN_CHANCE)) {
+      resources.push(this.spawnSaltRock())
+    }
```
jeremyckahn, line 29: This conditional check is new. It's a very subtle change to Stone spawning logic that I expect most players won't even notice.

src/data/ores/index.js
```diff
@@ -4,3 +4,4 @@ export { goldOre } from './goldOre'
 export { ironOre } from './ironOre'
 export { silverOre } from './silverOre'
 export { stone } from './stone'
+export { saltRock } from './salt-rock'
```
lstebner, line 7: the "salt-rock" filename doesn't match the casing convention of the others, is that on purpose?

src/data/ores/salt-rock.js
```diff
@@ -0,0 +1,19 @@
...
+ * @property farmhand.module:items.saltRock
+ * @type {farmhand.item}
+ */
+export const saltRock = freeze({
+  description: 'A large chunk of salt.',
+  doesPriceFluctuate: true,
+  id: 'salt-rock',
+  name: 'Salt Rock',
+  spawnChance: SALT_ROCK_SPAWN_CHANCE,
+  type: itemType.STONE,
```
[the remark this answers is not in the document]
lstebner: i can't remember what the type is really used for, but thinking in terms of real world items i am fine with it. what will this affect in the game though? or, what would be different if we added a new type? anything?

src/data/recipes.js
```diff
@@ -529,6 +551,7 @@ export const onionRings = itemify({
     [vegetableOil.id]: 1,
     [items.wheat.id]: 5,
     [soyMilk.id]: 1,
+    [salt.id]: 3,
```
lstebner, line 554: it would be neat if our recipes had optional ingredients so that we could add something like this salt, but not make it required. then if you add it you get the extra value of the salt, but if you don't have it you can still make the recipe it just has a lower value. just a thought 😄

src/factories/StoneFactory.js
lstebner, line 29: now that we're spawning more things in here i see some room to optimize this. something like...

```js
// this can live outside of generate just in this file, it's more of a "config" i suppose
const spawnableResources = [
  [stone, STONE_SPAWN_CHANCE],
  [saltRock, SALT_ROCK_SPAWN_CHANCE],
  [coal, COAL_SPAWN_CHANCE],
]

//generate can then work like this, and there is no need for the "spawn" methods since they don't actually do anything special
generate() {
  return spawnableResources.map((item) => {
    const [resource, spawnChance] = item
  
    if (isRandomNumberLessThan(spawnChance)) {
      return resource
    }
  
    return null
  }).filter((resource) => resource)
}
```

src/game-logic/reducers/minePlot.js
```diff
@@ -33,29 +33,28 @@ export const minePlot = (state, x, y) => {
...
+  for (let resource of spawnedResources) {
+    state = addItemToInventory(state, resource)
   }
 
   state = modifyFieldPlotAt(state, x, y, () => {
     return {
       isShoveled: true,
       daysUntilClear,
-      oreId: spawnedOre ? spawnedOre.id : null,
+      oreId: spawnedResource?.id ?? null,
```
lstebner, line 57: won't the `spawnedResource?` bit already return null if it doesn't exist? i think the `?? null` might be unnecessary here

lstebner (approved):
i really like the new art and updates to the recipes, had a couple comments but as usual i don't think i need to hold up approving for these things because most are stylistic or questions and everything here seems to be working well!

src/game-logic/reducers/minePlot.js
jeremyckahn: `spawnedResource` could potentially be `undefined` here. If that's the case, `spawnedResource?.id` would be `undefined`, not `null`:

```js
const value = undefined
console.log(value?.property) // undefined
```

Setting `oreId` to `undefined ` probably wouldn't break anything, but I have a stylistic preference to use `null` when explicitly setting a non-truthy value like this (it implies that the value was set intentionally, to me).

src/factories/StoneFactory.js
jeremyckahn: Great idea! I riffed on this a bit and went something slightly different: https://github.com/jeremyckahn/farmhand/commit/6f34a327920d3394bd49121ebd545252936c0d8f

I like your filter approach, I just tend to reach for syntax-level constructs (like `for...of`) rather than method calls where I can. 🙂

src/data/recipes.js
jeremyckahn: I like that idea! However, I think it might break the current model of all instances of an item having the same price at any given time. We could introduce item variants, but that could be unwieldy for players.

I'm open to the idea for the future. We'd just need to work through the UX and game design challenges that it might introduce. 🙂

src/data/ores/index.js
jeremyckahn: Whoops! That was not intentional. This is fixed in https://github.com/jeremyckahn/farmhand/commit/3019b2feedec61ab2ae85fe87f791bf821268944.

Thanks for raising this!

src/data/ores/salt-rock.js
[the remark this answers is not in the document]
jeremyckahn: I don't think introducing `itemType.SALT_ROCK` would affect the gameplay experience at all. At least for the moment, the choice between `itemType.STONE` or a new `itemType.SALT_ROCK` is strictly an implementation detail that just affects how data is organized within the code.

My thinking around sticking with `itemType.STONE` for the Salt Rock was that it functions very similarly to the Stone, aside from having a fluctuating value and being available for recipes. Also, I think it's worth reigning in the quantity of `itemType`s we have so they don't become too hard to work with.

Let's keep this as-is for the moment and revisit in the future if necessary.

jeremyckahn (commented):
Thanks for the helpful review @lstebner! I made the suggested changes. I'll merge and release this now. 🙂

Issue #412, opened 2023-05-18
Laboratory with item fusion system


jeremyckahn:
[From DIscord](https://discord.com/channels/714539345050075176/714539345637408793/1108766901640888360), Farmhand should have an item fusion system.

> Imagine being able to infuse a recipe with a crop of your choosing. Like, Carrot Soup infused with Pumpkin. Or a Fertilizer infused with Iron for a higher crop yield. Endless possibilities!
> Maybe we could have a new "Laboratory" section of the game to support item experiments?

PR #413, opened 2023-05-21
feat(Cellar): [closes #380] Fermentation

### LINKED ISSUE #380: add kegs

jeremyckahn:
### What this PR does

This PR implements an MVP for crop Fermentation. With this new game mechanic, players can ferment crops into kegs, wait for the kegs to mature, and then sell the kegs. Every day that a keg stays in the cellar inventory without being sold, it increases in value. Keg value growth is unbounded.

1. Give yourself enough money to buy a Cellar:

```js
farmhand.setState({ money: 1_000_000 })
```

2. Purchase a Cellar in the Shop's Upgrades tab.
3. Navigate to the Cellar's screen (at the end of the screen sequence).
4. Observe the empty Cellar inventory.
5. Click the Fermentation tab and observe that there is at least a Carrot recipe. If you have insufficient salt, the "Make" button will be disabled.
6. Give yourself a lot of salt:

```js
farmhand.setState({ inventory: [...farmhand.state.inventory, { id: 'salt', quantity: 1000 }]})
```

7. Make a Fermented Carrot (or whatever crops you have available) and observe that the appropriate amount of Salt is removed from your inventory, as well as one unit of the recipe base crop.
8. Go to the "Cellar Inventory" tab and observe that the kegged crop is present (but cannot yet be sold).
9. Progress through enough days to make the kegged crop completely fermented (days remaining will be shown in the Keg card)
10. Once the kegged crop is mature, observe that its value rises with every passing day that it is not sold.
11. Sell the kegged crop and observe that the correct amount of money is added to the player's total.
12. Repeat as necessary for other crops.

### Additional information

Cellar Inventory tab

Fermentation tab

This feature is in MVP state but ready for an initial public release. There are a number of obvious ways this new functionality could be improved in future iterations:
- Cellar sorting and searching
- Fermentation of Workshop recipes (like Fermented Carrot Soup)


src/utils/memoize.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  get(key) {
+    return this.cache[key]
+  }
+
+  set(key, value) {
+    if (Object.keys(this.cache).length > this.cacheSize) {
+      this.cache = {}
+    }
```
thatsanicehat, line 36: If you stored your cache entries as an array and lookup via filter, you could bump values to the "top" of the cache as you read them so that when it does come time to make room, you could just bump off enough entries from the back half of the array to get under your threshold without needing to clear the whole thing, potentially reducing "lag spikes" as the whole cache re-hydrates! It might be a minimal performance hit of a few cycles on lookup, but might have more of the intended effect! Also, consider making the cache size a kb metric rather than a keylength one!

jeremyckahn: This is an interesting idea! 

In the interest of avoiding premature optimization, I think it would be best to first identify a scenario where the current caching strategy negatively affects the player's experience in a perceptible way. I haven't personally noticed that, but that doesn't mean nobody else does.

It's not clear that the proposed alternative makes sense for all scenarios because of the `O(n)` lookup that would be introduced (it's currently `O(1)`). However, a variety of caching approaches can coexist in the code base and one or the other could be used depending on the use case.

@thatsanicehat is this an idea you'd like to explore more? If so I can make a ticket for you. I'm not personally seeing the need for the change, but I'd be happy to work with you on a PR. Let me know!

------------------------------

For context, this code was moved from [`utils/index.js`](https://github.com/jeremyckahn/farmhand/blob/eea169a886501d95f3633d571bee70dbc69e0b16/src/utils/index.js#L125-L153) rather than added added in this PR, so I'd want to save any substantive changes to it for another PR.

src/components/Cellar/InventoryTabPanel.js
```diff
@@ -0,0 +1,71 @@
...
+
+import { TabPanel } from './TabPanel'
+import { Keg } from './Keg'
+
+/**
+ * @param {Object} props
+ * @param {number} props.index
+ * @param {number} props.currentTab
+ */
+export const InventoryTabPanel = ({ index, currentTab }) => {
```
lstebner, line 22: this component name is tripping me up a little. "inventory tab panel" feels like something generic, but this one here is specific to the cellar right? should it be `CellarTabPanel`?

src/components/FermentationRecipeList/FermentationRecipe.js
```diff
@@ -0,0 +1,175 @@
...
+   * @returns number
+   */
+  (cellarInventory, item) => {
+    return cellarInventory.reduce((acc, keg) => {
+      if (keg.itemId === item.id) {
+        acc++
+      }
+
+      return acc
+    }, 0)
```
lstebner, line 45: another way to consider writing this which i think is a bit more concise, but ultimately the result is the same would be:

```
return cellarInventory.filter((keg) => keg.itemId === itemId).length
```

src/data/maps.js
```diff
@@ -64,6 +64,17 @@ export const itemsMap = {
...
 }
 
+/**
+ * @type {Object.<string, farmhand.item>}
+ */
+export const fermentableItemsMap = Object.fromEntries(
+  Object.entries(itemsMap).filter(([itemId]) => {
+    const item = itemsMap[itemId]
+
+    return 'daysToFerment' in item
```
lstebner, line 74: why doesn't this use the `canBeFermented` property to find fermentable items?

lstebner (approved):
just a few minor comments. there's a lot here so I really focused on testing that things work rather than every single minute detail of the code. looks like you reorganized some utils too. all seems to be working well for me though so i'm going to approve. i think my main question about this all is the one i asked in discord, which is, are we going to eventually have the ability to have more complex fermentation recipes? or will it always just be that certain crops can be fermented into a fermented version of their original form?

lstebner (approved):
just a few minor comments. there's a lot here so I really focused on testing that things work rather than every single minute detail of the code. looks like you reorganized some utils too. all seems to be working well for me though so i'm going to approve. i think my main question about this all is the one i asked in discord, which is, are we going to eventually have the ability to have more complex fermentation recipes? or will it always just be that certain crops can be fermented into a fermented version of their original form?

jeremyckahn (commented):
Thanks for the review @lstebner! Your suggestions helped to make the code more straightforward. 🙂

> are we going to eventually have the ability to have more complex fermentation recipes? or will it always just be that certain crops can be fermented into a fermented version of their original form?

Minimally, I want to add support for ferment-able recipes. Players should be able to take a recipe that already exists like [Tofu](https://github.com/jeremyckahn/farmhand/blob/57035dc70ddcfc4269f3ad7b53038cf89a7b06d4/src/data/recipes.js#L287-L295) and ferment it similarly to how they now can with crops. That will come in a later iteration.

I like what you're getting at with more complex recipe creation and I think it would fit well with Farmhand. What's not clear to me is if the current Kitchen/Fermentation systems should be augmented to support such complex recipes (a mutative change), or if we should add a new dedicated system for them (an additive change). The mutative option would be more challenging to develop, but it may result in a more cohesive experience for players. Right now I'm leaning towards building out a new, separate system so we can take our time to figure out how to make it a good experience for players. Once it's stable, we could explore merging it into either of the Kitchen/Fermentation systems if it makes sense to do so. I'm open to other opinions on this! In any case, I opened #412 recently to explore something like what you're proposing, though that concept needs a lot more refinement before it's actionable.

--------------------------

I think this should be ready to release with the latest commits. I'll merge this and finally ship the Fermentation feature! 🎉

PR #427, opened 2023-07-24
refactor(types): Set up type checking

jeremyckahn:
### What this PR does

This PR begins Farmhand's adoption of [Typed JavaScript](https://depth-first.com/articles/2021/11/03/typed-javascript/). Typed JavaScript strikes a nice balance between TypeScript's full type safety and JavaScript's simplicity. The type system is defined entirely by JSDoc-style comments that are already pervasive throughout the codebase.

This transition is detailed in [the ADR that is included in this PR](https://github.com/jeremyckahn/farmhand/pull/427/files#diff-6988ca4f7717f2157365addbf2d15b2cad4d77b6be1fa011c7cdb8a4c6ee2345). Additionally, this PR contains the following changes:

- Introduction of a `tsconfig.json` file
- Refactoring of `src/utils/index.js` to be completely type-safe
  - This seemed like the most sensible starting point
- Removal of JSDoc site generation and related tooling
  - It wasn't really useful and it was getting in the way of defining accurate types
- Removal of some unused/unnecessary code

A broad, light smoke test should be sufficient. `getPlotImage` needed the most substantive refactoring, so it would be worth giving the Field some extra attention to make sure it looks correct.

The adoption of Typed JavaScript will be a long road and may never be complete. I don't think this is a problem, as the code base and development experience benefits from each step towards complete type safety. The practical drawback with this change is that type errors can be visually annoying in development environments until they are corrected.

I'm personally motivated to update legacy code to be type safe as I go, but **there is no expectation or requirement for anyone to do so** as they work on the project. The type system is not integrated with the CI system for now, and that would not be a consideration until all type errors are fixed (if that ever happens). For now it's just another static analysis tool to aid in development that is roughly adjacent to ESLint and Prettier.

### Additional information

This is how type errors present in Neovim and VSCode environments that are set up to support TypeScript:

Type errors in Vim

Type errors in VSCode

To reiterate, these errors **do not** cause build failures. They are solely static analysis hints shown during development.


package.json
```diff
@@ -42,16 +39,19 @@
...
     "@testing-library/user-event": "^13.2.1",
+    "@types/dinero.js": "^1.9.0",
+    "@types/lodash.sortby": "^4.7.7",
     "@types/markdown-table": "^2.0.0",
     "@types/react": "^17.0.2",
+    "@types/uuid": "^9.0.2",
     "@typescript-eslint/eslint-plugin": "^2.23.0",
     "@typescript-eslint/parser": "^2.23.0",
     "bittorrent-tracker": "^9.19.0",
+    "browserslist": "^4.21.9",
```
jeremyckahn, line 54: This was yelling at me while running tests, so I updated the config. This shouldn't affect too much.

src/utils/index.js
```diff
@@ -279,99 +307,111 @@ export const isItemAFarmProduct = item =>
...
- * @param {farmhand.plotContent} plotContent
+ * @param {farmhand.plotContent | farmhand.shoveledPlot | null} plotContents
  * @param {number} x
  * @param {number} y
  * @returns {?string}
  */
-export const getPlotImage = (plotContent, x, y) => {
-  if (plotContent) {
-    if (getPlotContentType(plotContent) === itemType.CROP) {
+export const getPlotImage = (plotContents, x, y) => {
```
jeremyckahn, line 343: This function needed the most refactoring, but it should be functionally the same as before.

src/utils/index.js
```diff
@@ -386,19 +426,22 @@ export const getFinalCropItemIdFromSeedItemId = (
...
       } else {
         return growsInto === cropItemId
       }
-    }).id
+    })?.id
+
+    if (!seedItemId)
+      throw new Error(
+        `Crop item ID ${cropItemId} does not have a corresponding seed`
+      )
```
jeremyckahn, line 442: This error should never be thrown in practice, but it was necessary for this function to be type safe. I expect we'll need to use this pattern more as we adopt Typed JavaScript.

src/utils/index.js
```diff
@@ -661,39 +714,25 @@ export const getPriceEventForCrop = cropItem => ({
...
+  /**
+   * @param {(?farmhand.plotContent)[][]} field
+   * @param {function(?farmhand.plotContent): boolean} condition
+   * @returns {?farmhand.plotContent}
+   */
+  (field, condition) => {
+    const [plot = null] =
+      field.find(row => {
+        return row.find(condition)
+      }) ?? []
```
jeremyckahn, line 727: This function was actually functionally broken before, as it didn't return what I intended. It simply hadn't manifested in a noticeable way yet. Typed JavaScript exposed a preexisting bug! 😃

src/utils/index.js
```diff
@@ -661,39 +714,25 @@ export const getPriceEventForCrop = cropItem => ({
...
 
-      return acc
-    }, []),
```
jeremyckahn, line 688: This wasn't being used, so I removed it.

src/utils/index.js
```diff
@@ -956,71 +996,15 @@ export const unlockTool = (currentToolLevels, toolType) => {
...
-    ...cow,
-  }))
-
-  sanitizedState.cowForSale = {
-    ownerId: '',
-    originalOwnerId: '',
-    timesTraded: 0,
-    ...sanitizedState.cowForSale,
-  }
-
```
jeremyckahn, line 1023: This function was long overdue for a cleanup, so I removed support for deprecated save files as part of this PR.

tsconfig.json
jeremyckahn, line 1: This is pretty standard `tsconfig.json` boilerplate. Its presence is what causes the editor to recognize type errors as issues to present during development.

jeremyckahn:
[We discussed on Discord](https://discord.com/channels/714539345050075176/847884323330850847/1134941421367988294) that the code looks good. I've QA'ed this thoroughly myself and there don't appear to be any regressions, so I'm going to merge this now.

src/utils/index.js
jeremyckahn: It appears that this change may have broken Scarecrows. A fix is at #431.

PR #428, opened 2023-08-03
refactor(types): Make `Farmhand` component type-safe

jeremyckahn:
### What this PR does

This PR:

- Fixes all of the errors in `src/components/Farmhand/Farmhand.js`
  - This required some considerable refactoring switching to patterns that are statically analyzable
- Updates CI job to check types
  - Currently this does not cause a build failure and is strictly informational
- Uses a global `noop` util

This is a pure refactor, so a broad smoke test for regressions should be all that's necessary.

I don't really love the new `FarmhandReducers` pattern, but I couldn't find any other way to bind reducers in a statically analyzable way.

### DIFF
--- a/src/components/Farmhand/Farmhand.js
+++ b/src/components/Farmhand/Farmhand.js
@@ -1,8 +1,17 @@
 /**
  * @typedef {import("../../index").farmhand.item} farmhand.item
  * @typedef {import("../../index").farmhand.cow} farmhand.cow
+ * @typedef {import("../../index").farmhand.cowBreedingPen} farmhand.cowBreedingPen
  * @typedef {import("../../index").farmhand.keg} farmhand.keg
+ * @typedef {import("../../index").farmhand.plotContent} farmhand.plotContent
+ * @typedef {import("../../index").farmhand.peerMessage} farmhand.peerMessage
+ * @typedef {import("../../index").farmhand.priceEvent} farmhand.priceEvent
  * @typedef {import("../../index").farmhand.notification} notification
+ * @typedef {import("../../enums").cowColors} farmhand.cowColors
+ * @typedef {import("../../enums").cropType} farmhand.cropType
+ * @typedef {import("../../enums").dialogView} farmhand.dialogView
+ * @typedef {import("../../enums").fieldMode} farmhand.fieldMode
+ * @typedef {import("../../enums").stageFocusType} farmhand.stageFocusType
  */
 import React, { Component } from 'react'
 import window from 'global/window'
@@ -153,10 +162,10 @@ export const getPlantableCropInventory = memoize(inventory =>
 )
 
 /**
- * @param {Object.<number>} valueAdjustments
+ * @param {Record<string, number>} valueAdjustments
  * @param {farmhand.priceEvent} priceCrashes
  * @param {farmhand.priceEvent} priceSurges
- * @returns {Object.<number>}
+ * @returns {Record<string, number>}
  */
 const applyPriceEvents = (valueAdjustments, priceCrashes, priceSurges) => {
   const patchedValueAdjustments = { ...valueAdjustments }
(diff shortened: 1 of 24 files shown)


src/components/Farmhand/Farmhand.js
jeremyckahn, line 1: Note: This is hidden by GitHub by default but it's the heart of this PR.

src/components/Farmhand/FarmhandReducers.js
```diff
@@ -0,0 +1,227 @@
...
+ * This class serves as a sort of middleware for the main Farmhand component.
+ * It defines stub methods with names that correspond to various reducer
+ * functions. At the time of instantiation, it implements the methods with the
+ * matching reducer function and binds it to the class instance.
+ *
+ * This is done in the interest of coding convenience and decentralization of
+ * core game logic. It is implemented as a class so that the methods can be
+ * statically analyzed by the TypeScript compiler.
+ */
+export class FarmhandReducers extends Component {
```
jeremyckahn, line 23: The goal with this weird pattern in to replace the old `initReducers` method with an alternative solution that can be statically analyzed: https://github.com/jeremyckahn/farmhand/blob/ea69b10b6338db9a272fd6db7e622c94e6033486/src/components/Farmhand/Farmhand.js#L544-L604

With this, we're able to still maintain just a single list of functions to be bound without redundant pass-through logic for each one (see the `constructor` below).

src/components/Farmhand/Farmhand.js
```diff
@@ -390,10 +419,16 @@ export default class Farmhand extends Component {
       cowInventory: [],
       cowsSold: {},
       cowsTraded: 0,
+      cowTradeTimeoutId: -1,
       cropsHarvested: {},
       dayCount: 0,
       farmName: 'Unnamed',
       field: createNewField(),
+      fieldMode: OBSERVE,
+      getCowAccept: () => {},
```
lstebner, line 428: super minor suggestion, but for readability i like to create a `const noop = () => {}` and then use that instead of creating all these empty functions as default values. i sometimes go as far as to make a `noop` util so it can be imported and used across a codebase, but i'm not strongly opinionated on doing that

src/components/Farmhand/Farmhand.js
```diff
@@ -537,69 +572,7 @@ export default class Farmhand extends Component {
...
-          throw new Error(
-            `Reducer ${reducerName} is not exported from reducers/index.js`
-          )
-        }
-      }
-
-      this[reducerName] = (...args) => {
-        this.setState(state => reducer(state, ...args))
-      }
+      waterAllPlots: () => this.waterAllPlots(),
```
lstebner, line 575: can this just be `waterAllPlots: this.waterAllPlots`? or is there a scope issue if you do that?

src/components/Farmhand/FarmhandReducers.js
lstebner: i really don't like inheritance when it comes to react components, but i do think what you have here is an improvement vs the way it used to be (which was a little too "magic" feeling for my tastes). i don't have a better suggestion off hand so i'm generally cool with what you've cooked up here, but i may think on it some more and see if i can come up with another suggestion

lstebner (approved):

src/components/Farmhand/Farmhand.js
jeremyckahn: It can, but I'd prefer to keep this as-is for clarity's sake. It would be a harmless change in this case, but this change would obfuscate any function parameters (even though there are none in this case):

```suggestion
      waterAllPlots: this.waterAllPlots,
```

Tangentially, this is a change I would have likely made in the past. However, I've found that clever micro-optimizations like this tend to make code harder to follow. There's still plenty of them throughout this code base, but I am gradually trying to make Farmhand's code more obvious and straightforward by allowing for slightly more verbosity. 🙂

src/components/Farmhand/Farmhand.js
jeremyckahn: I think that's a great call-out. A global `noop` util is long overdue. Added in https://github.com/jeremyckahn/farmhand/commit/6397358d4cff7e87f9827e49d9fe02cfabc6abe8!

src/components/Farmhand/FarmhandReducers.js
jeremyckahn: Yeah I'm not thrilled about this either for the same reasons. It's the least-bad solution I could come up with that was syntactically valid and type-safe. Let's run with this for now and revisit if something better comes to us.

PR #434, opened 2023-08-20
feat(online) [for #432] "Cow offered" notification

jeremyckahn:
### What this PR does

For #432, this PR adds a notification that is shown when an online peer offers a cow for trade. When the notification is clicked, the "Active Players" dialog is opened.

This PR also:
  - Updates Trystero to the latest version and increases tracker redundancy for improved peer connectivity.
  - Gives clickable notifications the `pointer: cursor` style.

Set up two peers (they can be on the same computer but in different browsers) and join the same online room. Once peers are connected, offer a cow for trade. Observe that the other peer gets a notification like the one pictured below.

### Additional information

Screenshot of new notification

### DIFF
--- a/src/game-logic/reducers/updatePeer.js
+++ b/src/game-logic/reducers/updatePeer.js
@@ -24,7 +24,7 @@ export const updatePeer = (state, farmhand, peerMetadata, peerId) => {
   const previousCowOfferedId = previousPeerMetadata?.cowOfferedForTrade?.id
   const newCowOfferedId = peerMetadata.cowOfferedForTrade?.id
 
-  const isCowNewlyBeingOfferedForTrade =
+  const isNewTrade =
     newCowOfferedId &&
     previousCowOfferedId !== peerMetadata.cowOfferedForTrade?.id
 
@@ -34,7 +34,7 @@ export const updatePeer = (state, farmhand, peerMetadata, peerId) => {
   // it here.
   const { pendingPeerMessages = [] } = peerMetadata
 
-  if (isCowNewlyBeingOfferedForTrade) {
+  if (isNewTrade) {
     state = showNotification(
       state,
       NEW_COW_OFFERED_FOR_TRADE`${peerMetadata}`,
--- a/src/game-logic/reducers/updatePeer.js
+++ b/src/game-logic/reducers/updatePeer.js
@@ -11,12 +11,12 @@ import { showNotification } from './showNotification'
 
 /**
  * @param {farmhand.state} state
- * @param {string} peerId The peer to update
- * @param {farmhand.peerMetadata} peerMetadata
  * @param {Farmhand} farmhand
+ * @param {farmhand.peerMetadata} peerMetadata
+ * @param {string} peerId The peer to update
  * @returns {farmhand.state}
  */
-export const updatePeer = (state, peerId, peerMetadata, farmhand) => {
+export const updatePeer = (state, farmhand, peerMetadata, peerId) => {
   const peers = { ...state.peers }
 
   const previousPeerMetadata = peers[peerId]
--- a/src/game-logic/reducers/updatePeer.js
+++ b/src/game-logic/reducers/updatePeer.js
@@ -1,18 +1,36 @@
+/**
+ * @typedef {import('../../components/Farmhand/Farmhand').farmhand.state} farmhand.state
+ * @typedef {import('../../index').farmhand.peerMetadata} farmhand.peerMetadata
+ */
 import { MAX_LATEST_PEER_MESSAGES } from '../../constants'
+
+import { showNotification } from './showNotification'
 
 /**
  * @param {farmhand.state} state
  * @param {string} peerId The peer to update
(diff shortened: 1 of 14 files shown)


src/game-logic/reducers/updatePeer.js
```diff
@@ -1,18 +1,49 @@
...
-  peers[peerId] = peerState
+
+  const previousPeerMetadata = peers[peerId]
+
+  const previousCowOfferedId = previousPeerMetadata?.cowOfferedForTrade?.id
+  const newCowOfferedId = peerMetadata.cowOfferedForTrade?.id
+
+  const isCowNewlyBeingOfferedForTrade =
+    newCowOfferedId &&
+    previousCowOfferedId !== peerMetadata.cowOfferedForTrade?.id
```
lstebner, line 29:
```suggestion
    previousCowOfferedId !== newCowOfferedId
```

also, is the ID ever 0? if so this check will fail when it is. does it even need the first part or can you just check if the two id's are mismatch?

src/game-logic/reducers/updatePeer.js
```diff
@@ -1,18 +1,49 @@
...
+export const updatePeer = (state, farmhand, peerMetadata, peerId) => {
   const peers = { ...state.peers }
-  peers[peerId] = peerState
+
+  const previousPeerMetadata = peers[peerId]
+
+  const previousCowOfferedId = previousPeerMetadata?.cowOfferedForTrade?.id
+  const newCowOfferedId = peerMetadata.cowOfferedForTrade?.id
+
+  const isCowNewlyBeingOfferedForTrade =
```
lstebner, line 27: this variable is a bit of a mouthful so i wanted to suggest something simpler like `isNewTrade` instead. not a blocking comment by any means though

lstebner (commented):
minor comments but i did a quick test and it worked so i'm going to go ahead and approve 🎉 . i'm more than happy to take another look though if you make any changes!

lstebner (approved):

src/game-logic/reducers/updatePeer.js
jeremyckahn: Cow `id`s are strings (UUIDs, specifically), so a valid ID should never be falsy.

> does it even need the first part or can you just check if the two id's are mismatch?

Yes, the `newCowOfferedId &&` check prevents a notification from being displayed when an offering player rescinds their cow without a new one being offered in its place.

I realize now that the intent here isn't super obvious, so I'll add some test coverage for it!

src/game-logic/reducers/updatePeer.js
jeremyckahn: I like it! Done in https://github.com/jeremyckahn/farmhand/commit/d4b00419078efd8fb6f17e2208bd511abf07a795.

src/game-logic/reducers/updatePeer.js
jeremyckahn: Tests added in https://github.com/jeremyckahn/farmhand/commit/f6ebe36b0b7d7d4610f5030558a00be080e87837. I also made some test stub and typing improvements in this commit.

jeremyckahn (commented):
I made some improvements based on the initial review. Thanks for the helpful comments, @lstebner!

src/game-logic/reducers/updatePeer.js
lstebner: ah good to know about the id. it's just semantics but it might be nice to name the property "uid" whenever this is the case instead of just "id" because i strongly associate "id" to integer values, but i'm sure you have a typedef that also could have told me this 😄. 

any thoughts on my suggestion that i started this comment thread with? it seems like you're doing the same check twice when you could just compare your previous to next variables

jeremyckahn: > ah good to know about the id. it's just semantics but it might be nice to name the property "uid" whenever this is the case instead of just "id" because i strongly associate "id" to integer values, but i'm sure you have a typedef that also could have told me this 😄.

I think it's a good idea, but it would involve having to update users' save data (because these IDs are persisted there). We have infrastructure to do that, but I think it's worth avoiding unless absolutely necessary because it adds a lot of risk to deployments. 😕 I think in this case we're better off leaving it for now.

> any thoughts on my suggestion that i started this comment thread with? it seems like you're doing the same check twice when you could just compare your previous to next variables

Whoops I totally overlooked that, sorry! 😅 This is fixed in https://github.com/jeremyckahn/farmhand/commit/cb34b1d1121e30a73193379ce20bcd665eb61e29. Thanks for re-raising that!

lstebner (approved):
looks great! thanks for making the suggested updates

jeremyckahn:
Thanks for reviewing and approving @lstebner! Let's give @jmmullins some time to test this out and then I'll merge and release. 🙂

jeremyckahn:
@jmmullins has QA-approved this [on Discord](https://discord.com/channels/714539345050075176/847884323330850847/1144099688463728761). Merging and release now!

Issue #371, opened 2023-01-10
idea: change leveling to an experience based system


lstebner:
[I shared this idea in discord](https://discord.com/channels/714539345050075176/722286738344640552/1062057469007888445) but wanted to move to an issue for more discussion and potentially building it out.

I would like to explore changing leveling in the game from simply number of items sold to an experience based system. The main benefit would be giving the player experience for performing a variety of activities rather than just selling goods. A secondary benefit would be to help players level up more quickly and get to see more of the games content, which in turn benefits developers so they can build more and more high level content. It seems like right now only power users get to our level 30+ content and we have lots of ideas for things we can add still.

I'm not sure if it's well balanced, but it could go something like this:
- gain 1 experience point for selling an item
- gain 1 experience point per item in a recipe for crafting an item
- gain 1 experience point per ore used in recipe for upgrading a tool
- gain 1 experience for forging items
- gain 1 experience point for recycling recipes
- gain 3 experience for trading a cow
- gain ? experience for expanding field
- gain ? experience for expanding cow pen
- gain 1 experience point per item in request when completing a quest

For legacy players... we could maintain the option of sticking with the # of sales based leveling system. This could be a toggle in the game settings somewhere. Once this experience based system is built we could prompt players with a choice to switch to the new system and explain the difference. If they choose to switch, they will stay the same level they are and just begin earning experience from that point forward. New games could all start with the experience based system.

I would love to hear what others think about a change like this!

jeremyckahn:
This is a great idea! I like what you've got laid out here and I think it would fit well, especially if we can keep it disabled for legacy saves.

The full concept seems like it would be a pretty large effort to implement. What do you think about a scaled-down V1 that we can ship early and iterate on? I'm thinking that we could keep the focus on crafting and forging, and possibly layer on the other experience-giving events in further iterations. But I'm open to other ideas!

lstebner:
MVP is a good idea @jeremyckahn . I thought about it a bit and here's what I'm thinking...

milestone 1: add "experience" stat that works exactly like items sold (1 experience point per item sold), and option to stick to legacy system. this milestone would include persisting the data and any reducers we need for the new data.

milestone 2: add experience to recipe crafting. this step would involve figuring out how much experience recipes should give and hooking that up.

milestone 3: add experience to cow related activities

milestone 4: add experience to mining/forging related stuff

milestone 5: add experience for recycling recipes

[...]

PR #448, opened 2023-08-27
add experience part 1

lstebner:
### What this PR does

The goal of this PR is to lay some groundwork for an experience based level system as outlined in #371 

- add 'experience' to farmhand state
- add 'useLegacyLevelingSystem' to farmhand state and settings screen (only shown when the EXPERIENCE feature is enabled)
- update `levelAchieved` to be more flexible about how level is calculated based on which leveling system is being used
- update all calls to `levelAchieved`
- update utils unit tests for levelAchieved to test both systems
- add 'EXPERIENCE' feature flag
- add reducer for adding experience + unit test
- some utils were moved into their own files to make them easier to work with (mostly to avoid circular dependencies when one util relies on another util) and imports across the codebase were updated
- initialize `experience` value to match number of items sold when loading a saved game
- add 1 experience point per farm item sold (crops, milks, and crafted or forged items)
- items that are not farm items should not reward the player with any experience (seeds, weeds, other supplies)
- importing a save without experience should calculate experience based on items sold

This PR would be easiest to test if checked out and ran locally so that the feature flag can be flipped on and off

With the feature flag enabled:
- [x] there should be a setting in the settings window to toggle between the legacy system and the new system
- [x] the use legacy setting should be persist

With the flag disabled:
- [x] the settings menu should _not_ show the toggle for changing which leveling system is used

Regardless of flag enabled or disabled, these things should happen:
- [x] an `experience` value can be seen when inspecting the Farmhand root component
- [x] the `experience` value in state should match the total number of items sold when loading a saved game that previously did not have experience
- [x] `experience` should increment by 1 for each item sold

### Additional information

image


src/components/SettingsView/SettingsView.js
```diff
@@ -99,6 +100,17 @@ const SettingsView = ({
             }
             label="Display custom names for cows received from other players"
           />
+          <FormControlLabel
+            control={
+              <Switch
+                color="primary"
+                checked={useLegacyLevelingSystem}
+                onChange={() => console.log('todo: save change')}
```
jeremyckahn, line 108: Looks like this is needed before this PR can be promoted from Draft state.

src/components/SettingsView/SettingsView.js
```diff
@@ -99,6 +100,17 @@ const SettingsView = ({
             }
             label="Display custom names for cows received from other players"
           />
+          <FormControlLabel
```
jeremyckahn, line 103: Great idea to make this available to users!

src/components/Farmhand/Farmhand.js
```diff
@@ -497,6 +497,7 @@ export default class Farmhand extends FarmhandReducers {
         [toolType.WATERING_CAN]: toolLevel.DEFAULT,
       },
       useAlternateEndDayButtonPosition: false,
+      useLegacyLevelingSystem: true,
```
jeremyckahn, line 500: It seems like this should be added to the `PERSISTED_STATE_KEYS` array in `constants.js`.

src/components/SettingsView/SettingsView.js
```diff
@@ -99,6 +100,17 @@ const SettingsView = ({
...
+            control={
+              <Switch
+                color="primary"
+                checked={useLegacyLevelingSystem}
+                onChange={() => console.log('todo: save change')}
+                name="use-legacy-leveling-system"
+              />
+            }
+            label="Use legacy leveling system (items sold)"
```
jeremyckahn, line 112: What do you think about a small copy change to make this a bit clearer to unfamiliar players?

```suggestion
            label="Use legacy leveling system (experience is only gained by selling items)"
```

src/game-logic/reducers/addExperience.js
jeremyckahn, line 1: It would be good to get some basic unit test coverage for this reducer to keep test metrics from regressing.

jeremyckahn (commented):
I'm still working through a fuller review, I just wanted to make my in-progress comments available.

src/utils/index.js
```diff
@@ -1202,3 +1164,7 @@ export const isOctober = () => new Date().getMonth() === 9
 export const isDecember = () => new Date().getMonth() === 11
 
 export { default as totalIngredientsInRecipe } from './totalIngredientsInRecipe'
+export { farmProductsSold } from './farmProductsSold'
+export { isItemAFarmProduct } from './isItemAFarmProduct'
+export { isItemAGrownCrop } from './isItemAGrownCrop'
+export { levelAchieved } from './levelAchieved'
```
jeremyckahn, line 1170: If it's not too much of a pain, it would be good to keep the code base consistent by removing these exports and updating the downstream `import`s to reference these functions directly. For example: https://github.com/jeremyckahn/farmhand/blob/d96a158ee1d599eaeae296be36962eecd1e6e375/src/data/achievements.js#L12

src/utils/index.test.js
```diff
@@ -768,12 +768,41 @@ describe('farmProductsSold', () => {
...
+  const cases = [
+    [1, 0],
+    [2, 100],
+    [2, 150],
+    [3, 400],
+    [100, 980100],
+  ]
+
+  describe('with legacy system', () => {
+    test.each(cases)(
```
jeremyckahn, line 780: This is a much better approach! :clap:

jeremyckahn (changes requested):
Just a few more notes. I've gotten through all the files now. Great work here so far!

I haven't tested this out yet, but I will do so once this is promoted to "Ready for review."

src/utils/index.js
lstebner: oh, sure! i'm used to still relying on an index file for imports that are outside of the current directory, but i can make that change

src/game-logic/reducers/addExperience.test.js
```diff
@@ -0,0 +1,20 @@
...
+  it('adds experience to current experience', () => {
+    const newState = addExperience(gameState, 100)
+
+    expect(newState.experience).toEqual(100)
+  })
+
+  it('will check for leveling up when experience is added', () => {
+    addExperience(gameState, 100)
+
+    expect(processLevelUp).toHaveBeenCalled()
```
jeremyckahn, line 18: Ideally this unit test would not be concerned with implementation details like we're doing here. Could this test be changed to examine `gameState` to detect whether the level-up was processed instead?

src/utils/index.js
```diff
@@ -1162,9 +1162,3 @@ export const isOctober = () => new Date().getMonth() === 9
  * @returns {boolean}
  */
 export const isDecember = () => new Date().getMonth() === 11
-
-export { default as totalIngredientsInRecipe } from './totalIngredientsInRecipe'
-export { farmProductsSold } from './farmProductsSold'
-export { isItemAFarmProduct } from './isItemAFarmProduct'
-export { isItemAGrownCrop } from './isItemAGrownCrop'
-export { levelAchieved } from './levelAchieved'
```
jeremyckahn: Thanks for making this change and getting all of the `import`s straightened out!

src/utils/totalIngredientsInRecipe.js
```diff
@@ -1,4 +1,4 @@
-export default function totalIngredientsInRecipe(recipe, amount = 1) {
```
jeremyckahn, line 1: Oh good call to update this. Thanks for switching to a named export!

src/utils/index.js
jeremyckahn: I think there's a worthwhile discussion to be had about which form is more worth using. On the one hand, direct imports make for fewer layers of indirection and are thus a little easier to follow. On the other hand, index files reduce boilerplate.

For scope of this PR we're better off staying consistent, but I'd be open to revisiting which form this project uses in a dedicated PR. 🙂

jeremyckahn (changes requested):
Thanks for following up with my previous review items! I found a few more things to discuss with the latest revision. Crucially, this is currently breaking new game sessions (I've called out where in the `Farmhand` component) so we'll need to get that addressed by adding an `initialValue` to the `reduce`.

We're getting there!

src/components/Farmhand/Farmhand.js
```diff
@@ -590,17 +590,29 @@ export default class Farmhand extends FarmhandReducers {
...
-          // Defer these notifications so that notistack doesn't swallow all
-          // but the last one.
-          setTimeout(() => this.showNotification(message, severity), 0)
+      // saved games will likely have itemsSold but not experience since it's a new
+      // feature so this will "sync" them up
+      let { experience } = state
+      if (experience === 0) {
+        experience = Object.values(state.itemsSold).reduce(
+          (acc, value) => acc + value
+        )
```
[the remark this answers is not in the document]
lstebner: ah yeah, i see. it doesn't break on a _brand new_ game, but when reloading up until a sale has been made i do see the bug. good catch!

src/components/SettingsView/SettingsView.js
```diff
@@ -21,6 +21,7 @@ import './SettingsView.sass'
 
 const SettingsView = ({
   allowCustomPeerCowNames,
+  features = {},
```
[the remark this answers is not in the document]
lstebner: oops, totally forgot about propTypes here. another good catch 😄

src/game-logic/reducers/addExperience.test.js
lstebner: i have mixed feelings about this. not strong enough that I won't change it, but I do want to offer my perspective. i did things this way because I see this test as a unit test, and if it were to check that a level changed it would have to know about game logic conditions that cause the level to change (i.e. how much experience it takes to get from one level to another). the way i see, `addExperience` is only concerned with making sure that `processLevelUp` is invoked, which is what mocking it and checking for it to be called does. i think the place for doing a test on game logic belongs elsewhere, which is what might be missing from this PR, but i haven't found that spec yet 😄 .

jeremyckahn: > the way i see, `addExperience` is only concerned with making sure that `processLevelUp` is invoked

I think this is where our views diverge. Based on the semantics and API of this reducer, I would assume that `addExperience` only cares about adding experience. The fact that level-ups are processed within this reducer seems like an implementation detail that could conceivably be handled outside of this function (though I'm not suggesting we do that).

We definitely do want to test that adding experience causes a level up, but perhaps this isn't the appropriate place to validate that? Specifically, I'm thinking that it may be better to validate that gaining experience results in a level-up in a higher-level React Test Library test. I'm thinking something similar to this: https://github.com/jeremyckahn/farmhand/blob/f791e83c4d4994e30efbdaddead4c446118f598c/src/game-logic/item-sell.test.js#L9-L29

I realize that this is expanding the scope of what you set out to do here, and the current state isn't something I'd block this PR over or anything. It's just a difference of opinion. If you feel good about what we have here, then I'm okay with sticking with it. Just let me know which way you'd like to go. 🙂

jeremyckahn (commented):
The code looks good to merge, but I still haven't had a chance to QA this (sorry about that!). I will approve once I'm able to run through everything!

src/game-logic/reducers/addExperience.test.js
lstebner: i actually think i'm with you on processing level ups as being the thing that is out of place here. so i think i'll remove that. it definitely feels like an unexpected side effect not that you've mentioned it

lstebner: ok i've removed it. i see that `sellItem` is what actually does `processLevelUp` but I don't see any tests in there explicitly checking for level up. i always forget that `level` isn't part of farmhand state and that it's calculated when needed so testing that a level up was achieved is trickier than i expected (at least, without just mocking and checking for the call.) if you have suggestions on how to test it let me know and i am happy to add it to this PR

jeremyckahn: Thanks for making that change!

> testing that a level up was achieved is trickier than i expected (at least, without just mocking and checking for the call.)

If we're speaking in terms of React Testing Library tests, I'd approach this by checking to see if the level up notification text is present on the page: https://github.com/jeremyckahn/farmhand/blob/d37d41073867da6a8dbfa65048c0b008b303a9f5/src/templates.js#L184-L214

lstebner: ahh yes that makes sense. the `sellItem` test doesn't seem like the right place in that case (it isn't doing any rendering). i suppose we could check the notifications in state from there though?

jeremyckahn (changes requested):
I did a bit of testing. Here's what I found:

> the `experience` value in state should match the total number of items sold when loading a saved game that previously did not have experience

I imported one of my high-level save files and saw that `experience` was `0`. I saved the game and refreshed the page, and `experience` was appropriately updated. This wouldn't introduce instability if this PR was merged as-is, but I imagine that this would be perceived as a pretty significant bug once the Experience feature is complete and shipped. It seems like this issue should be addressed in this function: https://github.com/jeremyckahn/farmhand/blob/89bb2ee6e8158ddaa29afda3b234bd022bd7d163/src/utils/index.js#L941-L952

> `experience` should increment by 1 for each item sold

This is working as expected. I noticed that `experience` is also incremented when Weeds are sold. Is this intentional? I don't think we discussed this specifically, but it seems like that shouldn't be the case because Weeds do not currently count towards experience. What do you think?

Aside from this, things are looking good. Great work so far!

lstebner:
thanks for calling those things out @jeremyckahn . i've made some updates now to handle importing saves and making sure that experience is only rewarded for farm products sold.

src/utils/index.test.js
```diff
@@ -1046,3 +1047,52 @@ describe('getCowImage', () => {
     expect(image).toEqual(animals.cow.rainbow)
   })
 })
+
+describe('transformStateDataForImport', () => {
```
jeremyckahn, line 1051: Thank you for setting up test cases for this function! :pray:

src/utils/isItemAFarmProduct.js
```diff
@@ -2,13 +2,11 @@ import { itemType } from '../enums'
...
  * @param {farmhand.item} item
  * @returns {boolean}
  */
 export const isItemAFarmProduct = item =>
-  Boolean(
-    isItemAGrownCrop(item) ||
-      item.type === itemType.MILK ||
-      item.type === itemType.CRAFTED_ITEM
-  )
+  Boolean(isItemAGrownCrop(item) || FARM_PRODUCT_TYPES.includes(item.type))
```
jeremyckahn, line 12: This is an excellent refactor! :100:

jeremyckahn (approved):
This looks ready to merge! Thanks for sticking with this and addressing all the issues. Awesome work! :rocket:

PR #449, opened 2023-08-28
Lstebner/experience pt2

lstebner:
### What this PR does

this is part 2 of the experience update outlined in #371 

- update the level progress indicator to be able to reference experience instead of items sold when not using the legacy system. to make this easier to do, this part of the Navigation component was broken out into its own file
- added some constants to use for various experience values. these are just some initial ideas and definitely not set in stone
- add achievement for paying off loan and getting an experience bonus for it
- removing all references to the EXPERIENCE feature and legacy leveling system after some back and forth discussion in the comments here
- updating all calls to `levelAchieved` to pass in experience
- adding `processLevelUp` to `addExperience` so level ups are properly processed any time experience is modified
- removing `farmProductSalesNeededForLevel` and replacing with `experienceNeededForLevel` along with any wording referencing "sales needed" to mention "experience points" instead

- [x] start a new game with the experience feature enabled and check that the legacy leveling system toggle is "off" (in settings)
- [x] load an existing save and check that the legacy leveling system toggle is "on"
- [x] toggle the legacy leveling system on and off and check that the level indicator correctly displays "sales" or "experience" text
- [x] pay off the initial debt and check that you are rewarded with an achievement and also experience

the easiest way to monitor an experience change is to either have the legacy leveling system off and just watch the level indicator progress bar, or have the component inspector open and watch the "experience" state value go up 📈 

n/a

### Additional information

part 3 of this will be to hook up the rest of the experience rewards and do a QA pass on all of it. after that we can go live with this change!

when the legacy system is disabled the tooltip will indicate experience is required instead of sales:

image

achievement for debt-free:

image


jeremyckahn (commented):
Great work here! I haven't had a chance to test this yet, but I was at least able to do a preliminary code review. I'll test this out the next chance I get and approve once this is all validated.

src/components/Navigation/DayAndProgressContainer.js
```diff
@@ -0,0 +1,105 @@
...
+      experience,
+      experienceNeededForLevel(currentLevel),
+      experienceNeededForLevel(currentLevel + 1),
+      0,
+      100
+    )
+
+    tooltipText = `${integerString(
+      experienceNeededForLevel(currentLevel + 1) - experience
+    )} more experience needed to reach level ${integerString(currentLevel + 1)}`
```
jeremyckahn, line 60: What do you think about this tweaked wording?

```suggestion
    )} more experience points needed to reach level ${integerString(currentLevel + 1)}`
```

src/data/achievements.js
```diff
@@ -321,6 +326,15 @@ const achievements = [
       return addItemToInventory(state, itemsMap['gold-ingot'], 1, true)
     },
   }))(),
+
+  (() => ({
```
jeremyckahn, line 330: The achievements are presented to the player in the order they are implemented in this array. Since paying off the loan is typically an early game task, let's move this to the fourth position in this array (under `harvest-crop`).

jeremyckahn, line 330: I just realized that this is going to be shown to players when the Legacy leveling system is enabled. :grimacing:

This complicates things, because there's no great way to handle conditional achievements. The good news is that this doesn't actually break anything in the game because toggling legacy mode appropriately updates the level state. It's just a bit of an odd UX.

I see a few ways forward here:

1. Accept this bit of UX weirdness and do nothing about it. (Easiest solution)
2. Implement support for conditional achievements. (Hardest solution)
3. Eliminate support for the legacy leveling system and bring all players to the new system with no way back.

Of these three options, number 3 seems the most appealing to me. It's the most consequential solution because it increases the scope of this PR and it has the most significant impact on players. However, this approach would allow us to avoid the complexity of maintaining two interrelated systems and the potential bug vectors that would invite.

@lstebner what do you think about how to address this?

src/data/achievements.js
```diff
@@ -321,6 +326,15 @@ const achievements = [
...
+    id: 'debt-free',
+    name: 'Debt Free',
+    description: 'Pay off your initial loan from the bank.',
+    rewardDescription: 'Experience bonus',
+    condition: state => state.loanBalance === 0,
+    reward: state => addExperience(state, EXPERIENCE_VALUES.LOAN_PAID_OFF),
```
jeremyckahn, line 336: ‼️ I found a significant bug with this: Adding experience in this way does not process a level up, so the player does not receive their reward. It seems that `addExperience` should be updated to account for this so that the level-up processing logic is abstracted away. We [discussed this in the previous Experience PR](https://github.com/jeremyckahn/farmhand/pull/448/files/d96a158ee1d599eaeae296be36962eecd1e6e375..fd88b395cd0154059c2dfa2156ea5caa397593ac#r1307529634), but it looks like we'll need to revisit our resolution to that.

I recommend revisiting the implementation of `addExperience` and updating it to process the level-up. We can further discuss how to appropriately test that, but let's settle on a design decision for that first.

jeremyckahn (changes requested):
I did some testing and things are mostly working, but I found a bug that will need to be resolved before this can be merged. I think this PR is also a great opportunity to get ahead of some complexity that I've described in a comment.

src/data/achievements.js
lstebner: the achievement _does_ have access to state, so a potential option 4 is to give a different reward based on whether the legacy system is being used or not. i.e. the reward could be 25 crops, instead of 25 experience, which ultimately amounts to the same thing.

but if you just want to go all in on experience then 3 is the way to go. i think by naming the old system "legacy" we are essentially deprecating it for future removal anyways. the main goal with leaving the old system was just to not disrupt anyone's game that is currently in progress.

im open to whatever your preference is here! i don't mind the change in scope as it's basically just removing the legacy toggle and any checks for the boolean

src/data/achievements.js
lstebner: ahhh goood find. yeah it does feel like `addExperience` should also process level up so that makes sense to me. there are going to be lots of other places where we add experience that are going to need `processLevelUp` called at the same time so it will be easiest if it just does it for us. i'll add that back in

src/data/achievements.js
jeremyckahn: The fourth option you proposed is a great option that I hadn't considered! I think that would solve the problem well.

> but if you just want to go all in on experience then 3 is the way to go. i think by naming the old system "legacy" we are essentially deprecating it for future removal anyways. the main goal with leaving the old system was just to not disrupt anyone's game that is currently in progress.

Yep that was the idea. Thinking about this more, I see the potential for an exploit if we update `addExperience` to process level-ups (which still seems to be right way forward): If the player has, say, one sale below a level threshold and a greater amount of experience than items sold, they could toggle between experience systems to repeatedly get the level up reward. That's a solvable problem, but at that point the system gets so complex that it calls into question if it's worth the trouble to maintain both modes. I'm thinking that it's not.

I say let's be bold and remove the old leveling system and avoid all that extra complexity. We can do that in this PR or a later one, but either way let's plan to do that as a next step in building #371. 🙂

src/data/achievements.js
```diff
@@ -321,6 +326,15 @@ const achievements = [
       return addItemToInventory(state, itemsMap['gold-ingot'], 1, true)
     },
   }))(),
+
+  (() => ({
+    id: 'debt-free',
+    name: 'Debt Free',
```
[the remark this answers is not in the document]
lstebner: love it!

src/data/achievements.js
lstebner: added it back in and wrote a test for it based on notifications. open to feedback though of course!

src/game-logic/reducers/sellKeg.js
jeremyckahn, line 1: ‼️ I'm noticing that selling kegs no longer results in gaining experience, so we'll need to address that before this is merged. It seems like we'll just need to call `addExperience` in this file.

jeremyckahn (changes requested):
Great progress here! I noticed that selling kegs no longer gives experience, so we should fix that regression before merging. I also noticed a few prop type errors being introduced via some outdated tests, so it's worth addressing those as well.

src/game-logic/reducers/addExperience.test.js
```diff
@@ -1,11 +1,32 @@
...
+  it('process a level up when enough experience is achieved', () => {
+    const newState = addExperience(gameState, experienceNeededForLevel(2))
 
-    const newState = addExperience(gameState, 100)
+    const notification = /You reached \*\*level 2!\*\*/
 
-    expect(newState.experience).toEqual(100)
+    expect(notification.test(newState.todaysNotifications[0].message)).toEqual(
+      true
+    )
```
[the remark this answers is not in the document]
lstebner: nice one! never heard of this matcher before

src/components/OnlinePeersView/OnlinePeer/OnlinePeer.js
```diff
@@ -46,8 +46,8 @@ OnlinePeer.propTypes = {
   peer: shape({
     cowOfferedForTrade: object,
     dayCount: number.isRequired,
+    experience: number.isRequired,
```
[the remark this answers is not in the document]
lstebner: oops, thanks for catching this!

lstebner:
alright another round of updates is in @jeremyckahn . i think we're getting there! i've been doing some testing as well and things are seeming pretty stable

src/components/OnlinePeersView/OnlinePeer/OnlinePeer.js
```diff
@@ -24,7 +24,7 @@ const OnlinePeer = ({
             subheader: (
               <div>
                 <p>Day: {integerString(dayCount)}</p>
-                <p>Level: {integerString(levelAchieved({ experience }))}</p>
+                <p>Level: {integerString(levelAchieved(experience))}</p>
```
jeremyckahn, line 27: We'll need to ensure that `experience` has a fallback value because this data comes from online peers. When this PR is released, it will take some time for preexisting installs to update to use the new experience system.

```suggestion
                {/* TODO: Remove `?? 0` after 10/24 */}
                <p>Level: {integerString(levelAchieved(experience ?? 0))}</p>
```

jeremyckahn (commented):
Excellent work! I noticed one minor bug, but I can fix that myself. I'll approve this PR once I play test it again.

jeremyckahn (changes requested):
I noticed a bug with online play: Peers' levels are no longer being reported correctly.

Online peer bug

jeremyckahn:
> I noticed a bug with online play: Peers' levels are no longer being reported correctly.

This is fixed in https://github.com/jeremyckahn/farmhand/commit/fdab492ace02ba2e2a9550549d6196b66330ba5a.

jeremyckahn (approved):
Brilliant work @lstebner! I think this new experience system is much better and opens up the possibility of many more interesting incentives throughout the game. Feel free to merge and release as a `minor` version (unless @jmmullins is available to QA first).

LGTM!

PR #451, opened 2023-09-09
[closes #371] add experience to a variety of activities

### LINKED ISSUE #371: idea: change leveling to an experience based system

lstebner:
### What this PR does

this is meant to be the final part of #371 which adds experience for a ton of the in game activities including:

- [x] composter acquired
- [x] cellar purchased
- [x] cellar expanded
- [x] cow bred
- [x] rainbow cow bred
- [x] cow pen acquired
- [x] cow pen expanded
- [x] field expanded
- [x] kitchen recipe made
- [x] forge recipe made
- [x] fermentation recipe made
- [x] recycling recipe made
- [x] smelter acuired
- [x] on the 1st day of every "year"
- [ ] successful cow trade

test all of the above listed activities and check that experience is gained when performing them

i wasn't sure how to hook up to a cow being traded. the code around that was kind of confusing, but i'm ok with omitting it based on how many other things now give experience. but maybe you know how to do it offhand and can show me @jeremyckahn


src/game-logic/reducers/makeRecipe.test.js
```diff
@@ -37,4 +39,34 @@ describe('makeRecipe', () => {
...
+    beforeEach(() => {
+      state = {
+        experience: 0,
+        inventory: [],
+      }
+
+      jest.mock('../../utils', () => ({
+        ...jest.requireActual('../../utils'),
+        canMakeRecipe: jest.fn(() => true),
+      }))
```
jeremyckahn, line 55: I don't think this mock is actually doing anything because it's being defined after `makeRecipe` is loaded. When I remove this mock, the tests still pass.

In order for this mock to actually be used, I think we'd need to reload `makeRecipe` with [`jest.requireActual`](https://jestjs.io/docs/jest-object#jestrequireactualmodulename). Here's an example of where that's done: https://github.com/jeremyckahn/farmhand/blob/9a8c5567f932e3acbf28342ec1defd58aa7b2dc2/src/utils/index.test.js#L801-L821

jeremyckahn (changes requested):
Great work so far, @lstebner! The code is looking pretty good overall and I just noticed a few minor issues worth addressing. I haven't been able to QA this yet and that will take me some time given the scope of this change. I'll approve once I've successfully run through everything.

Thanks for making this huge improvement to the game!

src/game-logic/reducers/computeStateForNextDay.js
```diff
@@ -58,11 +60,19 @@ export const computeStateForNextDay = (state, isFirstDay = false) =>
...
+  let newState = reducers
     .concat([adjustItemValues])
     .reduce((acc, fn) => fn({ ...acc }), {
       ...state,
       cowForSale: generateCow(),
       dayCount: state.dayCount + 1,
       todaysNotifications: [],
     })
+
+  if (newState.dayCount > 1 && newState.dayCount % 365 === 1) {
```
jeremyckahn, line 73: I noticed during playtesting that I was awarded experience once I reached day 366 rather than 365. To me it makes more sense to give experience on day multiples of 365, and I think that would look something like this:

```suggestion
  if (newState.dayCount % 365 === 0) {
```

What do you think @lstebner?

jeremyckahn (changes requested):
I just ran through all of the experience scenarios, and they all seem to work great! Awesome work @lstebner. This looks good from a QA perspective to me.

Here's my QA save file: https://www.dropbox.com/scl/fi/kmf2klaib8r69uf4fsmdn/farmhand-2023-09-09.json?rlkey=lj9bfkvi255mh95xaadsdg5il&dl=0

> i wasn't sure how to hook up to a cow being traded. the code around that was kind of confusing, but i'm ok with omitting it based on how many other things now give experience. but maybe you know how to do it offhand and can show me @jeremyckahn

I'm fine with deferring that to a later PR, or adding it to this one. Either way, I think the place to implement that would be in `handleCowTradeRequestAccept`:

https://github.com/jeremyckahn/farmhand/blob/9a8c5567f932e3acbf28342ec1defd58aa7b2dc2/src/handlers/peer-events.js#L155-L241

-------------------

There's still some changes outstanding and this is still in Draft state, so I'll mark this review as "Request changes" for now.

src/game-logic/reducers/computeStateForNextDay.js
```diff
@@ -36,8 +38,8 @@ const adjustItemValues = state => ({
  * @param {farmhand.state} state
  * @returns {farmhand.state}
  */
-export const computeStateForNextDay = (state, isFirstDay = false) =>
-  (isFirstDay
```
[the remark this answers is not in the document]
lstebner: haha it's all good!

src/game-logic/reducers/makeRecipe.test.js
lstebner: good call. i wasn't thinking clearly when i did this haha

src/game-logic/reducers/computeStateForNextDay.js
lstebner: so, i went back and forth on this for awhile to be honest, i'm open to either but what I was thinking was the 365th day is the last day of the year, and this check happens _after_ dayCount is updated, so i _think_ day 366 is the 1st day of the year meaning this runs at "midnight" of the last day. i definitely spent too long going back and forth on this haha. open to whatever you prefer here!

lstebner:
thanks @jeremyckahn ! i made the changes you suggested and also added experience for the cow trade scenario since you pointed me to the place. i figured it's easier to just do it than open an issue and another PR.

src/game-logic/reducers/computeStateForNextDay.js
jeremyckahn: > the 365th day is the last day of the year, and this check happens _after_ dayCount is updated, so i _think_ day 366 is the 1st day of the year meaning this runs at "midnight" of the last day.

Logically this makes sense, but to me it _felt_ a little awkward from a UX perspective. Given that this is a game and not something "serious" like productivity software, I think it makes sense to bias towards what _feels_ right rather than what _is_ right for things like this.

src/game-logic/reducers/computeStateForNextDay.js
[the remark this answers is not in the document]
jeremyckahn: I'm glad you're not too bothered by it! 😅

I was developing Farmhand solo for a long time at first, and I assumed that would always be the case (though I'm happy that I was wrong!). I wanted to try out a few coding ideas because of that (many of which are still present on the code base today). Some worked well, but some didn't. 😆

jeremyckahn (changes requested):
It looks like the latest commit introduced a build breakage: https://github.com/jeremyckahn/farmhand/actions/runs/6133116793/job/16644692811?pr=451

jeremyckahn:
> It looks like the latest commit introduced a build breakage: https://github.com/jeremyckahn/farmhand/actions/runs/6133116793/job/16644692811?pr=451

This is fixed in https://github.com/jeremyckahn/farmhand/commit/5b26643e6f8da50d95013c8da0ef6f9ca505fca8.

src/handlers/peer-events.js
```diff
@@ -207,6 +211,10 @@ export const handleCowTradeRequestAccept = (farmhand, cowReceived, peerId) => {
         'success'
       )
 
+      if (!didOriginallyOwnReceivedCow) {
```
jeremyckahn, line 214: I added this to prevent the player from receiving experience by re-trading their own cows.

lstebner: good call!

src/game-logic/reducers/computeStateForNextDay.js
lstebner: agree with what you're saying here, made the change!

jeremyckahn (approved):
Thanks for making all the changes! This looks great. I've play tested all of the experience-giving events and everything seems to be working well. Once you're ready, feel free to merge and make a `patch` release.

LGTM! 😁

PR #464, opened 2023-10-30
feat(chat): [closes #453] Add chat

### LINKED ISSUE #453: add a chat
just so there's actual interaction and people can be friends and other things instead of just knowing people might be there

jeremyckahn:
### What this PR does

For #453, this PR adds chat functionality. Chat is only available when a player is online and in a non-`global` room.

1. Open a browser and join the `global` room. Observe that there is a placeholder for the chat button under the "Active players" button.
2. Open two browsers and join a room other than `global`. Open the chat modal and validate that the two browsers connect and can chat with each other.

### Additional information

[Screencast of chat feature](https://github.com/jeremyckahn/farmhand/assets/366330/013bb815-bf1e-4bcc-b35d-a01941fea688)


lstebner (approved):
i did not do an extensive amount of QA on this but the code all looks good to me, and nothing seems broken so i think it's good to go! but if you want me to do more QA, or if you want to sync up and do some testing with us both online let me know and i'd be happy to do so

PR #471, opened 2023-12-20
fix(settings): Prevent invalid state when booting save with Home screen disabled

jeremyckahn:
### What this PR does

This PR fixes two issues that I discovered during development:

1. The "Show the Home Screen" setting wasn't working properly. The setting's value was persisting, but the app bootup sequence didn't properly account for it. The end result is that the app booted up in an invalid state and broke some views.
2. There was a spurious warning being logged if the player cycled through views too quickly after booting. This PR improves the asynchronous handling of the dynamic cow image generation to prevent this warning.

1. Open the Settings dialog, disable "Show the Home Screen," save the game, and reboot. Verify that you can cycle through all of the views without seeing any errors.
2. Cycle through the Cows view as quickly as you can and verify that there are no errors or warnings in the browser console.

### DIFF
--- a/src/components/CowPen/CowPen.js
+++ b/src/components/CowPen/CowPen.js
@@ -34,6 +34,7 @@ export class Cow extends Component {
   repositionTimeoutId = null
   animateHugTimeoutId = null
   tweenable = new Tweenable()
+  isComponentMounted = false
 
   static flipAnimationDuration = 1000
   static transitionAnimationDuration = 3000
@@ -162,15 +163,21 @@ export class Cow extends Component {
   }
 
   componentDidMount() {
+    this.isComponentMounted = true
     this.scheduleMove()
     ;(async () => {
-      this.setState({ cowImage: await getCowImage(this.props.cow) })
+      const cowImage = await getCowImage(this.props.cow)
+
+      if (!this.isComponentMounted) return
+
+      this.setState({ cowImage: cowImage })
     })()
   }
 
   componentWillUnmount() {
     ;[this.repositionTimeoutId, this.animateHugTimeoutId].forEach(clearTimeout)
 
+    this.isComponentMounted = false
     this.tweenable.cancel()
   }
 
--- a/src/hooks/useMountState/useMountState.ts
+++ b/src/hooks/useMountState/useMountState.ts
@@ -1,0 +1,17 @@
+import { useEffect, useRef } from 'react'
+
+export const useMountState = () => {
+  const isMountedRef = useRef(false)
+
+  useEffect(() => {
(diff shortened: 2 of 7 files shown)


src/hooks/useMountState/useMountState.ts
jeremyckahn, line 1: This is a TypeScript file!

#sorryNotSorry :)

src/components/CowPen/CowPen.js
```diff
@@ -162,15 +163,21 @@ export class Cow extends Component {
   }
 
   componentDidMount() {
+    this.isComponentMounted = true
```
lstebner, line 166: is there a reason you didn't put this into state? just curious because my experience with this pattern has always been a state value

lstebner (approved):
one question but i did some testing as per your instructions and everything seems to be working for me

src/components/CowPen/CowPen.js
jeremyckahn: Unfortunately that wouldn't work here. I tried this:

```diff
diff --git a/src/components/CowPen/CowPen.js b/src/components/CowPen/CowPen.js
index 4d5f3173..8d3407ca 100644
--- a/src/components/CowPen/CowPen.js
+++ b/src/components/CowPen/CowPen.js
@@ -29,12 +29,12 @@ export class Cow extends Component {
     showHugAnimation: false,
     x: randomPosition(),
     y: randomPosition(),
+    isComponentMounted: false,
   }
 
   repositionTimeoutId = null
   animateHugTimeoutId = null
   tweenable = new Tweenable()
-  isComponentMounted = false
 
   static flipAnimationDuration = 1000
   static transitionAnimationDuration = 3000
@@ -163,12 +163,12 @@ export class Cow extends Component {
   }
 
   componentDidMount() {
-    this.isComponentMounted = true
+    this.setState({ isComponentMounted: true })
     this.scheduleMove()
     ;(async () => {
       const cowImage = await getCowImage(this.props.cow)
 
-      if (!this.isComponentMounted) return
+      if (!this.state.isComponentMounted) return
 
       this.setState({ cowImage: cowImage })
     })()
@@ -177,7 +177,7 @@ export class Cow extends Component {
   componentWillUnmount() {
     ;[this.repositionTimeoutId, this.animateHugTimeoutId].forEach(clearTimeout)
 
-    this.isComponentMounted = false
+    this.setState({ isComponentMounted: false })
     this.tweenable.cancel()
   }
 ```


But `this.state.isComponentMounted` wasn't set to `false` in time for it to be evaluated as we expect after the `const cowImage = await getCowImage(this.props.cow)` line.

PR #473, opened 2024-01-13
add the ability for crops to have more than 1 growing phase

lstebner:
### What this PR does

 this is a backwards compatible change that allows for the cropTimetable to support a growing phase that is an array of day lengths. each index in this array will be considered a "phase" and getPlotImage will use the phase to look for an image named `<crop>-growing-<phase>` instead of simply `<crop>-growing`.

if the GROWING phase is just a single value then this is considered "phase 0" by the code and this will use the default `<crop>-growing` image name.

this PR also updates the growing phase of every crop to the new array notation, and adds multiple phases to (with new art!) to a majority of them. everywhere that phases were added _should not_ change the total growing time of any crop. the goal was to keep the duration the same, but just add more visual flair to the process.

in this PR every crop was updated to use the array notation for the growing stage, and a majority were updated to have multiple phases visually. all updated crops are shown in the gifs below (however some of the images are out of date with the final state of the PR.)

it's also important to test other crops that were not updated and ensure that they still grow as expected.

### Additional information

[video shared in discord cause github said it's too big](https://discord.com/channels/714539345050075176/847884323330850847/1195883615901458452)

I went a little nuts updating a bunch of crops because I found it fun, so here are GIFS! note* these are slightly out of date but kind of a pain to regenerate so it is what it is 😄 

asparagus-growing
carrot-growing
corn-growing
garlic-growing
jalapeno-growing
onion-growing
potato-growing
pumpkin-growing
sunflower-growing
sweet-potato-growing
tomato-growing


src/img/index.js
```diff
@@ -44,6 +44,9 @@ import asparagusGrowing from './items/asparagus-growing.png'
 import asparagusSeed from './items/asparagus-seed.png'
 import carrot from './items/carrot.png'
 import carrotGrowing from './items/carrot-growing.png'
```
jeremyckahn, line 46: Do we need this old version of `carrotGrowing` anymore? If not, can we remove it?

src/utils/getCropLifecycleDuration.js
```diff
@@ -6,6 +6,12 @@ import { memoize } from './memoize'
...
-)
+export const getCropLifecycleDuration = memoize(({ cropTimetable }) => {
+  return Object.values(cropTimetable).reduce((acc, value) => {
+    if (Array.isArray(value)) {
+      return acc + value.reduce((acc2, value2) => acc2 + value2)
+    }
+
+    return acc + value
+  }, 0)
+})
```
jeremyckahn, line 17: Let's add some test coverage for the additional logic path here: https://github.com/jeremyckahn/farmhand/blob/cfdab66a5347427e16310e8e65b0484b08a08a1b/src/utils/getCropLifecycleDuration.test.js#L7-L11

src/utils/index.js
```diff
@@ -294,15 +294,52 @@ export const doesPlotContainCrop = plot =>
 
 export const getLifeStageRange = memoize((
   /** @type {farmhand.cropTimetable} */ cropTimetable
-) =>
-  [SEED, GROWING].reduce(
+) => {
+  return [SEED, GROWING].reduce(
```
jeremyckahn, line 298: Thanks for making this change. There are a lot of remnants in this code base of me getting overly clever with arrow function expressions. 😅

src/utils/index.js
```diff
@@ -294,15 +294,52 @@ export const doesPlotContainCrop = plot =>
...
-    (acc, stage) => acc.concat(Array(cropTimetable[stage]).fill(stage)),
+    (acc, stage) => {
+      if (Array.isArray(cropTimetable[stage])) {
+        return acc.concat(
+          cropTimetable[stage].reduce(
+            (acc2, value) => acc2.concat(Array(value).fill(stage)),
+            []
+          )
+        )
+      }
```
jeremyckahn, line 310: This new logic needs test coverage as well: https://github.com/jeremyckahn/farmhand/blob/cfdab66a5347427e16310e8e65b0484b08a08a1b/src/utils/index.test.js#L529-L537

src/utils/index.js
```diff
@@ -294,15 +294,52 @@ export const doesPlotContainCrop = plot =>
...
     []
   )
-)
+})
+
+/**
+ * @param {farmhand.crop} crop
+ * @returns {number}
+ */
+export const getGrowingPhase = memoize(crop => {
```
jeremyckahn, line 322: If we're going to memoize this function, let's give it a `cacheSize` of `PURCHASEABLE_FIELD_SIZES.get(3).columns * PURCHASEABLE_FIELD_SIZES.get(3).rows`. Without that `cacheSize`, the memoization cache is liable to be busted before the Field is fully calculated, thus eliminating any performance gains.

Here's an example of using `cacheSize`: https://github.com/jeremyckahn/farmhand/blob/cfdab66a5347427e16310e8e65b0484b08a08a1b/src/utils/index.js#L456-L476

And here's `PURCHASEABLE_FIELD_SIZES`: https://github.com/jeremyckahn/farmhand/blob/cfdab66a5347427e16310e8e65b0484b08a08a1b/src/constants.js#L37-L43

src/data/__mocks__/items.js
```diff
@@ -140,6 +140,19 @@ export const sampleCropItem3 = testItem({
...
+  id: 'sample-crop-4',
+  name: 'Sample Crop Item 4',
+  type: itemType.CROP,
+  value: 2,
+  cropTimetable: {
+    [SEED]: 2,
+    [GROWING]: [1, 2, 1],
+  },
+})
+
```
jeremyckahn, line 155: I'd like us to get away from using the `sampleCropItem`s for tests. This was one of my less-good ideas from early on in the project's development that there are better solutions for. It seems that `carrot` should cover our needs for testing growth phases, so let's use that for the tests introduced in this PR.

jeremyckahn (changes requested):
Nice work @lstebner! I didn't have enough time to give this a super close look or test it this morning, but I wanted to at least share some initial observations. I think this is a great addition to the game!

src/img/index.js
lstebner: good catch, can indeed remove this

src/utils/index.js
lstebner: ah good call, though now i wonder... do you think we want this memoized? i was more or less just copying the other utils nearby here and didn't think much about it. i don't mind making this change if we do though

src/data/__mocks__/items.js
lstebner: will do! i didn't actually realize you wanted to get away from these, so i kinda just followed the existing pattern. i fully support using the actual crops here though

lstebner: i went ahead and updated all the tests using mocked items in this test file

src/utils/getCropLifecycleDuration.js
lstebner: i think now that i'm using a real crop in the test and all crops have the array GROWING phase this is naturally covered

src/utils/index.js
lstebner: updated

src/utils/index.js
jeremyckahn: I think there's value in it because this is a fairly hot code path. That said, there's a case to be made for dropping the memoization, as it does qualify as a premature optimization.

I vote for keeping the memoization (with appropriate cache configuration), but I'm okay with whichever way you'd like to go with this @lstebner.

jeremyckahn (changes requested):
Thanks for making the changes @lstebner! We just need to get a resolution on the memoization issue and then this should be mergeable. 🙂

src/utils/index.js
lstebner: i'll add it, i thought maybe it didn't matter because the results to this method change almost every day, but i didn't analyze how often this might be called.

src/data/crops/corn.js
```diff
@@ -11,7 +13,7 @@ export const cornSeed = crop({
   cropType: cropType.CORN,
   cropTimetable: {
     [SEED]: 3,
-    [GROWING]: 7,
+    [GROWING]: [1, 1, 1, 2, 2],
```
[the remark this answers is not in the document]
lstebner: that's an interesting thought that had not occurred to me. so you're thinking to basically update this to `cropTimeline: [3, 1, 1, 1, 2, 2]` instead of `cropTimetable`, and then the "seed" images get renamed to "growing-1" (or maybe "growing-0" since i found out after i started daysWatered was 0 based when i assumed it was 1 based) since it now knows how to cycle through a series of images, and it doesn't really care what they're called 🤔 

is there anywhere in the code that relies on knowing something is in the seed phase outside of the images? i haven't looked yet but maybe you know offhand. if not this seems like a pretty straightforward change to make.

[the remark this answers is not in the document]
lstebner: well, are you thinking we would remove the concept of "SEED" stage entirely? or would we update `getLifeStageRange` to assume the 0 index of the cropTimeline is the "SEED" stage? 

i only see 1 place that isn't obvious to update if we go for completely remove SEED stage and that is in Plot.js where it sets the label text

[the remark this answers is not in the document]
jeremyckahn: > well, are you thinking we would remove the concept of "SEED" stage entirely?

Yeah exactly. The `cropTimetable` is mostly for determining which image to show, and with this PR there isn't a meaningful or functional distinction between the initial planted "seed" image and the intermediate growth stage images.

src/utils/index.js
```diff
@@ -319,27 +320,32 @@ export const getLifeStageRange = memoize((
...
+      daysGrowing -= value
+      phase += 1
+    }
 
-  return phase
-})
+    return phase
+  },
+  {
+    cacheSize: PURCHASEABLE_FIELD_SIZES.get(3).columns,
```
jeremyckahn, line 346: This needs to account for every plot in the field:

```suggestion
    cacheSize: PURCHASEABLE_FIELD_SIZES.get(3).columns * PURCHASEABLE_FIELD_SIZES.get(3).rows,
```

jeremyckahn (changes requested):

src/utils/index.js
lstebner: oops, you said that in your previous comment as well and i still did it wrong haha

src/data/crops/corn.js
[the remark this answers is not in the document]
jeremyckahn: Oh whoops I just noticed your previous comment. 😅

> update this to `cropTimeline: [3, 1, 1, 1, 2, 2]` instead of `cropTimetable`

Yep!

> and then the "seed" images get renamed to "growing-1"

I don't think we need to rename the `*-seed` image files, as those are used in non-Field  contexts (like the Shop).

> is there anywhere in the code that relies on knowing something is in the seed phase outside of the images?

Nothing that comes to mind. Even if so, such logic could be refactored to work with `cropTimetable` being replaced with `cropTimeline`.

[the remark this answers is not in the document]
lstebner: done! everything seems to be working but please double check and test this 😄

src/game-logic/reducers/processWeather.test.js
```diff
@@ -1,17 +1,18 @@
...
 
 describe('processWeather', () => {
   test('does not water plants when there is no precipitation', () => {
-    jest.resetModules()
-    jest.mock('../../constants', () => ({
-      PRECIPITATION_CHANCE: 0,
-    }))
-
-    const { processWeather } = jest.requireActual('./processWeather')
+    shouldPrecipitateToday.mockReturnValue(false)
```
jeremyckahn, line 15: This is a much better of setting up these tests! :clap:

src/img/index.js
jeremyckahn, line 1: Great idea to update all the names in this file!

src/utils/getCropLifecycleDuration.js
```diff
@@ -1,17 +1,11 @@
 import { memoize } from './memoize'
 
-// TODO: Refactor this to accept just a plain cropTimetable
-// https://github.com/jeremyckahn/farmhand/issues/415
 /**
- * @param {{ cropTimetable: Object }} crop
+ * @param {{ cropTimeline: Array}} crop
```
jeremyckahn, line 4: This type could be a little narrower:

```suggestion
 * @param {{ cropTimeline: number[] }} crop
```

src/utils/index.js
```diff
@@ -2,7 +2,6 @@
 /** @typedef {import("../index").farmhand.item} farmhand.item */
 /** @typedef {import("../index").farmhand.plotContent} farmhand.plotContent */
 /** @typedef {import("../index").farmhand.shoveledPlot} farmhand.shoveledPlot */
-/** @typedef {import("../index").farmhand.cropTimetable} farmhand.cropTimetable */
```
jeremyckahn, line 5: I think we'll want to keep this but change it to:

```ts
/** @typedef {import("../index").farmhand.cropTimeline} farmhand.cropTimeline */
```

src/utils/index.js
```diff
@@ -293,67 +293,60 @@ export const doesPlotContainCrop = plot =>
...
-    (acc, stage) => {
-      if (stage === GROWING) {
-        return acc.concat(
-          cropTimetable[stage].reduce(
-            (acc2, value) => acc2.concat(Array(value).fill(stage)),
-            []
-          )
-        )
-      }
+  let lifeStageRange = Array(cropTimeline.slice(0, 1)[0]).fill(SEED)
```
jeremyckahn, line 298: Could this be simplified to something like the following?

```suggestion
  let lifeStageRange = Array(cropTimeline[0]).fill(SEED)
```

jeremyckahn (approved):
Amazing work @lstebner! I called out a few things that could be tweaked, but I found no blocking issues. Feel free to merge and release. Thanks for sticking with this and giving Farmhand a much-appreciated graphical upgrade!

src/utils/index.js
lstebner: oh good call. i originally removed it because it's not a "special" type any more but i see now where it's being referenced and complaining. added back!

src/utils/index.js
lstebner: oh man why did i overcomplicate this so much haha

src/utils/index.js
jeremyckahn: I actually had this slightly wrong as `farmhand.cropTimeline` is not defined as a type. I've fixed this up in https://github.com/jeremyckahn/farmhand/commit/9a9135d55d35e175567ddd14fc7663ed07a3a4f8.

PR #480, opened 2024-01-30
feat: add total price to Item

tim-phillips:
### What this PR does

This PR adds total price next to item price in the Item card when quantity is greater than one.

Go to the store and view the Seeds tab, click the up arrow on a seed card.

In the inventory on the left sidebar, click the up arrow on a card that has the option of selling more than one.

For sell price, I'm not sure what happens if the text content runs into the edge of the card.

### DIFF
--- a/src/components/Item/Item.js
+++ b/src/components/Item/Item.js
@@ -82,6 +82,24 @@ const SellValueIndicator = ({
   />
 )
 
+const TotalPrice = ({ quantity, pricePerUnit }) => {
+  if (!quantity || quantity <= 1) return null
+
+  return (
+    <span>
+      {' '}
+      (Total:{' '}
+      <AnimatedNumber
+        {...{
+          number: quantity * pricePerUnit,
+          formatter: moneyString,
+        }}
+      />
+      )
+    </span>
+  )
+}
+
 export const Item = ({
   completedAchievements,
   handleItemPurchaseClick,
@@ -203,6 +221,12 @@ export const Item = ({
                       <AnimatedNumber
                         {...{ number: adjustedValue, formatter: moneyString }}
                       />
+                      <TotalPrice
+                        {...{
+                          quantity: purchaseQuantity,
+                          pricePerUnit: adjustedValue,
+                        }}
+                      />
                     </span>
                   </Tooltip>
                   {completedAchievements['unlock-crop-price-guide'] &&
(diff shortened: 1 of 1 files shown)


jeremyckahn (changes requested):
Thanks for making this addition, @tim-phillips! The code here looks good, but I'd like to get the button overlap issue resolved before merging this. For the Sell view, I think wrapping the parenthesized section to the next line should be sufficient.

Also, this seem like something worth adding test coverage for. We have `src/components/Item/Item.test.js` in place, but that file is using Enzyme (it's old...). These days, Farmhand's new tests use React Testing Library. We can either add RTL tests alongside the Enzyme tests in the same file, or we can spin up a new file for RTL tests at `src/components/Item/Item.rtl.test.js` (something we've done for other components that have a mix of Enzyme and RTL tests).

Let me know if you'd like some support with making these changes!

tim-phillips:
Glad to make the change! I agree that the button overlap issue needs addressing. I futzed around with the tests but couldn't achieve anything useful with enzyme. I'm glad to hear that other tests are using RTL! I'll take that approach and add some here. 

Since quantity can be `undefined` when the user first clicks on the input, the calculation will result in `NaN`, which is why I opted to only show the total when quantity is greater than 1. With this change, in order to prevent line jumps I can show "Total" all of the time. What do you think?

jeremyckahn:
> With this change, in order to prevent line jumps and card resizing I can show "Total" all of the time. What do you think?

For this first iteration, let's go with just showing "Total" at all times to avoid the line jumping. We can improve that later if it proves to be annoying!

tim-phillips:
I added some tests and changed Total to live on its own line for both buy and sell. I think it's a great improvement. Thanks for talking all that through!

jeremyckahn:
Thanks for the updates @tim-phillips! I don’t have much free time in the next few days, but I will prioritize reviewing this when I can.

src/components/Item/Item.test.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
-import CardHeader from '@mui/material/CardHeader'
-import { shallow } from 'enzyme'
```
jeremyckahn, line 3: Amazing! Thank you for getting Enzyme out of this file. :pray:

src/test-utils/index.js
```diff
@@ -28,5 +28,8 @@ export const testItem = (item = {}) => ({
   id: '',
   name: '',
   value: 0,
+  description: '',
+  doesPriceFluctuate: false,
+  isReplantable: false,
```
jeremyckahn, line 33: :heart:

jeremyckahn (approved):
LGTM! Thanks for this excellent addition to Farmhand, @tim-phillips.

I just realized that I missed this question:

> Also, would "Total sell price" be better copy?

In the interest of minimalism, I think that "Total" as you have it now should be fine. We can revisit this in later iterations if we end up not liking it.

I think this is ready to merge, so I will do so now and release it as version 1.18.9.

Congrats on your first contribution to Farmhand! :tada:

PR #493, opened 2024-04-19
feat(sprinklers): Increase Sprinkler range at level 25

jeremyckahn:
### What this PR does

This PR increases the Sprinkler range to 4 at level 25 (previous max was 3).

- [ ] Reach level 25 and verify that Sprinklers cover all plots that are 4 spaces out from them.

### Additional information

Sprinkler range of 4 screenshot


jeremyckahn:
It turns out that this is unneeded because the preexisting range of 3 allows for symmetrical Field coverage.

PR #494, opened 2024-04-28
feat(multiplayer): Optimize API usage

jeremyckahn:
### What this PR does

This PR optimizes multiplayer API usage by eliminating polling. Rather than using the Vercel-based central API to determine room occupancy, the number of connected peers is used.

:bangbang: Please **DO NOT** test this in the default `global` room, as that affects Production data. Instead, use another room such as `global-test`.

- [ ] Ensure that peers can connect to each other
- [ ] Ensure that the disconnection of peer A is (eventually) accurately reflected in peer B
- [ ] Ensure that there are no noticeable regressions with online play (note: cow trading should be unaffected by this change)

### Additional information

This change is needed because Farmhand is getting so much multiplayer activity that we're exceeding the capacity of the free Vercel tier that is powering it.

### DIFF
--- a/api-etc/constants.js
+++ b/api-etc/constants.js
@@ -7,5 +7,3 @@ export const ACCEPTED_ORIGINS = new Set([
   'https://www.farmhand.life',
   'https://v6p9d9t4.ssl.hwcdn.net', // itch.io's CDN that the game is served from
 ])
-
-export const MAX_ROOM_SIZE = 25
--- a/api/get-market-data.js
+++ b/api/get-market-data.js
@@ -5,8 +5,6 @@ const { promisify } = require('util')
 
 require('redis')
 require('../src/common/utils')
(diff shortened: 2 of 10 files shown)


jeremyckahn:
It looks like this PR can't be validated in the Preview Environment because of CORS domain restrictions. It can only really be tested on my machine because I can't give access to the Vercel environment without paying for the Pro tier. So, we'll just have to settle for a code review of this PR.

jeremyckahn:
> I'm noticing an issue in which the player cannot disconnect properly in the Preview Deployment, so I will need to get that fixed before merging this.

This should be fixed with the latest changes! This is ready for review.

lstebner (approved):
nice work! the code all looks good to me and i did some very basic testing, but things were working just fine. it was still very snappy to notice connect and disconnect events even. :tada:

PR #506, opened 2024-08-29
feat(project): Replace Create React App with Vite

jeremyckahn:
### What this PR does

Drop Create React App in favor of Vite. CRA is unmaintained, so this project would do well to transition to something more modern.

- [ ] Run all the scripts
- [ ] Fully test the game for regressions
- [ ] Hope for the best

I tried to keep the scope of this change as narrow as possible, but it still touches pretty much everything in the project. We'll probably find something broken after the fact, so we'll just have to fix forward if that's the case.


src/components/Field/Field.test.js
```diff
@@ -83,39 +83,6 @@ describe('field rendering', () => {
...
-describe('harvest-mode class', () => {
-  test('is not present when fieldMode != HARVEST', () => {
-    expect(component.find('.Field').hasClass('harvest-mode')).toBeFalsy()
-  })
-
-  test('is present when fieldMode == HARVEST', () => {
-    component.setProps({ fieldMode: fieldMode.HARVEST })
-    expect(component.find('.Field').hasClass('harvest-mode')).toBeTruthy()
-  })
-})
```
jeremyckahn, line 117: These and the removed tests below are obviated by `src/shell/field.test.js`, introduced by this PR: https://github.com/jeremyckahn/farmhand/blob/2b8721e394c9e3eae24263f42ec29ff6f8d5d110/src/shell/field.test.js#L15-L81

src/game-logic/reducers/purchaseForest.js
```diff
@@ -22,7 +22,7 @@ export const purchaseForest = (state, forestId) => {
   const { columns, price, rows } = PURCHASABLE_FOREST_SIZES.get(forestId)
 
   /*
-   * FIXME: using FOREST_AVAILABLE_NOTIFICATION here is temporary, this code path will
+   * TODO: using FOREST_AVAILABLE_NOTIFICATION here is temporary, this code path will
```
jeremyckahn, line 25: FYI @lstebner I changed this from a `FIXME` to a `TODO`. I find it helpful to treat `FIXME`s as an indication that something that needs to be addressed before a PR is merged, whereas a `TODO` indicates an improvement to be optionally made at some point in the future.

src/handlers/ui-events.test.js
```diff
@@ -38,20 +21,6 @@ describe('handleFieldModeSelect', () => {
...
-    })
-
-    test('updates fieldMode state', () => {
-      expect(component.state().fieldMode).toEqual(fieldMode.PLANT)
-    })
-
-    test('does not change state.selectedItemId', () => {
-      expect(component.state().selectedItemId).toEqual('sample-crop-3')
-    })
-  })
```
jeremyckahn, line 53: These have also been obviated by `src/shell/field.test.js`.

Issue #522, opened 2024-10-07
Make spacer display responsive in the menu


lillastormen:
I noticed that there is a spacer in the menu, which creates scroll bar when it's not needed for certain window heights. The spacer is likely there to prevent navigation buttons to cover the content of the menu on smaller screens. I'd want to suggest hiding the spacer for the menu for screens wider than 1k pixels.

jeremyckahn:
Hi @lillastormen, I think I know what you're talking about, but would you mind posting some screenshots to make sure we're talking about the same thing?

lillastormen:
Hi @jeremyckahn, 

here's the spacer in question:
![Screenshot 2024-10-08 at 23 20 15](https://github.com/user-attachments/assets/fc3a2191-ca72-4e83-99c7-88c6e8d553fa)

As you can see it creates a scroll bar when it's not needed. My idea was that maybe hiding it, when it's not needed, could create a bit cleaner interface :)

And here's the screenshot, where the spacer is removed: 

![Screenshot 2024-10-08 at 23 20 15](https://github.com/user-attachments/assets/337b3a27-1336-4737-a1ef-3575de0d9ca0)

jeremyckahn:
Got it! Thanks for clarifying, @lillastormen. As you deduced, that spacer is there to make sure that the navigation buttons at the bottom don't obscure any content on mobile devices:

![Screenshot of Farmhand on a small screen](https://github.com/user-attachments/assets/88e3b749-9c0d-47b2-82a9-dd527f8b8821)

I agree that it would be worth hiding the spacer on larger screens where it's not necessary. Good thinking @lillastormen! Feel free to open a PR. 🙂

PR #523, opened 2024-10-09
#522 adding media query to hide spacer at certain breakpoint 

lillastormen:
Simple change that hides the spacer in the manu above medium window width breakpoint.

### DIFF
--- a/src/components/Farmhand/Farmhand.sass
+++ b/src/components/Farmhand/Farmhand.sass
@@ -260,6 +260,8 @@ body
   // interacting with the elements within them.
   .spacer
     min-height: 7.5em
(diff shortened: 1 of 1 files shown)


jeremyckahn:
Thanks for opening this PR, @lillastormen! The Menu layout looks good, but it seems that this change has introduced an issue with the Home screen content being obscured by the buttons:

Screen shot of obscured content

I think this would affect all "Stage" screens (game content that's not in the Menu), not just the Home screen.

This doesn't appear to be happening in [Production](https://www.farmhand.life/). Would you be willing to address this? Once that's fixed, we can merge and ship this!

jeremyckahn:
Closing due to inactivity.

PR #526, opened 2024-11-25
#525 List search: added SearchBar, added new filter to inventory

az1plo:
**What this PR does**

- Added SearchBar component for filtering lists.
- Integrated SearchBar into the following sections:
   - Inventory
   - Shop
   - Workshop
   - Cows
   - Cellar
- Implemented a new filter logic for the inventory to enhance usability.

**How this change can be validated**

- Navigate to each section listed above and use the search bar to filter items.
- Verify that the displayed items match the search query.

### Questions or concerns about this change

- Are there sections where the SearchBar could be removed or added?

### Additional information

Screenshot 2024-11-25 113651
Screenshot 2024-11-25 113712
Screenshot 2024-11-25 113916
Screenshot 2024-11-25 113951
Screenshot 2024-11-25 114026
Screenshot 2024-11-25 113846

PR #528, opened 2024-11-29
fix(closes #527): Enable cow renaming

### LINKED ISSUE #527: Cow renaming is broken
Cows cannot be given custom names.

jeremyckahn:
### What this PR does

This PR fixes #527, a regression that prevented players from giving their cows custom names.

Successfully rename a cow (this is only possible with cows that were not received via trade).

I have no idea how long this was broken for. :sob:


tsconfig.json
```diff
@@ -46,7 +46,7 @@
     // "noPropertyAccessFromIndexSignature": true,  /* Require undeclared properties from index signatures to use element accesses. */
 
     /* Module Resolution Options */
-    // "moduleResolution": "node",                  /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+    "moduleResolution": "node" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
```
jeremyckahn, line 49: This seems to make type checking work better.

src/components/CowCard/CowCard.test.js
```diff
@@ -342,4 +343,40 @@ describe('CowCard', () => {
...
+
+      renderComponent()
+
+      const nameInput = screen.getByPlaceholderText('Name')
+
+      const customName = 'Custom'
+      userEvent.clear(nameInput)
+      userEvent.type(nameInput, customName)
+
+      await waitFor(() => {
```
lstebner, line 377: i think you should be able to await the userEvent.type calls and avoid waitFor entirely

src/hooks/useMountState/useMountState.js
```diff
@@ -11,7 +11,7 @@ export const useMountState = () => {
     }
   }, [isMountedRef])
 
-  const isMounted = () => isMountedRef.current
+  const isMounted = useCallback(() => isMountedRef.current, [isMountedRef])
```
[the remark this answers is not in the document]
lstebner: using a ref in this code is kind of strange to me. if you like the ref use this isn't a big deal but i would have written this hook like:

```js
const [isMounted, setIsMounted] = useState(false)

useEffect(() => setIsMounted(true), [])

return { isMounted }
```

in general though having isMounted checks is a code smell for me so it may be more
worthwhile to try to refactor all the places this hook is needed than to worry about rewriting it right now.

lstebner (approved):
leaving some thoughts but i'm also ok with a merge to get the bug fixed. i don't need to hold that up over semantics 😄

src/hooks/useMountState/useMountState.js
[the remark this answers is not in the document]
jeremyckahn: I'm glad you raised this because it made me realize that #526 introduced the inclusion of https://usehooks-ts.com/. This library provides [an implementation of `useIsMounted`](https://usehooks-ts.com/react-hook/use-is-mounted) which effectively abstracts away the contents of this hook. Ironically [the library implementation](https://github.com/juliencrn/usehooks-ts/blob/20667273744a22dd2cd2c48c38cd3c10f254ae47/packages/usehooks-ts/src/useIsMounted/useIsMounted.ts#L14-L26) seems equivalent to what we have here, but since it's in a library we don't have to worry too much about the implementation details so long as it works (which it does).

I've switched to using the `usehooks-ts` implementation in https://github.com/jeremyckahn/farmhand/commit/72d110b2155f1df3d7f631c27573997aa591da82 to eliminate the maintenance burden of our custom one.

> in general though having isMounted checks is a code smell for me so it may be more
worthwhile to try to refactor all the places this hook is needed than to worry about rewriting it right now.

I generally agree and try to avoid using `isMounted` checks for this reason. In this case it was necessary for Cow image generation: https://github.com/jeremyckahn/farmhand/blob/72d110b2155f1df3d7f631c27573997aa591da82/src/components/CowCard/CowCard.js#L129-L136

In this part of the code, the Cow image is generated dynamically as an asynchronous operation. By the time the image generation is complete, it's possible that the component has unmounted which would mean `setCowImage(cowImage)` would be improperly setting state on an unmounted component. There are probably cleaner abstractions for this sort of thing, but this seemed like the simplest solution. I don't know that we'd want to repeat this pattern too much, but I think this is the only part of the code base that we do this right now so I think that this bit of isolated awkwardness is acceptable.

src/components/CowCard/CowCard.test.js
jeremyckahn: > i think you should be able to await the userEvent.type calls and avoid waitFor entirely

It turns out that `clear` and `type` only return Promises in `@testing-library/react` v14. Farmhand currently uses v13, and upgrading to v14 breaks a number of other tests so it's not worth making that change in a narrowly-scope bugfix PR such as this.

However, it appears that the `await waitFor(...)` isn't even necessary here (I think it was for an earlier iteration), so I've removed it in https://github.com/jeremyckahn/farmhand/commit/9601f41a3dae4b43975f742f1bde9ec726e4e72c. Thanks for bringing this up helping to simplify this part of the code!

src/components/CowCard/CowCard.js
```diff
@@ -123,7 +123,7 @@ export const CowCard = (
     isCowOfferedForTradeByPeer && cowIdOfferedForTrade.length > 0
   )
 
-  const { isMounted } = useMountState()
+  const isMounted = useIsMounted()
```
jeremyckahn, line 126: As of https://github.com/jeremyckahn/farmhand/commit/72d110b2155f1df3d7f631c27573997aa591da82, this is now the solution to the root cause of #527.

jeremyckahn (commented):
Thanks for the review @lstebner! The code is looking better with your feedback incorporated, as always. 🙂

I'll merge this now!

src/hooks/useMountState/useMountState.js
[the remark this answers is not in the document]
lstebner: oh that's great we have a library version of this to use instead! also if it's only used for that one use case that's not a big deal at all. i did not look around for the uses

src/components/CowCard/CowCard.test.js
lstebner: ah right, i forgot the switch to async was in later versions

PR #529, opened 2024-11-29
Fix missing space between hugging machine and and

Danielv123:
### What this PR does

Fixes this typo 
image

### DIFF
--- a/src/components/CowCard/Subheader/Subheader.js
+++ b/src/components/CowCard/Subheader/Subheader.js
@@ -150,7 +150,7 @@ const Subheader = (
               placement: 'top',
               title: (
                 <Typography>
-                  Check this box to put {cow.name} into a {huggingMachine.name}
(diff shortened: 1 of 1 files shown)


src/components/CowCard/Subheader/Subheader.js
```diff
@@ -150,7 +150,7 @@ const Subheader = (
               placement: 'top',
               title: (
                 <Typography>
-                  Check this box to put {cow.name} into a {huggingMachine.name}
+                  Check this box to put {cow.name} into a {huggingMachine.name}
```
jeremyckahn, line 153: I think tweaking this would ensure that the change doesn't get wiped out by automatic whitespace-removal tools:
```suggestion
                  Check this box to put {cow.name} into a {huggingMachine.name}{' '}
```

jeremyckahn (changes requested):
Thanks for raising this issue and providing a fix, @Danielv123! I'd like to make one small tweak to make this more robust before approving and merging.

PR #533, opened 2024-12-17
fix(multiplayer): Use torrent strategy for Trystero

jeremyckahn:
### What this PR does

There are two themes spread across three commits for this PR:
- A variety of type fixes:
  - https://github.com/jeremyckahn/farmhand/commit/13c5e112406d1542567cca92116f5a8b5b35d365
  - https://github.com/jeremyckahn/farmhand/commit/06db008d16e540eff0f7bc55576eaf50c7cfd3b3
  - I don't recommend spending too much time reviewing these. It's pretty unexciting and not a great use of reviewer time. A quick skim should be sufficient.
- A regression fix (from #509) for multiplayer where the incorrect connection strategy was being used: https://github.com/jeremyckahn/farmhand/commit/b7c5d16ce9d0c377e2a8fea81be651d94ea22ca0
  - This is the commit that's actually worth reviewing

- [ ] Verify that multiplayer works

### DIFF
--- a/api-etc/utils.js
+++ b/api-etc/utils.js
@@ -56,7 +56,7 @@ export const allowCors = fn => async (req, res) => {
 
   if (
     ACCEPTED_ORIGINS.has(origin) ||
-    origin.match(/https:\/\/farmhand-.*-jeremyckahn.vercel.app/)
+    origin.match(/https:\/\/farmhand-.*-jeremy-kahns-projects.*.vercel.app/)
   ) {
     res.setHeader('Access-Control-Allow-Origin', origin)
   }
--- a/api/publish-global-market-values-to-discord.js
+++ b/api/publish-global-market-values-to-discord.js
@@ -22,10 +22,15 @@ const set = promisify(client.set).bind(client)
 export default allowCors(async (req, res) => {
   const { valueAdjustments } = await getRoomData('room-global', get, set)
   const content = MARKET_SUMMARY_FOR_DISCORD`${'global'}${valueAdjustments}`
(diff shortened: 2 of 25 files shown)


lstebner (commented):
i was able to connect to the shared room between two clients, but i gotta be honest, i'm not really sure if the rest is working or not. i keep selling carrots with my one player trying to influence the price for the other, but the prices seem to still be random for both players. am i forgetting something?

jeremyckahn:
> i was able to connect to the shared room between two clients, but i gotta be honest, i'm not really sure if the rest is working or not. i keep selling carrots with my one player trying to influence the price for the other, but the prices seem to still be random for both players. am i forgetting something?

Yikes, you're right. I'm seeing this in Production as well. That means it's not a regression introduced by this PR, but I'm wondering if it came out of #494. That was a somewhat hasty redesign of how multiplayer works that was necessitated by Vercel pricing changes. It could also be due to the inherent flakiness of the public pairing infrastructure we're relying on.

Either way, I'll address this issue and [the related one you brought up in Discord](https://discord.com/channels/714539345050075176/847884323330850847/1318435216259289108) in a followup PR and prioritize getting things stable again.

@lstebner Did you notice any new issues introduced by this PR?

lstebner (approved):

lstebner:
nope, nothing else. i just wanted to make sure i was testing correctly before approving

jeremyckahn:
Thanks for the thorough review as always, @lstebner! I'll prioritize fixing the issues you discovered and wait to release until they are resolved.

PR #534, opened 2024-12-18
fix(multiplayer): Make multiplayer stable again

jeremyckahn:
### What this PR does

This PR fix two bugs with multiplayer discovered by @lstebner:
  - Price instability: https://github.com/jeremyckahn/farmhand/pull/533#pullrequestreview-2507834230
    - Fixed in c0e82c10ba2bad022112fc97ed2f3d31bdb29393
  - Inconsistent online mode toggle behavior
    - >  it seems like if i enter a custom room name and then toggle "play online", the toggle quickly toggles back to looking disabled, but the room is ultimately created and i am connected online. if i then toggle it again, it will say it connected to the room and this time it will stay looking enabled.
    - Fixed in 07b7249c4d442ae2d1da4ee12c443128eb74f9a9

See if the reported issues above remain.

How long was this broken for? We will never know. :sob:

### DIFF
--- a/api-etc/utils.js
+++ b/api-etc/utils.js
@@ -32,7 +32,7 @@ export const getRoomData = async (roomKey, get, set) => {
 
   if (!valueAdjustments) {
     valueAdjustments = generateValueAdjustments()
-    roomData = { activePlayers: {}, valueAdjustments }
+    roomData = { valueAdjustments }
     set(roomKey, JSON.stringify(roomData))
   }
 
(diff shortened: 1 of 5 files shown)


lstebner (commented):
i think the primary issues this PR set out to fix are fixed, but i think the count for players connected to a room is wrong now. i don't know if this is a big deal, or even related to this PR. i see something with "active players" being removed that smells related though. i'm currently playing in 3 windows and getting counts 1, 1, and 2 when i expect to see 3 in all. 

image

lstebner (approved):

lstebner:
i did do several refreshes and some reconnecting in hopes it would resolve but it never did. i did see everyone mentioned when i opened chat though. so definitely just some flakiness

PR #548, opened 2025-06-02
feat(infra): set up end-to-end testing

jeremyckahn:
### What this PR does

This PR sets up infrastructure for end-to-end (E2E) testing and introduces some E2E tests for baseline app functionality.

It's a little involved to run the E2E tests locally as they run within a GitHub Actions job, but it is possible to do so. Instructions for this are in the README.

For the purposes of validating this PR, we just need to see that all the CI checks (particularly the "e2e" job) pass.

The current implementation of this E2E test infrastructure prioritizes remote, CI-based execution of the tests over local. Presumably, that's where these tests will run and matter the most. It's possible to speed up local execution of the tests by standing up additional configuration that's specific to local execution, but the current solution is complicated enough as-is and I didn't want to push even harder in that direction just yet. It might be worth revisiting this if local execution becomes too burdensome.

### Additional information

Successful local execution of the tests:

Screenshot of local E2E test passing


vite.config.js
```diff
@@ -99,6 +102,7 @@ const vitestConfig = vitestDefineConfig({
     environment: 'jsdom',
     setupFiles: './src/setupTests.js',
     restoreMocks: true,
+    dir: 'src',
```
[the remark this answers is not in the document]
jeremyckahn: This is intended behavior to ensure that Vitest only looks in `src` for unit/integration tests and not the `e2e` directory.

e2e/tests/price-fluctuation.js
```diff
@@ -0,0 +1,17 @@
+import { expect, test } from '@playwright/test'
```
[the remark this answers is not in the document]
jeremyckahn: Great catch. This is fixed in https://github.com/jeremyckahn/farmhand/commit/223e5f82bb8c8eb259a733e381b77d5491173ade.

e2e/docker-compose.yml
```diff
@@ -0,0 +1,108 @@
...
+    volumes:
+      - ./tests:/home/pwuser/e2e/tests
+    # Use ipc: host as recommended by Playwright docs for Chromium
+    ipc: host
+    # Set the working directory explicitly to match the WORKDIR in the Dockerfile
+    working_dir: /home/pwuser/e2e
+    # Run as root to avoid permission issues
+    user: root
+    entrypoint: ['/bin/sh', '-c']
+    command: ["exec su root -c 'yes | npx playwright test tests'"]
```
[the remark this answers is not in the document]
jeremyckahn: I tried making this change in https://github.com/jeremyckahn/farmhand/commit/bf599ef05a51bc210e5a573151a2ed5f2c263a09 but it caused a CI failure: https://github.com/jeremyckahn/farmhand/actions/runs/15393361741/job/43307799923

It has been reverted in https://github.com/jeremyckahn/farmhand/commit/486a3a59f453b77732dfc9db2808e3de7e997216.

e2e/tests/crop-growth.test.js
```diff
@@ -0,0 +1,71 @@
...
+  await page.getByRole('button', { name: 'Scarecrow' }).click()
+  await page
+    .locator('.Plot')
+    .first()
+    .click()
+
+  // Plant Carrot Seed
+  await page.getByRole('button', { name: 'Carrot Seed' }).click()
+  await page
+    .locator('.row > div:nth-child(2)')
```
lstebner, line 36: why did the scarecrow use `locator('.Plot')` but this one used this nth-child selector?

related.. might be worth moving these to a testid instead so they aren't so dependent on HTML elements/classNames, or even better, add some hidden text that identifies the element as "plot #" so that a screen reader could also tell the difference

e2e/tests/crop-growth.test.js
```diff
@@ -0,0 +1,71 @@
...
+  // Water Carrot Seed
+  await page
+    .getByRole('button', { name: 'A watering can for hydrating' })
+    .click()
+  await page
+    .locator('.row > div:nth-child(2)')
+    .first()
+    .click()
+
+  await page.getByRole('button', { name: 'End the day to save your' }).click()
```
lstebner, line 65: seeing this selector repeated a bunch makes me think we'll want to utilize the [Page Object Model](https://playwright.dev/docs/pom) in the future. then we can have a place for these selectors to live for easier maintenance, and also to make the tests easier to write. 

the example in the docs isn't super thorough, but what I am imagining is giving us a syntax more like `await farmhandPage.endDay()` here. likewise we could have easy access to changing pages (`farmhandPage.goToField()`) and selecting tools (`fieldPage.selectWateringCan()`) etc. this certainly does not need done in this PR, but it could be a future refactor/optimization to consider

e2e/tests/load-save.test.js
```diff
@@ -0,0 +1,63 @@
...
+})
+
+test('should show overnight notifications from previous day after loading save', async ({
+  page,
+}) => {
+  await openPage(page, 0.3)
+
+  await page.getByRole('button', { name: 'End the day to save your' }).click()
+
+  await expect(page.locator('#root')).toContainText('It rained in the night!')
```
lstebner, line 19: is this not in some sort of "alert" element? selecting `#root` is the entire app, right?

jeremyckahn:
@lstebner thank you for the excellent review! I think these are all great suggestions, but instead of addressing them now I'm going create some followup tickets and get to them afterwards. Reason being: The current E2E testing infrastructure, while very functional, is _painfully_ slow to run locally. Your suggestions will require a bit of iteration to get right, and for the sake of my sanity I'd like to wait to address them until after I improve the local execution speed of the tests. I will get to them, though!

The reason the local execution time is so slow is because the job is executed via [`act`](https://github.com/nektos/act), a tool for running GitHub Actions locally. As seen in `ci.yml`, the job attempts to download a prebuilt Docker image tied to the current commit hash from GitHub Packages. If it's unavailable (which it won't be for new commits), the image is built and pushed to GitHub Packages for quicker subsequent test runs. This makes a lot of sense in the context of a CI system (E2E tests are liable to be flaky and rerunning them should be minimally painful), but it is _extremely_ heavy in the context of local test development. This could be somewhat worked mitigated with some development workflow contortions, but it seems better to solve the problem directly and improve local E2E infrastructure instead. I left that out of this PR to try and reign in the scope, but clearly there's a need for it!

I will make a ticket to improve the local execution speeds, and tickets to implement your suggestions once it's easier to do so!

e2e/tests/load-save.test.js
jeremyckahn: This is what the [Playwright codegen tools](https://playwright.dev/docs/codegen-intro) provided. So far I've been taking the approach of not questioning it too much and just going with what it provides almost verbatim. There's some obvious brittleness (and verbosity) with that approach, but the idea is that it won't be "scary" to throw out a broken test when it inevitably gets out of date and rebuild it with the codegen tools.

This may be a flawed approach to E2E testing, but it's a pattern I've seen previously for high-level tests like this. I'm not an E2E testing SME, so I don't have a strong sense of what a best practice for this is. I'd love to get your perspective on that, @lstebner!

Regardless, I'm going to wait until we have better local E2E testing facilities (per [my other comment](https://github.com/jeremyckahn/farmhand/pull/548#issuecomment-2947752137)) to address this.

e2e/tests/crop-growth.test.js
jeremyckahn: Great idea. #552 has been filed to address this!

e2e/tests/crop-growth.test.js
jeremyckahn: I think that makes a lot of sense. I've filed #553 to organize a more focused effort around it!

PR #575, opened 2025-10-08
feat: Tumbleweeds

jeremyckahn:
### What this PR does

This PR implements tumbleweeds that are shown when there are no cows in the cow pen. The spawn at an increasing rate up to a limit and stop spawning when the player acquires a cow. They resume spawning when the player no longer has cows.

Empty out your cow pen and watch the calming progression of tumbleweeds as you contemplate life on the farm.

### Additional information

There's more we'd like to do with this addition, but this is a good first step.

[A screencast of the tumbleweeds in the cow pen](https://github.com/user-attachments/assets/1de1654d-95cd-465f-b6a2-50b492463f44)


src/components/CowPen/Cow.js
jeremyckahn, line 1: This was moved out of `CowPen.js`.

src/components/CowPen/Tumbleweeds.js
```diff
@@ -0,0 +1,153 @@
...
+        // If spawning is still active, schedule a new tumbleweed to maintain a constant stream.
+        if (doSpawn) {
+          scheduleSpawn()
+        }
+
+        return reducedTumbleweeds
+      })
+    },
+    [doSpawn, scheduleSpawn]
+  )
```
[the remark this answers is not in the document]
jeremyckahn: I'd prefer not to do this as it would be replacing an `O(1)` operation with an `O(n)` one.

src/components/CowPen/Tumbleweeds.js
```diff
@@ -0,0 +1,153 @@
...
+// The size in pixels of the tumbleweed sprite
+const tumbleweedSize = 48
+
+/**
+ * A single tumbleweed that animates across the screen.
+ * @type {React.FC<{onAnimationComplete: () => void}>}
+ */
+const Tumbleweed = ({ onAnimationComplete }) => {
+  // Randomize the vertical position of the tumbleweed.
+  const [yPercent] = useState(Math.random())
```
lstebner: i think this should be using the random number service you created. also a small nit-pick on the name, maybe `yPosition` instead of `yPercent`?

src/components/CowPen/Tumbleweeds.js
```diff
@@ -0,0 +1,153 @@
...
+    tumbleweedUuid => {
+      setTumbleweeds(prev => {
+        const idxTumbleweed = prev.indexOf(tumbleweedUuid)
+
+        const tumbleweedsWithCompletedElementRemoved = [
+          ...prev.slice(0, idxTumbleweed),
+          ...prev.slice(idxTumbleweed + 1),
+        ]
+
+        return tumbleweedsWithCompletedElementRemoved
```
lstebner, line 140: personal preference here, but i think filtering could be easier to read. something like:

```js
setTumbleweeds(prev => prev.filter((id) => id === tumbleweedUuid ? null : id).filter(Boolean).slice())
```

hmm, after writing this out i'm not sure it's actually that much more clear... going to leave this comment anyways though as food for thought.

src/components/CowPen/Tumbleweeds.js
jeremyckahn: Excellent suggestion! This is addressed in https://github.com/jeremyckahn/farmhand/commit/6dc476a02ed53c5e220a82b5bf61596c653a653f. Thanks for catching that!

src/components/CowPen/Tumbleweeds.js
jeremyckahn: Gemini made a similar suggestion: https://github.com/jeremyckahn/farmhand/pull/575#discussion_r2412359598

I decided to keep it as-is because this is an `O(1)` operation. This is a slightly hot (warm?) code path, so it seems worth keeping optimized since it doesn't muddy up the code terribly (IMO).

lstebner: ah, after seeing gemini's version i see that the extra `.filter(Boolean).slice()` moves are excessive. the version i wrote was not as good as geminis (i should have tested it :sweat: ). i think without those chained calls the filter approach is actually more optimized. the spread approach has to create 2 temp arrays in memory and then merge them into another one. I'm not really good with figuring out O-values, but some quick research makes me think both approaches are linear.

i'm still fine with either version though, i'm pretty sure JS will have good performance with far larger arrays than we'll ever have for tumbleweeds.

jeremyckahn: I'm actually curious now.

@gemini-code-assist: What are the performance benefits and tradeoffs of the two approaches being discussed here? What is the Big-O complexity of the current implementation as well as the alternative solution being proposed?

jeremyckahn: lol I went with the worst possible solution. :joy:

I'll fix it in another PR!

jeremyckahn: 🫣 https://github.com/jeremyckahn/farmhand/pull/576

lstebner: dang that's a cool analysis you summoned there!

PR #672, opened 2026-07-08
Refactor: Convert Farmhand component from class-based to functional

jeremyckahn:
## Summary

Converts `Farmhand` from a ~1600-line class component into a functional component decomposed across a single orchestrating hook (`useFarmhand`) and four focused sub-hooks, plus a static helper module for the pure/memoized selectors that used to live as class methods. `Farmhand.tsx` itself is now ~190 lines of JSX with no state or lifecycle logic.

The refactor was done incrementally and surfaced a number of real regressions along the way (races, stale closures, effect-dependency bugs); those fixes and their regression tests are bundled into this same PR rather than split out, since they were discovered and fixed as part of validating the new architecture.

## Before / after

**Before:** one class component owned `this.state`, all lifecycle methods (`componentDidMount`/`componentDidUpdate`/`componentWillUnmount`), all bound instance methods, and delegated actual event/network logic to `ui-events.tsx` and `peer-events.tsx`, which took the class instance (`this`) as their first argument.

**After:** state lives in a single `useState` inside `useFarmhand`; each concern that used to be a slice of `componentDidUpdate`/instance methods is now its own hook; and `ui-events.tsx`/`peer-events.tsx` continue to work **unmodified** against an `instanceProxy` object that reconstructs the shape of the old class instance (see "Compatibility shim" below).

## Module breakdown

| Module | Lines | Responsibility |
|---|---|---|
| `Farmhand.tsx` | 189 | Pure presentational component. Calls `useFarmhand(props)` and renders JSX from the returned `gameState`/`handlers`. No state, no effects. |
| `useFarmhand.ts` | 738 | Orchestrator. Owns the single `farmhand.state` via `useState`, runs the boot effect (restore persisted state or start a new game), the post-increment-day persistence effect, the unmount cleanup effect, and the achievements/menu/money `componentDidUpdate`-equivalent effect. Wires all sub-hooks together and assembles `instanceProxy`. |
| `useFarmhandReducers.ts` | 38 | Reflects over `FarmhandReducers.prototype` method names and auto-binds every pure reducer in `game-logic/reducers` to `setState`, so adding a new reducer doesn't require manually wiring a new bound method. |
| `hooks/useFarmhandNavigation.ts` | 58 | Dialog view open/close and stage-focus (view) navigation, including the previous/next-view keyboard & FAB handlers. |
| `hooks/useFarmhandNetwork.ts` | 400 | The largest sub-hook: Trystero P2P room join/leave, heartbeat scheduling, cow-trade request/accept/reject/timeout flow, and the online-status/room `componentDidUpdate` reconciliation effect (including the redirect-race fix described below). |
| `hooks/useFarmhandDayAdvancement.ts` | 167 | `incrementDay`, posting positions to the server for market value adjustments, and persisting state to `localforage`. |
| `hooks/useFarmhandNotifications.ts` | 54 | Inventory-full and recipe-learned toast triggers, diffed against previous state. |
| `usePrevious.ts` | 11 | Small utility hook exposing the previous render's value, replacing the class's manual `prevProps`/`prevState` capture in `componentDidUpdate`. |
| `FarmhandService.ts` | 72 | Static, memoized pure selectors extracted from class methods: `computePlayerInventory`, `getFieldToolInventory`, `getPlantableCropInventory`, `applyPriceEvents`. |

## Compatibility shim: `instanceProxy`

`ui-events.tsx` and `peer-events.tsx` were written against a class instance and call things like `this.purchaseItem(...)`, `this.state`, `this.setState(...)`. Rewriting every handler to hooks would have made this a much larger, riskier PR. Instead, `useFarmhand` builds an `instanceProxy` object (`useMemo`, mirrored into a ref every render) that exposes the same shape — current `state`, a `setState` that merges like the class's did, and all bound reducers/handlers — and passes that proxy as `this` to the existing handler modules. This let the handler files stay untouched while the component underneath was fully restructured.

```mermaid
graph TD
    subgraph Presentation
        A["Farmhand.tsx (JSX only)"]
    end

    subgraph Orchestration
        B["useFarmhand.ts<br/>useState(farmhand.state) + boot/persist/cleanup effects"]
    end

    subgraph "Sub-hooks"
        C["useFarmhandNavigation<br/>dialogs, view focus"]
        D["useFarmhandNetwork<br/>P2P room, heartbeat, cow trades"]
        E["useFarmhandDayAdvancement<br/>incrementDay, server sync, persistence"]
        F["useFarmhandNotifications<br/>inventory-full / recipe-learned toasts"]
        G["useFarmhandReducers<br/>auto-binds game-logic reducers to setState"]
        H["usePrevious<br/>prior-render state snapshot"]
    end

    I["FarmhandService<br/>memoized pure selectors"]

    subgraph "Unmodified legacy handlers"
        J["ui-events.tsx"]
        K["peer-events.tsx"]
    end

    A -- props --> B
    B --> C
    B --> D
    B --> E
    B --> F
    B --> G
    B --> H
    B --> I
    B -- assembles --> L["instanceProxyRef<br/>(mimics class 'this')"]
    L --> J
    L --> K
    J -- "bound handlers" --> A
    B -- "gameState + handlers" --> A
```

## Regressions found and fixed during the refactor

Decomposing `componentDidUpdate` into separate `useEffect`s changed the ordering and dependency-tracking guarantees the old code relied on, which surfaced several real bugs:

- **Online-toggle race** (`64b54adc`): toggling online/offline triggered a redirect, but the network effect re-read the stale `path` before the router had caught up, immediately stomping the just-set `isOnline`. Fixed by detecting "a redirect was just requested" and skipping reconciliation for that render (`useFarmhandNetwork.ts`, the `redirectJustRequested` check).
- **Peer-room disconnect / room-fallback bugs** (`f266266d`): leaving a room or falling back to the default room left stale peer/heartbeat state.
- **Cow-sale persistence race** (`850a3b6e`, `85e9789a`): a day-increment could race with an in-flight persistence write, dropping the cow sale from the saved state.
- **Stale closures and state-persistence timing** (`949c81d2`, `b6d27ec1`, `c5fc0477`): several callbacks captured `state` from the render they were created in instead of the latest value, addressed by routing through `stateRef`/`instanceProxyRef` where a callback needs the current state rather than the state at closure-creation time.
- **React 18 / effect-dependency violations** (`923e48cf`, `1224e84d`): resolved `react-hooks/exhaustive-deps` warnings that were previously suppressed with FIXME comments, which had been masking some of the above.

## Test coverage added

- `useFarmhandNetwork.test.ts` — unit coverage for the network hook, including a real (non-mocked) `COW_TRADE_TIMEOUT` firing.
- `e2e/tests/multiplayer/cow-trade-races.test.ts`, `online-toggle-race.test.ts`, `room-fallback.test.ts` — e2e regression coverage for the races above.
- `e2e/tests/save-data/cow-sale-persistence-race.test.ts` — e2e coverage for the persistence race.
- `src/shell/achievement-concurrency.test.ts`, `src/shell/unmount-safety.test.ts` — coverage for the achievement-update effect and unmount cleanup effect.

## Test plan
- [ ] `npm test` (vitest unit/integration suite)
- [ ] `npm run test:e2e` (Playwright e2e suite, including the new multiplayer race tests)
- [ ] Manual smoke test: go online/offline, sell/buy livestock, advance day, verify no console errors


jeremyckahn:
Some Claude-generated regression testing notes, pasting here for posterity:

---

## 0\. How to drive this

```
npm run start        # single client, http://localhost:3000
npm run dev          # client + api + backend + tracker, for full multiplayer testing
```

Open the browser console and confirm `window.farmhand` is defined (it's a proxy object exposing live state — used for debugging, same as the old `window.farmhand = this` in the class).

* * *

## 1\. HIGH-RISK: previously-regressed areas (test these first and hardest)

These map 1:1 to bugs introduced and then fixed mid-refactor. Confirm each is actually fixed, not just superficially working:

1.  **Online-toggle race** — Click "Play Online" from the home/settings screen. Confirm:
    
    - You actually land online (URL becomes `/online/<room>`, UI shows online state) and *stay* online — it should not flash "Connecting…" and then silently revert to offline.
    - No spurious "disconnected from server" notification appears immediately after toggling on.
    - Toggle online → offline → online rapidly (as fast as you can click) and confirm state doesn't get stuck or desync from the URL.
2.  **Peer-room dropped after ~10s** — Go online with two clients/tabs in the same room. Wait **at least 15–20 seconds** (longer than one heartbeat tick) and confirm both peers are *still* connected (peer list/count doesn't drop to zero, chat/trading still works). This bug caused the room to be silently left on every heartbeat (~10s), so a quick check right after connecting would have missed it.
    
3.  **Room fallback when URL omits room param** — Navigate to a URL that doesn't specify a room (or use in-app navigation that omits it) while already in a custom room. Confirm the app falls back to the *current* room in state, not `DEFAULT_ROOM`.
    
4.  **Achievement state loss** — Trigger an achievement (e.g. cross a money/harvest threshold) at the same time other rapid state updates are happening (e.g. right after a purchase or sale). Confirm the achievement notification fires and other state changes from the same tick aren't dropped.
    
5.  **Cow-trade stale closure** — With two online players, offer a cow for trade and request a peer's cow (`tradeForPeerCow`). Confirm the request uses current inventory/peer state, not a stale snapshot (test right after switching which cow is offered for trade, then immediately requesting a trade).
    
6.  **Cow-trade timeout** — Initiate a peer cow trade and let it time out (don't respond from the other peer). Confirm the "trade unavailable" notification appears and `isAwaitingCowTradeRequest` clears (UI unblocks) rather than hanging forever.
    
7.  **Combine harvester run-once-on-boot** — With a save that has the combine purchased (`isCombineEnabled`) and pending day notifications, reload the page. Confirm the field is harvested automatically **exactly once** on boot (not once per queued notification, not zero times if there happen to be no queued notifications — this is an intentional documented behavior change from the class version, verify it doesn't over- or under-harvest).
    
8.  **Combine harvester after ending the day** — With combine purchased, click "End the day" while focused on the Field view. Confirm the 1-second animation delay still happens before the auto-harvest, and the harvest happens exactly once (not per-notification).
    
9.  **Day-persistence race / cow-selling regression** — Sell a cow, then immediately end the day (or end the day and reload quickly). Confirm the sale is reflected in the persisted state after reload (localForage save actually captures the latest state, not a stale pre-sale snapshot).
    
10. **Unmount during increment-day** — Start "End the day" (which awaits a server call), then quickly navigate away (e.g. via browser back button or route change) before it resolves. Confirm no console errors about setting state on an unmounted component, and no notification fires for a screen that's gone.
    

* * *

## 2\. Boot / persistence

- [ ] **Fresh browser / no localForage data**: loads as a brand-new game — starting loan notification ("increased loan" / starting balance) shows, `dayCount` starts appropriately, starting inventory (scarecrow) present.
- [ ] **Existing saved game**: reload the page — game state (money, inventory, field, day count, achievements, cows, etc.) is restored exactly as it was.
- [ ] **Queued `newDayNotifications` on boot**: save state with pending day-end notifications (e.g. via ending a day then reloading before they clear), reload, and confirm each notification is shown (not swallowed, not duplicated) — the old code deferred these with `setTimeout(...,0)` per item; confirm ordering/display still looks right with several stacked notifications.
- [ ] **`isCombineEnabled` + boot**: see item 7/8 above.
- [ ] Confirm `window.farmhand` getter always reflects the *latest* render's state (open console, check a few fields, take an action, re-check — should not be stale).
- [ ] **Clearing persisted data** (dev-mode `shift+d` or menu action): confirms data wiped and "data deleted" notification shows, next reload starts a new game.

## 3\. Core single-player gameplay loop (regression sanity — should be unaffected, but confirm)

- [ ] Plant, water, harvest, and clear (weeds) on field plots using each tool (hoe/scythe/watering can/shovel) via both toolbar clicks and keyboard shortcuts.
- [ ] Buy/sell items in the shop; inventory and money update correctly.
- [ ] Selling reduces `todaysLosses`/increases revenue appropriately; buying decreases money (`handleAddMoneyClick`\-style debug actions still work if exposed in dev tools).
- [ ] Loan balance / loan payments still function.
- [ ] Crafting/recipes: learn a new recipe and confirm the "recipe learned" (singular) and "recipes learned" (plural, ≥2 in one tick) notifications both trigger correctly depending on how many are learned in the same state transition.
- [ ] Inventory-full notification: fill inventory to capacity and confirm the "inventory full" warning fires exactly once when crossing from space-remaining to full (not repeatedly every render).
- [ ] Cellar (aging/fermenting) purchase and produce workflow, if applicable to current save/level.
- [ ] Forest view unlocks at appropriate level/experience and appears in the view list only when unlocked and the FOREST feature flag is on.
- [ ] Cow Pen: buying pen, buying/selling cows, hugging cows (huggingMachine flag reset works), cow color purchases.
- [ ] Cow breeding: move two cows into the breeding pen, end the day through the gestation period, confirm a calf is born and both parent cows return to the pen. (E2E test for this changed — re-verify manually since the test no longer manually clicks "Breed" before ending days; confirm placing cows in the breeding pen via the UI still works as expected.)
- [ ] Breeding-pen checkbox now has an `aria-label="move cow to breeding pen"` — confirm the checkbox still visually toggles cow breeding-pen membership correctly (this was an accessibility-only change, verify no functional regression).

## 4\. Day advancement ("End the day" / shift+c)

- [ ] Click the "End day" FAB and confirm: input blocks during the transition (`isWaitingForDayToCompleteIncrementing`), day count increments, money/inventory/field state transitions correctly, "progress saved" notification shows (but NOT on the very first day).
- [ ] Repeated/rapid presses of shift+c or the End Day button while already incrementing should be a no-op (guarded by `isWaitingForDayToCompleteIncrementing`) — should not double-advance or throw.
- [ ] Online + end day: server round-trip for market positions happens, "positions posted" notification shows if you bought/sold that day, and it's also broadcast to peers (peer sees a message about your positions).
- [ ] Offline + end day: no server call attempted, no online-only notifications appear, day still advances locally.
- [ ] Server error during end-of-day POST (simulate by going offline mid-request, or check that dev/test env can trigger a 500): confirm "server error" notification, redirect to `/` (booted offline), `cowIdOfferedForTrade` resets, and the day still eventually completes without leaving the UI stuck blocked.
- [ ] Price crashes/surges (`priceCrashes`/`priceSurges`) still apply the ±50% value adjustment overrides correctly after a day advances, both online (server-provided adjustments) and offline (locally computed).

## 5\. Navigation & UI chrome

- [ ] Left/Right FAB buttons and left/right arrow keys cycle through `viewList` correctly, wrapping at both ends.
- [ ] Number keys 1–9 jump directly to the corresponding view in `viewList`, and no-op gracefully if that index doesn't exist (fewer than 9 views).
- [ ] Arrow-key/number-key navigation is suppressed while a tab element has focus (per `document.activeElement?.getAttribute('role') === 'tab'` guard) — click on a tab, then try arrow keys, confirm it doesn't also flip the stage focus.
- [ ] Menu (hamburger) FAB and `m` key toggle the sidebar Drawer open/closed; icon/class state (`is-open`) matches.
- [ ] Changing `stageFocus` while the menu is open and `doesMenuObstructStage()` is true auto-closes the menu (was `componentDidUpdate` logic, now a `useEffect`) — trigger a view change on a small viewport/mobile size and confirm the menu auto-collapses if it would obstruct the new stage.
- [ ] `MobileStepper` dot indicator's `activeStep` matches the current `stageFocus`'s index in `viewList`.
- [ ] Dialogs open/close correctly via keyboard shortcuts: `b` accounting, `a` achievements, `l` farmer's log, `e` price events, `s` stats, `,` settings, `shift+?` keybindings. Confirm each opens the right `dialogView` and that `closeDialogView` (X button / escape, whatever the dialog uses) closes it.
- [ ] Clicking the notification indicator icon opens the farmer's log (`FARMERS_LOG` dialog).
- [ ] All hotkeys and Fab click handlers are **disabled** while `isInputBlocked` is true (during network requests, cow trade waiting, or day-increment waiting) — confirm keyboard shortcuts and click handlers genuinely do nothing during these windows (not just visually disabled/no-op but actually blocked), and that they resume working the instant the block clears.
- [ ] `has-booted` / `block-input` / `menu-open`/`menu-closed` / alternate-end-day-button-position CSS classes on the root div all reflect state correctly (visual smoke test).

## 6\. Multiplayer / online (needs `npm run dev` + 2 browser tabs/profiles)

- [ ] Going online triggers "connecting to server" then "connected to room" notifications, in order, exactly once per connection attempt.
- [ ] `activePlayers` count reflects the number of connected peers and updates when peers join/leave (`onPeerJoin`/`onPeerLeave`).
- [ ] Heartbeat: confirm money accrues (`moneyTotal(money, activePlayers)`) on each heartbeat tick while online, and that this **stops** immediately after going offline (no more heartbeat fires after leaving).
- [ ] Switching rooms while online (e.g. via room-change UI) triggers a resync: notification sequence repeats, peer list resets and repopulates for the new room, no leftover peers from the old room.
- [ ] Peer metadata (position, name, cow colors, etc.) is sent/received continuously and throttled (5s window) rather than spamming — confirm peers see your position update but not on every single state change.
- [ ] Going offline from online: "disconnected from server" notification, peer room actually leaves (verify from the other peer's perspective that you disappear), heartbeat clears, `activePlayers`/`peerRoom` reset to null.
- [ ] Chat: `ChatRoom` only renders when `isChatAvailable` (online AND not in the default room) — confirm it appears/disappears correctly as you go online/offline and change rooms.
- [ ] Peer cow trade accept/reject flow end-to-end (offer, peer accepts; offer, peer rejects; offer times out) all leave state consistent (`isAwaitingCowTradeRequest` clears in all three cases).
- [ ] Attempting to trade for a cow you already own, or for a cow whose owner can't be resolved (e.g. peer just left), shows the correct error notification and does not send a bogus trade request.
- [ ] Reconnect/error handling: force a `getMarketData` failure (e.g. kill the API server while connecting) and confirm you're redirected to `/` (booted offline) with a "server error" notification, rather than getting stuck "Connecting…".

## 7\. Redirect / routing edge cases

- [ ] Directly loading `/online/<room>` in the URL boots straight into online mode for that room.
- [ ] Directly loading `/` boots offline.
- [ ] Toggling online/offline via UI updates the URL (via `<Redirect>`) and that redirect clears itself afterward (shouldn't get stuck re-redirecting).
- [ ] Manually editing the room segment of the URL while online triggers a room change/resync (not just a silent state mismatch between URL and displayed room).

## 8\. Persistence correctness after refactor-specific changes

- [ ] Confirm `reduceByPersistedKeys` output written to localForage after ending a day matches expectations (spot check a couple of fields in devtools' IndexedDB/localForage inspector).
- [ ] First day (`isFirstDay`) should **not** persist `historicalValueAdjustments` reset as part of the "progress saved" flow noise — i.e., first-day boot shouldn't show a "progress saved" toast, only subsequent day-ends should.
- [ ] Import/load a legacy fixture/save (`transformStateDataForImport`) and confirm nothing breaks compared to pre-refactor behavior (use existing e2e fixtures under `e2e/fixtures/` as real save files to load manually, e.g. `crops-mature.json`).

## 9\. Regression checks against automated test changes

The e2e suite changed for `cow-breeding`, `cow-pen`, and `go-online`. Since these were adjusted to pass against the refactored component, manually re-walk each scenario once outside of Playwright to make sure the *product* behavior is right and the tests weren't just adjusted to paper over a UI change:

- [ ] Cow breeding full flow (buy pen → buy/breed cows → advance 3 days → calf born).
- [ ] Cow pen purchase → buy cow → hug (animation class toggles) → sell.
- [ ] Basic "go online" smoke flow.

## 10\. Dev-mode-only shortcuts (only testable running a dev build, `import.meta.env.MODE === 'development'`)

- [ ] `shift+d` clears persisted data (with confirmation/notification).
- [ ] `w` waters all plots instantly.

* * *

## Notes on why these areas matter

- The refactor moved all `componentDidUpdate` logic into multiple `useEffect`s split by concern (network/online sync, achievements+menu+money, notifications). Effects can run in different relative order or with different dependency-array staleness than the single synchronous `componentDidUpdate` did — anything that depended on ordering between these concerns is worth double-checking (e.g. does an achievement notification and a "disconnected" notification ever race or duplicate when both conditions trigger in the same tick?).
- `ui-events.tsx` handlers still run against a plain-object "instance proxy" (`instanceProxyRef.current`) rather than a real component instance, rebuilt every render — this is a deliberate compatibility shim so the handler file didn't need a rewrite. Anything that reads `this.state` synchronously right after calling `this.setState(...)` in a handler is a spot where stale-proxy bugs could hide; fast double-clicks / rapid-fire handler calls on buy/sell/trade actions are good stress tests.

PR #673, opened 2026-07-08
Fix release workflow failure: allow git-protocol deps under npm 12

jeremyckahn:
## Summary
- The [release workflow run](https://github.com/jeremyckahn/farmhand/actions/runs/28982128095/job/86003080590) failed on `npm ci` with `EALLOWGIT`, because npm 12.0.0 (published the same day) now defaults `allow-git` to `"none"`, blocking git-protocol dependency fetches.
- `@electron/rebuild` (devDependency) pulls in `@electron/node-gyp` via a `git+ssh` URL — this is the only git dependency in the tree, and it's transitive, so `allow-git=root` wouldn't cover it.
- Added `allow-git=all` to `.npmrc` so `npm ci` can resolve it regardless of which npm version CI installs.

## Test plan
- [x] Verified locally against actual npm 12.0.0 (via `npx npm@12`) that `npm ci` completes successfully with this `.npmrc` change, reproducing and resolving the exact CI failure.
- [x] Confirmed npm 10.9.2 (currently installed locally) tolerates the new `.npmrc` key without warning/error, so no regression for local dev or other CI jobs.
- [ ] Re-run the release workflow to confirm the fix in CI.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_01Sbodhp6t32yKatoek287iD

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,6 @@
 progress=false
 legacy-peer-deps=true
+# npm >=12 blocks git-protocol dependency fetches by default. @electron/rebuild
+# (a devDependency) pulls in @electron/node-gyp via a git+ssh URL, so allow it.
+# https://github.com/npm/cli/releases/tag/v12.0.0
+allow-git=all
--- a/.npmrc
+++ b/.npmrc
@@ -1,6 +1,7 @@
 progress=false
 legacy-peer-deps=true
-# npm >=12 blocks git-protocol dependency fetches by default. @electron/rebuild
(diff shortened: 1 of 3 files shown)


jeremyckahn:
Good catch — updated to `allow-git=root` in 73c08585 and explicitly declared `@electron/node-gyp` as a devDependency (pinned to the exact commit already in the lockfile) so only this known, trusted git dependency is permitted rather than any git dependency anywhere in the tree.

PR #710, opened 2026-07-19
Add an E2E test case for Rainbow Mulch mechanics

jeremyckahn:
Added an E2E test case to exercise the Rainbow Mulch mechanics. I included an appropriate test fixture to simplify the test setup by copying the existing mulch fixture. The new test applies Rainbow Mulch to a tree, verifies the item is consumed, and checks that the correct tooltip state applies.

---
*PR created automatically by Jules for task [17910047291748339874](https://jules.google.com/task/17910047291748339874) started by @jeremyckahn*

### DIFF
--- a/e2e/tests/forest.test.ts
+++ b/e2e/tests/forest.test.ts
@@ -258,20 +258,18 @@ test('should apply rainbow mulch to one tree and leave another unmulched, advanc
   await expect(page.getByRole('tooltip')).toContainText('Rainbow Mulched')
 
   // Advance the days so the rainbow mulched tree grows faster than the other
-  // FERTILIZER_BONUS is 0.5. So 1 day = 1.5 days of growth.
-  // Apple sapling stages are [5, 5, 5, 5, 5]. It takes 25 days to fully grow.
-  // The trees in the fixture are 3 days old. Let's advance 14 days.
-  // The unmulched tree will be 17 days old (17 < 25), so still Growing.
-  // The rainbow mulched tree will be 3 + (14 * 1.5) = 24 days grown (24 < 25), still Growing...
-  // Wait, 15 days: unmulched = 18 days, mulched = 3 + 22.5 = 25.5 (Grown!).
+  // The trees in the fixture are 3 days old.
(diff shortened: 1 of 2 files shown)


e2e/tests/forest.test.ts
```diff
@@ -235,6 +235,28 @@ test('should purchase a Wood Chipper and craft the Wood Chips -> Mulch recipe ch
   ).toBeVisible()
 })
 
+test('should apply rainbow mulch to a tree and consume it from inventory', async ({
```
jeremyckahn: @jules Adjust this test to plant two trees: One with Rainbow Mulch, and one with no mulch. The days should be advanced to the point where the Rainbow Mulched tree is fully grown, at which point there should be expectations to assert that the Rainbow Mulched tree is grown and the non-mulched tree is not.

e2e/tests/forest.test.ts
```diff
@@ -235,6 +235,57 @@ test('should purchase a Wood Chipper and craft the Wood Chips -> Mulch recipe ch
   ).toBeVisible()
 })
 
+test('should apply rainbow mulch to one tree and leave another unmulched, advancing time to assert growth differences', async ({
```
jeremyckahn, line 238: @jules Let's iterate on this test a bit more. The test should instead have two trees, one with Rainbow Mulch and one with regular Mulch. The test should bring them both to maturity and then to the point of fruiting. The Rainbow Mulched tree should spawn fruit more quickly than the regular Mulched tree, so the expectations should assert and prove that.

jeremyckahn: @jules this isn't quite right. The second tree is not Mulched at all. Update the fixture to include both a Mulch and Rainbow Mulch. The first tree should be Rainbow Mulched, and the second tree should regular Mulched. The days should then be advanced to the point where the first tree has the "Ready to pick!" tooltip message. The regular Mulched tree should still be "Fruiting" at that point.

jeremyckahn:
Superseded by https://github.com/jeremyckahn/farmhand/pull/685/changes/14aa67598af106914840ed64b448220395074494.

PR #711, opened 2026-07-19
Forest refinements: Picker Pole, Toolbelt filtering, and layout fixes

lstebner:
### What this PR does

A batch of Forest-feature refinements:

- Adds the **Picker Pole** tool for harvesting forest fruit — unlocks alongside the Axe and the Forest itself at level 15. Picking fruit now requires the pole to be the actively selected tool, mirroring how the Axe/CHOP mode gates chopping.
- Adds **Picker Pole upgrade tiers** through the Forge: each tier (Bronze/Iron/Silver/Gold) picks one more fruit per harvest than the last (Basic 1 → Gold 5), with real per-tier art and Forest plot tooltip text showing the current tier's yield.
- Filters the **Toolbelt** to only show field-relevant tools (Watering Can, Scythe, Hoe, Shovel) on the Field screen and forest-relevant tools (Axe, Picker Pole) on the Forest screen.
- Moves **Saplings** into their own Shop tab instead of being grouped with crop seeds, and hides that tab until the Forest itself is unlocked (there's nowhere to plant a sapling before then, unlike Field's seeds which are always plantable).
- Fixes the level-up notification silently dropping rewards when a single level grants more than one (level 15 grants a shop item, tools, and Forest access at once — only the shop item message ever showed).
- Fixes `"Expand Forest"` casing to match the other Shop upgrade titles.
- Shows tree age in the Forest plot tooltip.
- Adds a new-day notification when forest fruit becomes ready to pick, mirroring the cow milking notification.
- Highlights choppable trees yellow (was red) and harvestable trees green while the matching tool is selected.
- Fixes forest fruit not earning experience when sold, unlike crops.
- Staggers alternating Forest rows so a tree's overhanging canopy lands in the gap between two plots above it instead of colliding with one, with a small deliberate overlap on both sides for a denser look. Also fixes the grid sitting lower than intended, and makes the whole tree/fruit sprite clickable (not just the plot square) so players can harvest by clicking the canopy directly.
- Fixes Forest plot sizing on narrow screens: plot size, the gap between plots, and the side padding now all shrink together instead of the gap staying oversized while plots got squeezed down; plots also stop shrinking below a legible minimum size and the grid scrolls horizontally instead.
- Fixes the Forest grid centering across the full viewport width without accounting for the fixed right-side Toolbelt sidebar in landscape mode, which ran the grid into it.
- Moves the Forest grid up on narrow (portrait) screens, where it previously sat noticeably lower than the space actually available above the bottom Toolbelt bar.
- Fixes selecting a sapling right after chopping or mulching a tree silently failing to plant — `fieldMode` stayed on CHOP/FERTILIZE, and `handleForestPlotClick`'s early-return branches for those modes never fell through to the planting logic.
- Fixes a selected sapling staying visually selected after a Toolbelt tool is clicked, which doesn't happen in the Field.
- Renames `apple-tree-wood.png` to `applewood.png`.

Also includes one fix that isn't Forest-specific:

- Fixes the app's left sidebar only updating whether it should be shown/hidden at boot or when navigating between screens — resizing the browser window (or rotating a device) never opened/closed it otherwise, which in practice required a full page reload. Affects every screen, not just Field/Forest.

- `npm run lint`, `npm run check:types`, and `npx vitest run` all pass.
- Run the app with `?enable_FOREST=true`:
  - Confirm the Toolbelt shows different tools on the Field vs. Forest screens.
  - Confirm Saplings has its own Shop tab, hidden until the Forest is unlocked.
  - Level up to 15 and confirm the notification mentions both the Apple Sapling and the Forest becoming available.
  - Plant and grow a few trees; confirm alternating rows are staggered with a slight, even overlap on both sides, clicking a canopy harvests fruit (only with the Picker Pole selected), chopping highlights yellow, harvesting highlights green.
  - Forge the Picker Pole up through Gold and confirm each tier picks more fruit and shows the right art/tooltip count.
  - Sell a harvested fruit and confirm experience is awarded.
  - Chop down a tree, then select a sapling without touching another tool first, and confirm it plants in the now-empty plot.
  - At a narrow browser width, confirm plots/gaps shrink together (not lopsided), the grid doesn't run into the sidebar Toolbelt in landscape mode, and the grid sits higher up in portrait mode.
  - Resize the browser window across the sidebar's breakpoint (no reload) on any screen and confirm the sidebar shows/hides accordingly in both directions.

None known.

### Additional information

Screenshot from 2026-07-19 10-10-59
image
image

https://github.com/user-attachments/assets/7fee33ae-894f-49e8-97a1-80921ceb54fd


lstebner:
@claude give this a review

lstebner:
@jeremyckahn this PR is addressing feedback you gave while playtesting the previous PRs. let me know if i missed anything!

lstebner:
@claude review again please

jeremyckahn (approved):
> Highlights choppable trees yellow (was red) and harvestable trees green while the matching tool is selected.

This could be worth adding an e2e test for. I don't feel super strongly about this, so I'll leave it up to your judgment @lstebner.

> Fixes the app's left sidebar only updating whether it should be shown/hidden at boot or when navigating between screens — resizing the browser window (or rotating a device) never opened/closed it otherwise, which in practice required a full page reload. Affects every screen, not just Field/Forest.

This is a really nice touch. Thanks for adding this!

lstebner:
> > Highlights choppable trees yellow (was red) and harvestable trees green while the matching tool is selected.
> 
> This could be worth adding an e2e test for. I don't feel super strongly about this, so I'll leave it up to your judgment @lstebner.

:+1: added
 
> > Fixes the app's left sidebar only updating whether it should be shown/hidden at boot or when navigating between screens — resizing the browser window (or rotating a device) never opened/closed it otherwise, which in practice required a full page reload. Affects every screen, not just Field/Forest.
> 
> This is a really nice touch. Thanks for adding this!

no problem! :bow:

PR #714, opened 2026-07-20
feat: Add four Forest-related achievements

lstebner:
### What this PR does

Adds four achievements tied to the Forest feature:

- **Orchardist** — pick 1000 fruits from trees (reward: 50 Apple Saplings)
- **Piemaker** — make 100 pies (reward: $100,000)
- **Deforestation** — chop down 250 trees (reward: 500 Wood)
- **Landscaper** — spread 100 bags of mulch, any type (reward: 25 Rainbow Mulch)

These required adding four new cumulative counters to `state` (`treesChopped`, `treeFruitsHarvested`, `recipesMade`, `mulchApplied`), following the existing `cropsHarvested`/`itemsSold` pattern, incremented in `chopForestPlot`, `harvestForestPlot`, `makeRecipe`, and `mulchForestPlot` respectively.

Also includes a fix for `treeFruitsHarvested`: it was originally incrementing by a flat `1` per harvest action, but `develop`'s Picker Pole tool tiers (merged after this branch was cut) let a single harvest yield up to 5 fruit at once. It now counts the actual fruit yield so Orchardist tracks real fruit picked, not harvest actions.

Reward amounts were bumped up from the initial pass since all four are late-game achievements with steep goals.

- `npx vitest run` — full unit suite passes, including new tests in `achievements.test.ts`, `chopForestPlot.test.ts`, `harvestForestPlot.test.ts`, `makeRecipe.test.ts`, and `mulchForestPlot.test.ts`.
- `npm run check:types` and `npm run lint` both pass.
- Manually: use the `window.farmhand.setState(...)` debug hook to push one of the new counters (e.g. `treesChopped: 250`) close to/over its goal and confirm the achievement toast fires and shows as completed in the Achievements view.

None currently.


lstebner:
@claude review please

src/components/Farmhand/useFarmhand.ts
jeremyckahn, line 1: Thank you for backfilling these!

src/data/achievements.ts
```diff
@@ -35,6 +35,25 @@ const sumOfCropsHarvested = memoize(
...
+const sumOfPartialRecordValues = memoize(
+  (record: Partial<Record<string, number>>) =>
+    Object.values(record).reduce(
+      (sum: number, value: number | undefined) => sum + (value || 0),
+      0
+    ),
+  {}
+)
+
+const PIE_RECIPE_IDS = [
```
jeremyckahn: @claude Rather than maintaining a separate list of pie recipes, would it make sense to extend the `recipe` type with an optional `isPie` boolean?

https://github.com/jeremyckahn/farmhand/blob/08f5128f5ac35a2c8be5847db9a107bf2a076e4e/src/farmhand.d.ts#L264-L268

Something like:

```ts
interface recipe extends item {
  recipeType: recipeType
  ingredients: Record<string, number>
  condition: recipeCondition
  isPie?: boolean
}
```

From there, `PIE_RECIPE_IDS` could be computed.

jeremyckahn (changes requested):
@lstebner, Claude called this out in its review:

> **`treeFruitsHarvested` can over-count relative to actual inventory gain when space is nearly full.** In `harvestForestPlot.ts:26-46`, `doesInventorySpaceRemain` only checks that *any* space remains (>0), not that all of `fruitYield` (up to 5 with a Gold Picker Pole) fits. `addItemToInventory` silently caps what it actually adds to available space, but `treeFruitsHarvested` is still incremented by the full `fruitYield`. This means with 1 inventory slot free and a Gold Picker Pole, the player receives 1 fruit but the Orchardist counter advances by 5. That said, this exactly mirrors the pre-existing `cropsHarvested` pattern in `harvestPlot.ts` (same `doesInventorySpaceRemain` + unconditional `harvestedQuantity` increment), so it's a consistent, intentional tradeoff in this codebase rather than a regression introduced by this PR — flagging for awareness only, not requesting a change.

Is this correct? If so it seems worth addressing as it may feel like a weird bug to players. It seems at least worth investigating before merging. I think we can leave the preexisting `cropsHarvested` issue in place for now, we can just add an informative TODO so I can hand it off to an agent later.

This is the only potentially blocking issue, IMO.

jeremyckahn:
> @lstebner, Claude called this out in its review:
> 
> > **`treeFruitsHarvested` can over-count relative to actual inventory gain when space is nearly full.** In `harvestForestPlot.ts:26-46`, `doesInventorySpaceRemain` only checks that _any_ space remains (>0), not that all of `fruitYield` (up to 5 with a Gold Picker Pole) fits. `addItemToInventory` silently caps what it actually adds to available space, but `treeFruitsHarvested` is still incremented by the full `fruitYield`. This means with 1 inventory slot free and a Gold Picker Pole, the player receives 1 fruit but the Orchardist counter advances by 5. That said, this exactly mirrors the pre-existing `cropsHarvested` pattern in `harvestPlot.ts` (same `doesInventorySpaceRemain` + unconditional `harvestedQuantity` increment), so it's a consistent, intentional tradeoff in this codebase rather than a regression introduced by this PR — flagging for awareness only, not requesting a change.
> 
> Is this correct? If so it seems worth addressing as it may feel like a weird bug to players. It seems at least worth investigating before merging. I think we can leave the preexisting `cropsHarvested` issue in place for now, we can just add an informative TODO so I can hand it off to an agent later.
> 
> This is the only potentially blocking issue, IMO.

@claude address this issue.

jeremyckahn (approved):
LGTM! 👏

PR #720, opened 2026-07-21
test: Skip animations in E2E tests to fix fake-clock race condition flakiness

jeremyckahn:
Fixes a race condition in E2E tests where Playwright assertions read animated numbers while they are still tweening. By skipping the animation entirely in the E2E environment using a new `SKIP_ANIMATIONS` feature flag, we make the tests deterministic without relying on complex, fragile Playwright fake clock choreography.

1. `SKIP_ANIMATIONS` feature flag added (implicitly via `VITE_ENABLE_SKIP_ANIMATIONS`).
2. `AnimatedNumber` component skips tweening and sets value synchronously when the flag is active.
3. `package.json`'s `start:e2e` updated to inject the flag.
4. Test files (`price-fluctuation.test.ts`) assert final values without pausing the clock.
5. `AnimatedNumber.test.tsx` coverage added for the new sync execution path.

---
*PR created automatically by Jules for task [5003453968568563097](https://jules.google.com/task/5003453968568563097) started by @jeremyckahn*

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -2,6 +2,6 @@ progress=false
 legacy-peer-deps=true
 # npm >=12 blocks git-protocol dependency fetches by default. @electron/node-gyp
 # is declared as an explicit devDependency (it's a transitive dependency of
-# @electron/rebuild via a git+ssh URL) so that allow-git=root permits it.
+# @electron/rebuild via a git+ssh URL) so that allow-git=all permits it.
 # https://github.com/npm/cli/releases/tag/v12.0.0
-allow-git=root
+allow-git=all
--- a/.npmrc
+++ b/.npmrc
@@ -2,6 +2,6 @@ progress=false
 legacy-peer-deps=true
 # npm >=12 blocks git-protocol dependency fetches by default. @electron/node-gyp
 # is declared as an explicit devDependency (it's a transitive dependency of
-# @electron/rebuild via a git+ssh URL) so that allow-git=all permits it.
+# @electron/rebuild via a git+ssh URL) so that allow-git=root permits it.
 # https://github.com/npm/cli/releases/tag/v12.0.0
-allow-git=all
+allow-git=root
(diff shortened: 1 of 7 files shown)


src/components/AnimatedNumber/AnimatedNumber.tsx
```diff
@@ -27,6 +29,11 @@ const AnimatedNumber = ({
 
   useEffect(() => {
     if (number !== previousNumber) {
+      if (features.SKIP_ANIMATIONS) {
```
jeremyckahn: @jules, @lstebner raised a better approach on Discord:

Rather than leverage the feature flag system, let's adjust `AnimatedNumber` to detect the user's `prefers-reduced-motion` (the CSS CSS media feature) setting and short circuit the animation that way. Then, update the E2E tests to set `prefers-reduced-motion` to `false`.

.npmrc
jeremyckahn, line 1: @jules why were the changes in this file made?

src/components/AnimatedNumber/AnimatedNumber.tsx
```diff
@@ -27,6 +27,11 @@ const AnimatedNumber = ({
 
   useEffect(() => {
     if (number !== previousNumber) {
+      if (window.matchMedia?.('(prefers-reduced-motion: reduce)').matches) {
```
jeremyckahn, line 30: @claude the code here are all seems correct, but I'm still seeing the numbers animate when motion is reduced via the Chrome devtools. I am also seeing the numbers animate when I monitor the E2E tests. What's going on here?

jeremyckahn: @claude Set aside the Chrome Devtools issues for now. Can you fix this PR such that `prefers-reduced-motion` is appropriately `false` in the E2E test environment, and that `AnimatedNumber` correctly evaluates it and appropriately short-circuits the animation when applicable?

PR #724, opened 2026-07-22
fix: Cap SearchBar width to match card list content

lstebner:
### What this PR does

The shared `SearchBar` component (used above the card list on Cellar, Inventory, Recipe list, and Cow Pen screens) only capped itself to `maxWidth: 100%` of its parent container. `ul.card-list > li` items are globally capped to `layout.cardMaxWidth` (550px) and centered. On any container wider than that (e.g. Cellar's 600px tab panel), the search input visibly overhung the cards below it.

Fixes the search input to use the same `layout.cardMaxWidth` token and centering (`margin: auto`) as the cards, so its edges line up with the content below it on every screen that reuses this component.

**Before:**
Search bar wider than cards

**After:**
Search bar aligned with cards

- `npx tsc --noEmit` and `npx vitest run src/components/SearchBar/SearchBar.test.tsx` both pass.
- Visually confirmed on the Cellar Inventory screen (screenshots above); the same `SearchBar` component is reused on Inventory, Recipe list (Fermentation/Winemaking/Vinegar Making), and Cow Pen screens, so the fix applies there too.

None.

### Additional information

none


jeremyckahn:
@lstebner you'll want to rebase this onto `develop` to resolve the test flakiness.

jeremyckahn (approved):
This looks much better! Good thinking, @lstebner!

PR #725, opened 2026-07-22
Fix flaky AnimatedNumber test

jeremyckahn:
Fixes a flaky test (`AnimatedNumber.test.tsx`) that caused occasional timeouts and errors ("Unable to find an element with the text: 0"). Replaced the brittle instantaneous DOM assertion and fake timer approach with `waitFor` and real timers to let the animation safely conclude.

---
*PR created automatically by Jules for task [5659193664609138478](https://jules.google.com/task/5659193664609138478) started by @jeremyckahn*

### DIFF
--- a/src/components/AnimatedNumber/AnimatedNumber.test.tsx
+++ b/src/components/AnimatedNumber/AnimatedNumber.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen } from '@testing-library/react'
+import { render, screen, waitFor } from '@testing-library/react'
 import { vi } from 'vitest'
 
 import AnimatedNumber from './AnimatedNumber.js'
@@ -48,7 +48,7 @@ test('skips animation and updates synchronously when prefers-reduced-motion is r
   window.matchMedia = originalMatchMedia
 })
 
-test('animates when prefers-reduced-motion is no-preference', () => {
+test('animates when prefers-reduced-motion is no-preference', async () => {
   const originalMatchMedia = window.matchMedia
 
   window.matchMedia = vi.fn().mockImplementation((query) => ({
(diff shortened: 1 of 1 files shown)


jeremyckahn:
@claude review

jeremyckahn:
> **Claude finished @jeremyckahn's task in 1m 47s** —— [View job](https://github.com/jeremyckahn/farmhand/actions/runs/29914122584)
> 
> ---
> ### Code Review
> 
> - [x] Gather context (PR diff, changed files)
> - [x] Read `AnimatedNumber` component and test file
> - [x] Review the diff for correctness, flakiness risk, best practices
> - [x] Post review feedback
> 
> This is a solid fix for the flakiness. Swapping the synchronous `expect(screen.getByText('0'))` assertion for `waitFor` + real timers removes the race against Shifty's `requestAnimationFrame`-driven tweening, and the reasoning in the added comments (`AnimatedNumber.test.tsx:74-76`) is accurate — jsdom doesn't guarantee rAF ticks align with a fixed schedule, so asserting an exact intermediate value was inherently brittle.
> 
> I ran the updated test suite for this file (`npx vitest run src/components/AnimatedNumber/AnimatedNumber.test.tsx`) and it passed in ~800ms, comfortably inside the new 2000ms `waitFor` timeout for the 750ms tween duration. `eslint` on the file is also clean.
> 
> A couple of minor, non-blocking notes:
> 
> 1. **Trailing semicolon** — `AnimatedNumber.test.tsx:80` has `..., { timeout: 2000 });` with a trailing `;`, but the project's `.prettierrc` sets `"semi": false` and no other line in this file (or most of the codebase) uses semicolons. Worth removing for consistency, though it won't fail lint since there's no stylistic semi rule wired into `eslint.config.js`.
> 2. **Weakened assertion** — the old check verified the displayed number was still `'0'` right after the re-render; the new one only checks it isn't `'100'` yet (`AnimatedNumber.test.tsx:77`). That's the right trade-off for eliminating flakiness, but it does mean the test no longer confirms the animation actually starts from the previous value — just that it doesn't jump immediately. Acceptable given the goal, just flagging the slightly reduced rigor.
> 3. **Cleanup on timeout** — if the `waitFor` on line 80 ever times out and throws, `window.matchMedia = originalMatchMedia` on line 82 won't run, leaking the mocked `matchMedia` into later tests in the file. This pattern predates this PR (same shape in the sibling test above), so it's not a regression introduced here, but wrapping the body in `try/finally` would make this test (and its neighbor) more robust long-term.
> 
> None of these block the fix — the core change correctly addresses the reported flakiness ("Unable to find an element with the text: 0").

@claude fix the issues you surfaced.

PR #727, opened 2026-07-23
fix: Prevent duplicate "Progress saved!" toasts from stacking on rapid End Day clicks

lstebner:
### What this PR does

Ending the day shows a "Progress saved!" toast that fades after a few seconds. Ending the day rapidly many times in a row caused these toasts to queue up and keep reappearing for many seconds after the player had stopped.

Root cause: `NotificationSystem` relied on notistack's `preventDuplicate` option, but it compared the message value passed to `enqueueSnackbar` by reference - and `showNotification` constructs a fresh object on every call, so identical consecutive notifications (e.g. "Progress saved!" fired on every end-of-day cycle) never actually deduped against each other.

Fix: give each notification a stable, content-derived key (`severity:message`) so `preventDuplicate` can actually match it against what's currently shown or queued. This generalizes to any notification message, not just "Progress saved!".

An earlier version of this fix used `persist: true` with a self-managed dismiss timer to also *extend* a visible toast's duration on a duplicate trigger. That approach caused a real regression - it filled notistack's 4-slot concurrent limit with persisted toasts from unrelated messages (loan interest, price events, etc. all fire from a single day-end), starving "Progress saved!" from reappearing until other toasts cooled down. It was reverted in favor of the simpler, native `preventDuplicate` + stable key approach, which lets notistack's own auto-hide behavior work as it always did for every other notification type.

That same earlier version also had a genuine infinite-recursion bug: its `onClose: () => closeSnackbar(key)` handler caused notistack's `closeSnackbar(key)` to invoke its own snack's `onClose` before removing it, recursing until a stack overflow the moment any toast's timer naturally elapsed. This is fixed by not passing an `onClose` handler at all - notistack fully manages a snack's lifecycle (auto-hide, then removal) once `autoHideDuration` and a `key` are set.

- `npx vitest run src/components/NotificationSystem/NotificationSystem.test.tsx` - unit tests covering the dedup key logic and a regression test asserting no `onClose` handler is passed (guards against the recursion bug reappearing).
- `npx playwright test e2e/tests/notifications.test.ts` - new e2e coverage: duplicate notifications don't stack, distinct notifications aren't over-deduped, and a notification clears after its display duration elapses without crashing the page (this test was verified to actually catch the recursion bug by temporarily reintroducing it).
- Manual: run `npm start`, end the day rapidly many times in a row - only one "Progress saved!" toast is ever visible at a time, and it clears out normally rather than lingering.
- Full regression pass: `npx vitest run` (full unit suite), `npm run lint`, `npm run check:types`, and a broad sweep of the e2e suite (all passing files unrelated to this change still pass; the only failures are a pre-existing, unrelated "Play online" checkbox/tracker-connectivity issue in this environment, confirmed to fail identically without this change).

None currently.

### Additional information

comparison videos.

#### before

notice how the "progress saved" notification repeats frequently (often while a previous one is still in view) and the animation pacing is quite distracting, almost impossible to read at times.

https://github.com/user-attachments/assets/c775fb15-4a8c-4dd5-b429-5d0daeeeeece

#### after

the "progress saved" notification no longer repeats itself while still in view, and the flow is much less chaotic feeling

https://github.com/user-attachments/assets/fb350f0d-5a01-4e6d-9532-dcd42db78b4d

### DIFF
--- a/e2e/tests/notifications.test.ts
+++ b/e2e/tests/notifications.test.ts
@@ -1,5 +1,6 @@
 import { expect, test } from '@playwright/test'
 
+import { NOTIFICATION_DURATION } from '../../src/constants.js'
 import { openPage } from '../test-utils/open-page.js'
 
 // window.farmhand is a real, supported debug hook (documented in
@@ -68,11 +69,11 @@ test('clears a notification after its display duration elapses without crashing 
   await showNotification(page, 'Progress saved!')
   await expect(page.getByText('Progress saved!')).toBeVisible()
 
-  // NOTIFICATION_DURATION (src/constants.ts) is 6000ms outside of unit
-  // tests. The page's clock is virtualized (see openPage), so this jumps
-  // the notification's own auto-hide timer forward instead of a real
-  // multi-second wait.
-  await page.clock.fastForward(7000)
+  // The page's clock is virtualized (see openPage), so this jumps the
+  // notification's own auto-hide timer forward instead of a real
+  // multi-second wait. The extra 1000ms is just a margin past the
+  // notification's actual configured duration.
+  await page.clock.fastForward(NOTIFICATION_DURATION + 1000)
 
   await expect(page.getByText('Progress saved!')).toBeHidden()
   expect(pageErrors).toEqual([])
--- a/e2e/tests/notifications.test.ts
+++ b/e2e/tests/notifications.test.ts
@@ -1,0 +1,89 @@
+import { expect, test } from '@playwright/test'
+
+import { openPage } from '../test-utils/open-page.js'
+
+// window.farmhand is a real, supported debug hook (documented in
+// README.md's "Debugging" section) but its ambient type declaration lives
+// in src/react-app-env.d.ts, which isn't part of this project's
+// TypeScript scope - so it's redeclared locally here with the shape this
+// file actually relies on.
+declare global {
+  interface Window {
+    farmhand?: {
+      setState: (partialState: Record<string, unknown>) => void
+    }
+  }
+}
+
+// These tests drive the notification system directly through the
+// window.farmhand.setState debug hook (documented in README.md's
+// "Debugging" section) rather than through repeated real End Day clicks.
+// That keeps them focused on NotificationSystem's own behavior and fast/
+// deterministic, without unrelated game-economy side effects (loan
+// interest, bankruptcy, price events, day-transition timing) that a real
+// end-of-day flow would introduce.
+const showNotification = (
+  page: import('@playwright/test').Page,
+  message: string,
+  severity: 'info' | 'success' | 'warning' | 'error' = 'info'
+) =>
+  page.evaluate(
+    ({ message, severity }) =>
+      window.farmhand?.setState({ latestNotification: { message, severity } }),
+    { message, severity }
+  )
+
+test('does not stack duplicate toasts when the same notification is triggered rapidly', async ({
+  page,
+}) => {
+  await openPage(page)
+
+  for (let i = 0; i < 5; i++) {
+    await showNotification(page, 'Progress saved!')
+  }
+
+  await expect(page.getByText('Progress saved!')).toHaveCount(1)
+})
+
+test('shows two different notifications concurrently rather than over-deduping', async ({
+  page,
+}) => {
+  await openPage(page)
+
+  await showNotification(page, 'Test notification A')
+  await showNotification(page, 'Test notification B')
+
+  await expect(page.getByText('Test notification A')).toBeVisible()
+  await expect(page.getByText('Test notification B')).toBeVisible()
+})
+
+test('clears a notification after its display duration elapses without crashing the page', async ({
+  page,
+}) => {
+  const pageErrors: string[] = []
+  page.on('pageerror', error => pageErrors.push(error.message))
+
+  await openPage(page)
+
+  await showNotification(page, 'Progress saved!')
+  await expect(page.getByText('Progress saved!')).toBeVisible()
+
+  // NOTIFICATION_DURATION (src/constants.ts) is 6000ms outside of unit
+  // tests. The page's clock is virtualized (see openPage), so this jumps
+  // the notification's own auto-hide timer forward instead of a real
+  // multi-second wait.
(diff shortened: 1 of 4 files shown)


lstebner:
@claude review this and look for any oversights or possible regressions

e2e/tests/notifications.test.ts
```diff
@@ -0,0 +1,89 @@
...
+// interest, bankruptcy, price events, day-transition timing) that a real
+// end-of-day flow would introduce.
+const showNotification = (
+  page: import('@playwright/test').Page,
+  message: string,
+  severity: 'info' | 'success' | 'warning' | 'error' = 'info'
+) =>
+  page.evaluate(
+    ({ message, severity }) =>
+      window.farmhand?.setState({ latestNotification: { message, severity } }),
```
jeremyckahn, line 34: @claude it seems like this is sidestepping the native `showNotification` logic and preventing this test from exercising the critical mechanism: https://github.com/jeremyckahn/farmhand/blob/53917e471e0b99a78863c37808bd9e00e7372d46/src/game-logic/reducers/showNotification.ts#L8-L32

Do I have that right? If so, can you adjust this test to properly leverage that code path so that this test fully validates the relevant logic?

e2e/tests/notifications.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  await openPage(page)
+
+  await showNotification(page, 'Progress saved!')
+  await expect(page.getByText('Progress saved!')).toBeVisible()
+
+  // NOTIFICATION_DURATION (src/constants.ts) is 6000ms outside of unit
+  // tests. The page's clock is virtualized (see openPage), so this jumps
+  // the notification's own auto-hide timer forward instead of a real
+  // multi-second wait.
+  await page.clock.fastForward(7000)
```
jeremyckahn: @claude can we use `NOTIFICATION_DURATION + 1000` (or something similar) here to make this a bit more durable in case the constant value changes?

src/components/NotificationSystem/NotificationSystem.test.tsx
```diff
@@ -35,52 +34,52 @@ test('calls enqueueSnackbar when latestNotification is provided', () => {
...
     <NotificationSystem
       {...defaultProps}
       enqueueSnackbar={enqueueSnackbar}
-      latestNotification={null}
+      latestNotification={latestNotification}
     />
   )
 
-  expect(enqueueSnackbar).not.toHaveBeenCalled()
+  expect(enqueueSnackbar.mock.calls[0][1]).not.toHaveProperty('onClose')
```
jeremyckahn: @claude What is this assertion buying us? Should we remove it?

src/components/NotificationSystem/NotificationSystem.tsx
```diff
@@ -7,6 +7,11 @@ import { withSnackbar } from 'notistack'
 import { NOTIFICATION_DURATION } from '../../constants.js'
 import FarmhandContext from '../Farmhand/Farmhand.context.js'
 
+const getNotificationKey = ({
+  message,
+  severity,
+}: farmhand.notification): string => `${severity}:${message}`
```
jeremyckahn, line 13: @claude would it be worth returning a template literal type here to make the notification logic more type-safe?

src/components/NotificationSystem/NotificationSystem.test.tsx
```diff
@@ -35,52 +34,52 @@ test('calls enqueueSnackbar when latestNotification is provided', () => {
   )
 
   expect(enqueueSnackbar).toHaveBeenCalledWith(latestNotification, {
+    key: 'info:Test notification',
```
jeremyckahn: @claude should we use `getNotificationKey` to calculate keys like this in the tests to make them a bit more robust?

jeremyckahn (changes requested):
This is an excellent improvement! Thanks @lstebner. I've got some questions for Claude that might help to make this all a bit more robust.

src/components/NotificationSystem/NotificationSystem.test.tsx
jeremyckahn: @claude let's do this:

> Delete the test entirely and fold its explanatory comment into the first test, relying on the existing exact-match assertions to guard the regression.

e2e/tests/notifications.test.ts
jeremyckahn: @claude I think your change here got dropped. I still see `window.farmhand?.setState({ latestNotification: { message, severity } }),`.

jeremyckahn: The `window.farmhand?.setState({ latestNotification: { message, severity } })` line you're seeing is still there, but it now only belongs to the `showNotification` helper used by the *other two* tests in this file (distinct-notifications and display-duration) — not the primary dedup test.

That primary test ("does not stack duplicate toasts...") was rewritten in 0040f987 to drive `endDay(page)` in a loop instead, which goes through the real `handleClickEndDayButton` → `incrementDay` → `showNotification.ts` reducer path. So it does exercise the actual object-identity churn that caused the bug — nothing got dropped there.

The other two tests intentionally keep using the debug hook (per the comment above the helper) since they're testing `NotificationSystem`'s generic dedup/lifecycle behavior, not this specific reducer bug, and using the hook keeps them fast and free of unrelated game-economy side effects. Let me know if you'd rather those be converted to real game actions too, or if this looks right as-is.

src/components/NotificationSystem/NotificationSystem.tsx
jeremyckahn: Tried applying this, but it's blocked by the toolchain: this repo pins `prettier@1.19.1` (2019), which predates TypeScript template literal types entirely (added in TS 4.1, 2020) and can't parse the syntax at all — the pre-commit `pretty-quick` hook crashes on `` `${farmhand.notificationSeverity}:${string}` ``. There are no existing template literal types anywhere else in this codebase either, so this would be the first and would need a prettier upgrade to land safely.

Dropping this for now per your call — happy to revisit if/when prettier gets upgraded separately.

jeremyckahn (approved):
LGTM! Thanks for finally fixing this, @lstebner!

PR #723, opened 2026-07-22
Add Apple Juice, Grape Syrup, and Vinegar-making to the Cellar

lstebner:
### What this PR does

Adds a small beverage/condiment recipe chain:

- **Apple Juice** and **Grape Syrup**: plain Kitchen recipes (Apple + Sugar; Grapes reduced into syrup).
- **Apple Cider Vinegar** and **Balsamic Vinegar**: a new Vinegar family that ferments in the Cellar (Apple Juice + Yeast; Grape Syrup + Yeast), mirroring Wine's aging mechanic - kegs that never spoil and appreciate in value over time - but generalized to arbitrary per-recipe ingredient maps instead of Wine's hardcoded flat grape/yeast constants. Surfaced via a new "Vinegar Making" Cellar tab.
- A **"kegs ready to sell" notification**, scoped to Wine and Vinegar only (generic fermented crops already have their own spoilage notification, and don't need this too).
- A new **floorboard background** for the Cellar view, replacing the generic green dot pattern, better fitting the wine/vinegar-aging theme.

### How this change can be validated

- `npx vitest run` - full unit suite passes (services, reducers, utils, and component-level coverage for the new Vinegar UI).
- `npx tsc --noEmit` and `npx eslint src --max-warnings=0` both clean.
- `npx playwright test e2e/tests/vinegar-making e2e/tests/winemaking` - end-to-end make -> age -> sell flow verified for both Vinegar and Wine (no regressions to Wine).
- Manually verified in the browser: Kitchen tab shows Apple Juice/Grape Syrup once unlocked, Cellar's new "Vinegar Making" tab lists both vinegar recipes, kegs age and become sellable, and the ready notification fires only for Wine/Vinegar (not generic fermented crops).

### Questions or concerns about this change

- Ingredient quantities and `daysToMature` values (10 Apple Juice + 5 Yeast -> 14 days; 24 Grape Syrup + 6 Yeast -> 21 days, etc.) are reasonable, but i'm open to adjustment of course.

### Additional information

(note: I took these screenshots before adjusting the recipe ingredients)
Screenshot from 2026-07-21 21-44-31
Screenshot from 2026-07-21 21-44-26
Screenshot from 2026-07-21 21-01-34
Screenshot from 2026-07-21 21-01-25
Screenshot from 2026-07-21 21-45-32
Screenshot from 2026-07-21 21-45-13

full screen shot to show new background idea:
image

[...]

PR #728, opened 2026-07-23
feat: Add Apple Juice, Grape Syrup, keg-ready notifications, and a Cellar background

lstebner:
### What this PR does

Adds a small set of independent Cellar/Kitchen improvements, split out from #723 so the Vinegar mechanic there can go through a redesign separately:

- **Apple Juice** and **Grape Syrup**: plain Kitchen recipes (Apple + Sugar; Grapes reduced into syrup). Grape Syrup has no consumer yet - it's being kept in anticipation of a reworked Vinegar (or similar) recipe that will use it once that design lands.
- A **"kegs ready to sell" notification**, scoped to Wine only (generic fermented crops already have their own spoilage notification, and don't need this too).
- A new **floorboard background** for the Cellar view, replacing the generic green dot pattern.

- `npx vitest run` - full unit suite passes.
- `npx tsc --noEmit` and `npx eslint src --max-warnings=0` both clean.
- `npx playwright test e2e/tests/winemaking` - Wine's make -> age -> sell flow verified, including the new ready notification, with no regressions.
- Manually verified in the browser: Kitchen tab shows Apple Juice/Grape Syrup once unlocked, and the Cellar's new floorboard background renders correctly.

- Grape Syrup is included with no current recipe consuming it - intentional, since it's meant to carry over into whatever the reworked Vinegar design ends up being, rather than re-deriving it later.

### Additional information

Split out of #723, which is being kept open for reference while Vinegar's design gets rethought.


jeremyckahn:
@claude review

src/img/ui/floorboard.png
jeremyckahn, line 1: This is a big improvement!

jeremyckahn (changes requested):
From Claude:

> #### Minor: test coverage gap (non-blocking)
> 
> `src/game-logic/reducers/processCellar.test.ts:60-71` — `'does not notify for a keg that is already ready or not yet ready'` uses `carrot`/`garlic`, which are non-wine items. Since `isNotifiableKeg` is `false` for those regardless of `daysUntilMature`, this test passes trivially and doesn't actually exercise the `keg.daysUntilMature > 0 && daysUntilMature <= 0` boundary for a _wine_ keg. A regression that broke that boundary check (e.g. re-notifying an already-mature wine keg every day, or notifying a couple days early) wouldn't be caught by the current suite.

This seems worth addressing, but once it is this seems good to merge.

@claude, can you implement the above?

jeremyckahn:
> #### Minor: test coverage gap (non-blocking)
> 
> `src/game-logic/reducers/processCellar.test.ts:60-71` — `'does not notify for a keg that is already ready or not yet ready'` uses `carrot`/`garlic`, which are non-wine items. Since `isNotifiableKeg` is `false` for those regardless of `daysUntilMature`, this test passes trivially and doesn't actually exercise the `keg.daysUntilMature > 0 && daysUntilMature <= 0` boundary for a _wine_ keg. A regression that broke that boundary check (e.g. re-notifying an already-mature wine keg every day, or notifying a couple days early) wouldn't be caught by the current suite.
> 
> Suggest adding two wine-specific cases alongside the existing ones:
> 
>     * a wine keg with `daysUntilMature` already `<= 0` before this tick → should not re-notify
> 
>     * a wine keg with `daysUntilMature` still several days out → should not notify yet

@claude implement these recommended changes.

jeremyckahn (approved):
LGTM!

PR #737, opened 2026-07-25
Add tappable navigation stepper icons

### LINKED ISSUE #66: Add tappable navigation stepper icons for each section

lstebner:
### What this PR does

Closes #66.

Replaces the bottom navigation's non-interactive dots with a row of tappable, tooltipped icon buttons (one per unlocked view, in order) that jump straight to that screen. The active view is highlighted (white background, blue outline); others are gray. Icons are plain unicode emoji for now (still deciding on final art - see discussion), chosen so swapping in different icons later is a small, isolated change (`STAGE_ICON_MAP` in `BottomControls.tsx`).

The URL-hash view/tab persistence that was originally part of this PR has been split out into #740 so it can be reviewed independently.

- `npx vitest run` / `npm run check:types` / `npm run lint`
- e2e: `npx playwright test e2e/tests/bottom-navigation.test.ts`
- Manually: navigate around via the new icon row (hover for tooltips) at a narrow viewport (~320px) with several views unlocked to confirm the icon row doesn't overflow.

Still deciding on the final icon art - currently using plain unicode emoji.

### Additional information

### DIFF
--- a/e2e/fixtures/bottom-navigation-all-views.json
+++ b/e2e/fixtures/bottom-navigation-all-views.json
@@ -1,0 +1,263 @@
+{
+  "allowCustomPeerCowNames": false,
+  "cellarInventory": [],
+  "completedAchievements": {
+    "plant-crop": true,
+    "water-crop": true
+  },
+  "cowBreedingPen": {
+    "cowId1": null,
+    "cowId2": null,
+    "daysUntilBirth": -1
+  },
+  "cowColorsPurchased": {},
+  "cowForSale": {
+    "baseWeight": 1601,
+    "color": "GREEN",
+    "colorsInBloodline": {
+      "GREEN": true
+    },
+    "daysOld": 1,
+    "daysSinceMilking": 0,
+    "daysSinceProducingFertilizer": 0,
+    "gender": "FEMALE",
+    "happiness": 0,
+    "happinessBoostsToday": 0,
+    "id": "b93a9c59-f73f-4986-b44e-ffb422bcd4c2",
+    "isBred": false,
+    "isUsingHuggingMachine": false,
+    "name": "Guava",
+    "ownerId": "",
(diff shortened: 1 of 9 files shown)


lstebner:
@claude give this a thorough review

jeremyckahn:
Nice work so far @lstebner! I'm onboard with the idea here, but there are a few things I think are worth addressing before merging:

- I think it's fine to have this enabled by default, but players should be able to disable this new UI via the settings in case they find it to clutter their screen too much.
- Given that Farmhand is playable on mobile, the implementation/design should support scenarios where the stepper icon strip is wider than the viewport. Perhaps we just make it scrollable for now?
- We should avoid using emojis as UI like the current version does. It's inconsistent with the rest of the UI and variable from platform to platform. I'm not sure if that's just a temporary placeholder, but I wanted to mention it regardless. IMO this warrants custom pixel art images for each of the stepper buttons to achieve a cohesive aesthetic. We can reuse images here where sensible (e.g. the Carrot graphic for the Field).
- Currently the stepper button backgrounds are white. Assuming that's a placeholder, let's change that to something more visually engaging (perhaps [MUI's `primary.light` color](https://mui.com/material-ui/customization/color/#official-color-tool) or similar)?
- We may need to adjust the scrollable areas in some screens to ensure all interactable areas (e.g. Field plots) are fully accessible.

jeremyckahn:
Note: this also needs to be made more compatible with the route changes for online multiplayer. Currently, when the player goes online, the route gets into a weird, inconsistent state.

lstebner:
i'm probably going to sit on this branch for awhile @jeremyckahn . you kind of touched on the visuals and also the space these take up in your comment, but i can't really find a way to do these that looks as good as i want it to look. i'm splitting out the navigation changes i made because i think those are useful on their own without this

PR #740, opened 2026-07-28
Persist current view and tab in the URL hash across reloads

lstebner:
### What this PR does

Split out of #737 so it can be reviewed/iterated on independently of the bottom-nav icon redesign.

- Persists the current view - and, for tabbed screens (Shop, Workshop, Cellar), the active tab - in the URL hash (e.g. `#online/room?view=SHOP&tab=Upgrades`), so a page reload (or a shared/bookmarked online-room link) restores where you were instead of always landing back on Home.
- Validates the hash's `view` against what's actually unlocked (via a new `getViewList` helper) before applying it, both on cold boot and when a save loads asynchronously - a bookmarked/shared URL pointing at a screen the current save hasn't unlocked falls back to Home rather than rendering it anyway.

- `npx vitest run` / `npm run check:types` / `npm run lint`
- e2e: `npx playwright test e2e/tests/hash-navigation.test.ts`
- Manually: navigate to Shop, switch to the Upgrades tab, then reload the page - you should land back on Shop with Upgrades still selected. Try bookmarking a URL with a locked view's hash (e.g. `#?view=FOREST` before Forest is unlocked) and confirm it falls back to Home instead of rendering Forest.

The `view`/`tab` URL sync uses raw `history.replaceState` (see `src/utils/hashQueryParams.ts`) rather than react-router's `history`/`location`, since `useTabQueryParam` is used by components that don't always render under a `<Router>` in tests. This also means it doesn't go through react-router's own history stack, so it won't interfere with back/forward navigation.

### Additional information

This PR's e2e tests navigate via the existing sidebar view-select dropdown rather than the bottom-nav icon buttons from #737, so it doesn't depend on that PR landing first.

https://github.com/user-attachments/assets/db79dfa8-86df-426a-8e2e-64e33a6d9cbc


jeremyckahn (changes requested):
Great work so far @lstebner! I'm onboard with the general idea here, but I think it needs to be made compatible with the browser navigation system for it to be intuitive for players. Right now the URL changes, but because we're using `replaceState` the browser's back button just takes the player back to the previous (non-Farmhand) page they were on (`about:blank` in the case of this video):

https://github.com/user-attachments/assets/77c70c4a-5ba2-43ae-8b29-c53e1831e6be

It seems that this was a conscious choice, but IMO it feels more broken than leveraging the browser history stack. In other cases, it feels very weird to use the browser back/forward buttons in the game after navigating around, see the URL change, but not have a predictable effect on the game's navigation state. What do you think @lstebner?

src/components/Farmhand/useFarmhand.ts
```diff
@@ -567,7 +624,21 @@ export const useFarmhand = (props: FarmhandProps) => {
...
+        // state rather than calling createInitialState() again - that
+        // would re-run generateCow()/uuid(), consuming extra draws from
+        // the seeded RNG sequence that other code (and e2e fixtures tied
+        // to specific seeds) depends on staying stable.
+        const restoredStageFocus = getValidatedStageFocusFromHash(
+          stateRef.current
+        )
+
+        setState(s => ({
+          ...s,
```
jeremyckahn: @claude nit: For convention's sake, let's rename `s` to something like `prev`:
```suggestion
        setState(prev => ({
          ...prev,
```

src/hooks/useTabQueryParam.ts
jeremyckahn, line 1: @claude As a matter of due diligence, is there a library we could/should instead here? Not a big deal if we need a bespoke solution, I just want to see if we can reduce our maintenance burden.

src/utils/getValidatedStageFocusFromHash.ts
```diff
@@ -0,0 +1,40 @@
...
+ * since Stage.tsx renders whichever view stageFocus names with no unlock
+ * check of its own. Used both at boot and when reacting to popstate.
+ */
+export const getValidatedStageFocusFromHash = (
+  candidateState: Pick<
+    farmhand.state,
+    'experience' | 'purchasedCellar' | 'purchasedCowPen' | 'showHomeScreen'
+  >
+): farmhand.stageFocusType | undefined => {
+  const stageFocusFromUrl = getHashQueryParams().get('view')
```
jeremyckahn: @claude `'view'` is a magic string. Let's move it to a named constant, here and anywhere else it's used.

src/utils/getValidatedStageFocusFromHash.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const availableViews = getViewList({
+    isForestUnlocked: isForestUnlockedForState,
+    purchasedCellar: candidateState.purchasedCellar,
+    purchasedCowPen: candidateState.purchasedCowPen,
+    showHomeScreen: candidateState.showHomeScreen,
+  })
+
+  return availableViews.includes(stageFocusFromUrl as farmhand.stageFocusType)
+    ? (stageFocusFromUrl as farmhand.stageFocusType)
```
jeremyckahn: @claude this `as` is a code smell for poor type safety. Use a guard or some similar construct to verify that `stageFocusFromUrl` is the expected type.

src/utils/hashQueryParams.ts
jeremyckahn, line 1: @claude similarly, is there an opportunity to use a third-party library instead of a bespoke solution here?

jeremyckahn (changes requested):
Great work @lstebner! This is an excellent addition.

Things are looking good functionally, but I'd like to see if we can make the implementation a bit more robust.

src/utils/hashQueryParams.ts
jeremyckahn: I think this is fine to leave for now.

src/hooks/useTabQueryParam.ts
jeremyckahn: We can save this for another day.

src/utils/getValidatedStageFocusFromHash.ts
jeremyckahn: @claude what happened with 61ea4ba47dd2069e87901055f4f13ee2365bed2f? Something went sideways with that commit. Revert and try again.

jeremyckahn (approved):

PR #741, opened 2026-07-28
chore: refactor compost recipes

lstebner:
### What this PR does

this PR rebalances compost and fertilizer recipes to make them more obtainable and useful.

unlock the composter and then check out the recipes.


jeremyckahn (approved):
Good call here, @lstebner!

Issue #81, opened 2021-05-30
add progress bar per achievement


lstebner:
![image](https://user-images.githubusercontent.com/628757/120114664-14b89280-c135-11eb-8a1a-e0a7d367c7eb.png)

for some of these long achievements, a progress bar would help indicate how far along the player is. there is already a progress bar component used in this view that should be reusable for this purpose. this might warrant some new data in the achievement model, we can discuss that before starting on this

jeremyckahn:
I think this would be a great addition! I think a reasonable way to approach this would be to start with adding to the `farmhand.achievement` data structure: https://github.com/jeremyckahn/farmhand/blob/1462ca1f755ba3ab320ff388b1917b4043550372/src/index.js#L122-L132

I think it would make sense to add an **optional** property named something like `progressComponent`. That property could be a React component that the [`Achievement` component](https://github.com/jeremyckahn/farmhand/blob/develop/src/components/Achievement/Achievement.js) renders if present. That passed-in component would then receive the game's `farmhand.state` object to compute and render the progress for its related achievement.

I'm not married to this approach, but this feels like the most straightforward one off the top of my head!

PR #742, opened 2026-08-01
feat: Add a progress bar for achievements requiring repeated actions

lstebner:
### What this PR does

Implements #81: achievements that require repeated actions (e.g. "Sell 10,000 units") now show a progress bar instead of a flat complete/incomplete state.

Rather than the issue's suggested `progressComponent` (a React component stored on the achievement data), each achievement gets an optional `getProgress(state)` function returning `{ currentValue, goal }`, reusing the `goal` it already has for its `condition`. This keeps `src/data/achievements.ts` as plain data (no React) and reuses the existing `ProgressBar` component. Added it to the ~20 achievements that are genuinely "do it N times"; one-shot/time-gated/exact-match achievements are unaffected.

Also tweaked `ProgressBar` itself (shared by these new bars and the existing aggregate bar): percent label now sits beside the bar instead of below it, and the empty-track color is a translucent overlay instead of a flat gray that clashed with the app's warm card backgrounds. This was a subject design change, open to feedback.

- `npx vitest run src/data/achievements.test.ts src/components/Achievement/Achievement.test.tsx src/components/ProgressBar/ProgressBar.test.tsx src/components/AchievementsView/AchievementsView.test.tsx`
- `npx tsc --noEmit`
- Manually: `window.farmhand.setState({ treeFruitsHarvested: { apple: 400 } })` in the browser console, then open Achievements (`a`) and confirm Orchardist shows a 40% bar.

Open to feedback on which achievements should/shouldn't get a bar — I skipped a few borderline cases like Financial Freedom (tracks a shrinking loan balance, not a growing counter).

Also wondering if we want to floor the % that we display. Showing things like "54.17%" seems needlessly specific.

### Additional information

image

Screenshot from 2026-07-31 21-51-10

### DIFF
--- a/src/components/Achievement/Achievement.test.tsx
+++ b/src/components/Achievement/Achievement.test.tsx
@@ -43,4 +43,70 @@ describe('<Achievement />', () => {
       screen.getByText(new RegExp(achievementObject.rewardDescription))
     ).toBeInTheDocument()
   })
+
+  test('does not render a progress bar when the achievement has no getProgress', () => {
+    expect(document.querySelector('.ProgressBar')).not.toBeInTheDocument()
+  })
+})
+
+describe('<Achievement /> with getProgress', () => {
+  let achievementObject: farmhand.achievement
+
+  beforeEach(() => {
+    achievementObject = {
+      description: 'the best achievement',
+      id: 'achievement-1',
+      name: 'achievement one',
+      rewardDescription: 'the greatest reward',
+      condition: () => false,
+      reward: state => state,
+      getProgress: () => ({ currentValue: 25, goal: 100 }),
+    }
+
+    const farmhandContextValue = createContextData()
+
+    render(
+      <FarmhandContext.Provider value={farmhandContextValue}>
+        <Achievement
+          achievement={achievementObject}
+          completedAchievements={{}}
+        />
+      </FarmhandContext.Provider>
+    )
+  })
+
+  test('renders a progress bar', () => {
+    expect(document.querySelector('.ProgressBar')).toBeInTheDocument()
+  })
+})
+
+describe('<Achievement /> with getProgress and isComplete', () => {
+  let achievementObject: farmhand.achievement
+
+  beforeEach(() => {
+    achievementObject = {
+      description: 'the best achievement',
+      id: 'achievement-1',
+      name: 'achievement one',
+      rewardDescription: 'the greatest reward',
+      condition: () => false,
+      reward: state => state,
+      getProgress: () => ({ currentValue: 100, goal: 100 }),
+    }
+
+    const farmhandContextValue = createContextData()
+
+    render(
+      <FarmhandContext.Provider value={farmhandContextValue}>
+        <Achievement
+          achievement={achievementObject}
+          completedAchievements={{ 'achievement-1': true }}
+        />
+      </FarmhandContext.Provider>
+    )
+  })
+
+  test('does not render a progress bar once the achievement is complete', () => {
+    expect(document.querySelector('.ProgressBar')).not.toBeInTheDocument()
+  })
 })
--- a/src/components/Achievement/Achievement.tsx
+++ b/src/components/Achievement/Achievement.tsx
@@ -25,6 +25,7 @@ const Achievement = ({
   isComplete?: boolean
 }) => {
   const { description, name, rewardDescription } = achievement
+  const progress = achievement.getProgress?.(gameState)
 
   return (
     <Card
@@ -44,15 +45,13 @@ const Achievement = ({
       />
       <CardContent>
         <p>{description}</p>
-        {!isComplete && achievement.getProgress && (
+        {!isComplete && progress && (
           <Div sx={{ marginTop: '1em' }}>
             <ProgressBar
               {...{
                 percent: Math.min(
                   100,
-                  (achievement.getProgress(gameState).currentValue /
-                    achievement.getProgress(gameState).goal) *
-                    100
+                  (progress.currentValue / progress.goal) * 100
                 ),
               }}
             />
--- a/src/components/Achievement/Achievement.tsx
+++ b/src/components/Achievement/Achievement.tsx
@@ -9,35 +9,59 @@ import BeenhereIcon from '@mui/icons-material/Beenhere.js'
 import { bool, object, shape, string } from 'prop-types'
 
 import FarmhandContext from '../Farmhand/Farmhand.context.js'
+import { Div } from '../Elements/index.js'
+import ProgressBar from '../ProgressBar/index.js'
 
 const Achievement = ({
-  achievement: { description, id, name, rewardDescription },
+  achievement,
   completedAchievements,
+  gameState,
 
-  isComplete = Boolean(completedAchievements[id]),
+  isComplete = Boolean(completedAchievements[achievement.id]),
 }: {
   achievement: farmhand.achievement
   completedAchievements: Partial<Record<string, boolean>>
+  gameState: farmhand.state
   isComplete?: boolean
-}) => (
-  <Card
-    {...{ className: classNames('Achievement', { 'is-complete': isComplete }) }}
-    sx={{
-      '& .MuiSvgIcon-root': { color: isComplete ? '#13b747' : '#666' },
-    }}
-  >
-    <CardHeader
+}) => {
+  const { description, name, rewardDescription } = achievement
+
+  return (
+    <Card
       {...{
-        avatar: isComplete ? <BeenhereIcon /> : <AssignmentLateIcon />,
-        title: name,
-        subheader: <p>Reward: {rewardDescription}</p>,
+        className: classNames('Achievement', { 'is-complete': isComplete }),
       }}
-    />
-    <CardContent>
-      <p>{description}</p>
-    </CardContent>
-  </Card>
-)
+      sx={{
+        '& .MuiSvgIcon-root': { color: isComplete ? '#13b747' : '#666' },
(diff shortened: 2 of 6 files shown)


lstebner:
@claude review

lstebner:
pushed an update for finding #1. the rest are not currently actionable, but awaiting human feedback :robot:

jeremyckahn (changes requested):
This is a great addition @lstebner! Thanks for adding this. I'm a little concerned that we're adding excessive duplicate logic to the Achievement definitions and I'd like to see if that can/should be DRYed up.

@claude What do you think? Is DRYing the Achievements worth doing here? If so, propose some strategies for doing this. Consider how we might add any worthwhile missing test coverage for per-Achievment progress calculation as well.

jeremyckahn:
> **Claude finished @jeremyckahn's task in 2m 5s** —— [View job](https://github.com/jeremyckahn/farmhand/actions/runs/30707161116)
> ### Review complete
> 
>     * [x]  Gather context (diff, changed files, CLAUDE.md — none present)
> 
>     * [x]  Review `src/data/achievements.ts` for duplication patterns
> 
>     * [x]  Assess whether DRYing is worth it, propose strategies if so
> 
>     * [x]  Propose missing test coverage for per-achievement progress calculation
> 
>     * [x]  Post final response
> 
> 
> ### Is DRYing worth it? Yes.
> 
> Looking at `src/data/achievements.ts` on this branch, essentially every achievement that got a `getProgress` also duplicates the exact same "how do I compute the current count" logic in both `condition` and `getProgress`. A few examples:
> 
>     * `daily-profit-1`/`2`/`3` (e.g. [src/data/achievements.ts:183-196](https://github.com/jeremyckahn/farmhand/blob/lstebner/progress-bar-per-achievement/src/data/achievements.ts#L183-L196)) call `getProfitRecord(state.recordSingleDayProfit, state.todaysRevenue, state.todaysLosses)` **twice** — once in `condition`, once in `getProgress` — with identical arguments.
> 
>     * `sale-goal-1`/`2`/`3`, `profit-average-1..6`, `i-am-rich-1..3`, `sell-10000-jack-o-lanterns`, `unlock-crop-price-guide`, `orchardist`, `piemaker`, `deforestation`, `landscaper` all repeat the same shape: `condition: state => X(state) >= goal` and `getProgress: state => ({ currentValue: X(state), goal })`, where `X(state)` is copy-pasted between the two.
> 
> 
> That's not just verbatim duplication, it's a correctness risk: `condition` and `getProgress` are two independent sources of truth that must stay in sync. If someone tweaks the threshold operator, the field being read, or the goal in one place and forgets the other, the achievement's "complete" state and its progress bar can silently disagree (e.g. bar shows 100% but the achievement never unlocks, or vice versa). A couple of achievements are already subtly inconsistent in _how_ they express the same idea — `purchase-all-cow-colors` ([achievements.ts:139-148](https://github.com/jeremyckahn/farmhand/blob/lstebner/progress-bar-per-achievement/src/data/achievements.ts#L139-L148)) uses `.every()` for `condition` but a separate `.filter().length` for `getProgress` to express the same "all colors purchased" check.
> ### Proposed strategy
> 
> Introduce one small helper that derives both `condition` and `getProgress` from a single `getCurrentValue(state)` function, and spread its result into each achievement definition. This keeps the existing IIFE/closure authoring pattern (no big restructuring) and only touches the ~20 achievements that need it:
> 
> ```ts
> const progressAchievement = (
>   goal: number,
>   getCurrentValue: (state: farmhand.state) => number
> ): Pick<farmhand.achievement, 'condition' | 'getProgress'> => ({
>   condition: state => getCurrentValue(state) >= goal,
>   getProgress: state => ({ currentValue: getCurrentValue(state), goal }),
> })
> ```
> 
> Usage (replacing the current `unlock-crop-price-guide` block):
> 
> ```ts
> ((goal = 10000) => ({
>   id: 'unlock-crop-price-guide',
>   name: 'Prove Yourself as a Farmer',
>   description: `...`,
>   rewardDescription: 'Crop Price Guide',
>   ...progressAchievement(goal, state => state.revenue),
>   reward: state => state,
> }))(),
> ```
> 
> And for the `daily-profit-*` family, the duplicated `getProfitRecord(...)` call collapses to one call site per achievement:
> 
> ```ts
> ((goal = 5000, reward = 25) => ({
>   id: 'daily-profit-1',
>   ...
>   ...progressAchievement(goal, state =>
>     getProfitRecord(state.recordSingleDayProfit, state.todaysRevenue, state.todaysLosses)
>   ),
>   reward: state => addItemToInventory(state, itemsMap['fertilizer'], reward, true),
> }))(),
> ```
> 
> This also naturally unifies `purchase-all-cow-colors`, since "all colors purchased" is equivalent to `filteredCount >= totalColorCount`:
> 
> ```ts
> ...progressAchievement(
>   Object.keys(standardCowColors).length,
>   state => Object.values(standardCowColors).filter(
>     color => (state.cowColorsPurchased[color] || 0) > 0
>   ).length
> ),
> ```
> 
> (worth a quick behavioral sanity check since it's a genuine logic change, not just a mechanical extraction — though functionally equivalent to the current `.every()`.)
> 
> **Alternative (bigger refactor, not recommended unless you want more structure):** a full `createProgressAchievement({ id, name, ..., goal, getCurrentValue, reward })` factory that returns a complete `farmhand.achievement`. This is more self-documenting ("this is a progress-style achievement") but means mixing factory calls and raw IIFEs in the same array, and is a larger diff for marginal benefit over the spread-helper approach above.
> 
> One secondary win: today `sumOfPiesMade` (used by `piemaker`) isn't wrapped in the existing `memoize` util the way `sumOfCropsHarvested`/`sumOfPartialRecordValues` are, so it recomputes the full reduce over `PIE_RECIPE_IDS` twice per achievement-processing pass. Worth wrapping it in `memoize` while touching this file, consistent with its siblings.
> ### Test coverage
> 
> `src/data/achievements.test.ts` currently only asserts `getProgress` for `orchardist`, `piemaker`, `deforestation`, `landscaper`, and the `i-am-rich` variants. Achievements that gained `getProgress` in this PR but have **no test asserting it**: `unlock-crop-price-guide`, `purchase-all-cow-colors`, `sell-10000-jack-o-lanterns`, `daily-profit-1/2/3`, `profit-average-1..6`, `sale-goal-1/2/3` — about 13 achievements.
> 
> If you adopt the `progressAchievement` helper above, the most valuable test isn't per-achievement repetition — it's a **single focused unit test for the helper itself**, since every achievement built on it inherits the guarantee:
> 
> ```ts
> describe('progressAchievement', () => {
>   test('condition is false and getProgress reflects currentValue below goal', () => {
>     const { condition, getProgress } = progressAchievement(10, () => 9)
>     expect(condition({} as any)).toEqual(false)
>     expect(getProgress({} as any)).toEqual({ currentValue: 9, goal: 10 })
>   })
> 
>   test('condition is true when currentValue meets goal', () => {
>     const { condition } = progressAchievement(10, () => 10)
>     expect(condition({} as any)).toEqual(true)
>   })
> 
>   test('condition is true when currentValue exceeds goal', () => {
>     const { condition } = progressAchievement(10, () => 11)
>     expect(condition({} as any)).toEqual(true)
>   })
> })
> ```
> 
> That covers the invariant "condition and getProgress can never disagree" for all ~20 achievements in one place, rather than needing to hand-verify it per achievement. It's also a good argument for the refactor beyond readability: it shrinks the required test surface, not just the source.
> 
> If you'd rather not introduce the helper right now, the fallback is the more tedious path: add a `getProgress` assertion to each of the ~13 achievements listed above, mirroring the existing `orchardist`/`piemaker` pattern (state with `currentValue` one below goal, assert `{ currentValue, goal }`). I'd also suggest one currently-untested edge case regardless of which path you take: an achievement whose `currentValue` _exceeds_ `goal` (e.g. selling 10,050 units in one action against a 10,000 goal) — confirming `getProgress` reports the true value while `Achievement.tsx`'s `Math.min(100, ...)` still clamps the displayed bar.
> 
> [Implement the `progressAchievement` helper and add helper tests →](https://claude.ai/code?q=In+src%2Fdata%2Fachievements.ts%2C+add+a+progressAchievement%28goal%2C+getCurrentValue%29+helper+that+returns+%7B+condition%2C+getProgress+%7D+derived+from+a+single+getCurrentValue%28state%29+function%2C+then+refactor+the+~20+achievements+that+currently+duplicate+the+same+computation+between+condition+and+getProgress+%28unlock-crop-price-guide%2C+purchase-all-cow-colors%2C+sell-10000-jack-o-lanterns%2C+daily-profit-1%2F2%2F3%2C+profit-average-1..6%2C+sale-goal-1%2F2%2F3%2C+i-am-rich-1..3%2C+orchardist%2C+piemaker%2C+deforestation%2C+landscaper%29+to+use+it+via+object+spread.+Also+wrap+sumOfPiesMade+in+the+existing+memoize+util+like+sumOfCropsHarvested+and+sumOfPartialRecordValues.+Add+a+focused+unit+test+suite+for+progressAchievement+in+src%2Fdata%2Fachievements.test.ts+covering+currentValue+below%2Fat%2Fabove+goal.+Branch%3A+lstebner%2Fprogress-bar-per-achievement&repo=jeremyckahn/farmhand)
> 
> This was an analysis/proposal per the request — no code changes were made. Happy to implement the helper and migrate the achievements if you'd like to go ahead with it. ([branch](https://github.com/jeremyckahn/farmhand/tree/lstebner/progress-bar-per-achievement))

@claude make the recommended changes.

jeremyckahn (approved):

PR #749, opened 2026-08-06
feat: add craftable Lightning Rods

### LINKED ISSUE #196: Anti-lightning Rod
Add an Anti-lightning Rod upgrade that costs $500,000 in the Shop. Once purchased and enabled, all storms are prevented.

_Originally posted by @jeremyckahn in https://github.com/jeremyckahn/farmhand/issues/193#issuecomment-910288604_

lstebner:
### What this PR does

Adds three new craftable Lightning Rod tiers (Basic/Superior/Ultimate), crafted from Iron/Silver/Gold Ingots via the Forge. Placed in an empty field plot like a Scarecrow, a rod intercepts the storm-destroys-scarecrow event and absorbs one strike per storm day instead of the scarecrow being destroyed, withstanding 10/25/100 strikes before being destroyed and refunding a small amount of raw ore. If multiple rods are placed, a storm randomly targets one of them (mirroring how `applyCrows.ts` picks a random plot).

Rods are a non-removable fixture once placed - the Hoe can't clear them - since their per-instance strike counter lives on the field plot itself rather than round-tripping through inventory (the codebase has no existing precedent for per-instance durability surviving a trip back into inventory).

closes #196 . once done, we can open up a followup issue for handling "energy" as a new resource

- `npx vitest run src/game-logic/reducers/setLightningRod.test.ts src/game-logic/reducers/applyPrecipitation.test.ts src/game-logic/reducers/clearPlot.test.ts`
- `npm run check:types` / `npm run lint`
- Manually: craft a rod in the Workshop's Forge tab (requires the Smelter and enough Iron/Silver/Gold Ingots sold), place it from the toolbelt into an empty field plot, confirm the Hoe can't remove it, and confirm it absorbs storm strikes (temporarily bumping `PRECIPITATION_CHANCE`/`STORM_CHANCE` in `constants.ts` makes this faster to observe) until it's destroyed and refunds ore.

Recipe ingredient quantities and `itemsSold` unlock thresholds are a reasonable starting point, not a final balance pass - easy to retune.

### Additional information

Screenshot from 2026-08-05 20-19-49
Screenshot from 2026-08-05 20-20-00
Screenshot from 2026-08-05 20-13-03
Screenshot from 2026-08-05 20-13-54

https://github.com/user-attachments/assets/70285698-6f5d-4012-b2f3-d066d50dab3c


lstebner:
@claude take a look

jeremyckahn:
@claude we want to understand how Lightning Rods fit into the game's overall balance. Give us tables of value data for all types of Ores, Ingots, Lightning Rods, and anything else that Ores/Ingots are used to produce.

jeremyckahn:
Here's a little tool we can use to play with the numbers to rebalance the game: https://claude.ai/public/artifacts/d04fe4e4-2bb6-4aae-bf60-35818db09d81

src/components/Field/Field.tsx
```diff
@@ -36,6 +39,12 @@ const colorBrownFertilize = 'rgba(125, 56, 0, 0.75)'
 const colorGreenOk = 'rgba(0, 255, 0, 0.5)'
 const colorYellow = 'rgba(255, 255, 0, 0.75)'
 
+const lightningRodImagesByItemId: Record<string, string> = {
+  'basic-lightning-rod': basicLightningRodImg,
+  'superior-lightning-rod': superiorLightningRodImg,
+  'ultimate-lightning-rod': ultimateLightningRodImg,
```
jeremyckahn: Let's change these object IDs to be dynamic references to the associated item `id` properties. Something like:

```suggestion
  [basicLightningRod.id]: basicLightningRodImg,
  [superiorLightningRod.id]: superiorLightningRodImg,
  [ultimateLightningRod.id]: ultimateLightningRodImg,
```

Let's also tighten the `lightningRodImagesByItemId` type definition accordingly, if possible.

src/farmhand.d.ts
```diff
@@ -169,6 +171,13 @@ declare namespace farmhand {
     // How many days a tree stays GROWN before dying (becoming DEAD) - see
     // getTreeLifeStage.ts. Undefined means the tree never dies.
     lifespan?: number
+    // How many lightning strikes a placed Lightning Rod can sustain before
+    // being destroyed - see plotContent.lightningStrikesSustained and
+    // applyPrecipitation.ts.
+    lightningStrikeCapacity?: number
+    // What gets added back to inventory when this item is destroyed by a
+    // lightning strike (see applyPrecipitation.ts).
+    destructionOreRefund?: { itemId: string; quantity: number }
```
jeremyckahn: Can we make the property name here a little more generic? This semantically ties the lightning rod to yielding ore, but I'd like to keep the door open to yielding other items in the future:

```suggestion
    destructionItemRefund?: { itemId: string; quantity: number }
```

src/game-logic/reducers/clearPlot.ts
```diff
@@ -24,6 +24,14 @@ export const clearPlot = (
     return state
   }
 
+  // Lightning rods are a non-removable fixture - the only way to lose one
+  // is by it exhausting its strike capacity (see applyPrecipitation.ts).
+  // Without this guard the Hoe would delete it for no refund, since it's
+  // neither replantable nor a crop/weed.
+  if (getPlotContentType(plotContent) === itemType.LIGHTNING_ROD) {
+    return state
+  }
```
jeremyckahn: I'm concerned about locking the player into any given item being stuck in the Field (even though the lightning rod eventually gets destroyed). I think this could lead to the player feeling stuck. IMO the player should be able to remove lightning rods from the Field with the Hoe and just trash it (with no ores being returned as the penalty). What do you think @lstebner?

src/img/items/lightning-rod-basic.png
jeremyckahn, line 1: Love the art! Nice and simple. :)

jeremyckahn (changes requested):
Overall this is looking great! I just have some recommended adjustments for the code and mechanics. Also a few design questions for you @lstebner:

- What do you think about naming the lightning rods after their associated materials (Bronze Lightning Rod, Silver Lightning Rod, Gold Lightning Rod) for consistency?
- Are lightning rods compatible with Rainbow Fertilizer? If not, let's make sure that they are so they get automatically replanted once exhausted (similar to the Scarecrow).

lstebner:
> Overall this is looking great! I just have some recommended adjustments for the code and mechanics. Also a few design questions for you @lstebner:
> 
>     * What do you think about naming the lightning rods after their associated materials (Bronze Lightning Rod, Silver Lightning Rod, Gold Lightning Rod) for consistency?

we discussed this in discord and came to an agreement that the naming was justified since these are not upgradeable items, they're distinctly different craftables. so it's a little unique.

>     * Are lightning rods compatible with Rainbow Fertilizer? If not, let's make sure that they are so they get automatically replanted once exhausted (similar to the Scarecrow).

updated! nice catch

src/components/Field/Field.tsx
lstebner: updated

src/farmhand.d.ts
lstebner: updated

src/game-logic/reducers/clearPlot.ts
lstebner: agreed. good call out. this has been updated and as we discussed in slack, if the rod has not sustained any damage the player will get the ore refund, otherwise the hoe simply destroys it

lstebner:
oh,and also to note: i rebalanced the recipes for the lightning rods. they now get more cost effective the higher the tier and i increased the ore gained on destruction slightly (2->4)

src/components/Plot/Plot.tsx
```diff
@@ -270,6 +300,20 @@ export const Plot = ({
...
+              <Typography>
+                Lightning strikes sustained:{' '}
+                {plotContent?.lightningStrikesSustained ?? 0} /{' '}
+                {item.lightningStrikeCapacity}
+              </Typography>
+            )}
+            {isLightningRod && propsFieldMode === CLEANUP && (
+              <Typography>
+                {lightningRodIsUndamaged && lightningRodDestructionYieldItem
+                  ? `Using the Hoe will scrap this for ${item?.destructionYield?.quantity} ${lightningRodDestructionYieldItem.name}.`
```
jeremyckahn: I think this would read 'Using the Hoe will scrap this for 4 Gold Ore.', which feels slightly awkward to me. How about 'Using the Hoe will scrap this for 4 **units of** Gold Ore'?

```suggestion
                  ? `Using the Hoe will scrap this for ${item?.destructionYield?.quantity} units ${lightningRodDestructionYieldItem.name}.`
```

(Maybe with some logic to handle the pluralization if `.quantity` is 1.)

src/components/Plot/Plot.tsx
```diff
@@ -270,6 +300,20 @@ export const Plot = ({
...
+                  : 'Using the Hoe will remove this - since it has taken damage, no materials will be recovered.'}
```
jeremyckahn: nit: Let's avoid dashes in display text since it might trigger people's AI sensors:
```suggestion
                  : 'Using the Hoe will remove this. Since it has taken damage, no materials will be recovered.'}
```

src/game-logic/reducers/applyPrecipitation.test.ts
```diff
@@ -60,6 +65,207 @@ describe('applyPrecipitation', () => {
...
+                message: LIGHTNING_ROD_DESTROYED_MESSAGE,
+                severity: 'error',
+              })
+            })
+          })
+        })
+      })
+
+      describe('multiple rods are planted', () => {
+        test('randomly targets one of them', () => {
```
jeremyckahn: We may want to revisit this mechanic in the future to allow lightning to strike all rods equally. This would consume more rods, but it would result in more captured energy (once we get there).

No change needed today, just sharing the thought. 🙂

jeremyckahn (approved):
A few minor suggestions, but LGTM! Great work @lstebner. This was an awesome idea! 🙂

src/game-logic/reducers/applyPrecipitation.test.ts
lstebner: yes! i was thinking that too. or potentially changing storms to have a varying amount of lightning strikes. that change could happen sooner than energy if you think it's a good idea

src/components/Plot/Plot.tsx
lstebner: this was way too wordy in general. this is one of the things i didn't pay attention to as claude was working. i've updated both tooltips you commented on to be much simpler. they're now like this:

- "Destroy and recoup 4 units of Iron Ore"
- "Will be destroyed" (if no ore will be regained)

what do you think?

jeremyckahn: That's much better! 👍

> i've updated both tooltips

I'm still seeing the old versions on GitHub. Was this change pushed?

src/components/Plot/Plot.tsx
jeremyckahn: @lstebner in case you missed this comment ☝️

src/game-logic/reducers/applyPrecipitation.test.ts
jeremyckahn: I'm glad we're on the same page! Thinking about this more, I'd lean towards doing this (hitting all the rods equally) now if it's not too much of a lift. That way we can avoid churn around mechanics, which players might find disorienting later.

jeremyckahn (approved):

src/components/Plot/Plot.tsx
lstebner: pluralization is handled now :+1:

src/components/Plot/Plot.tsx
lstebner: it wasn't! it is now :upside_down_face:

src/utils/pluralize.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    })
+  })
+
+  describe('count is not 1', () => {
+    test('returns the word with an s appended by default', () => {
+      expect(pluralize('unit', 0)).toEqual('units')
+      expect(pluralize('unit', 2)).toEqual('units')
+    })
+
+    describe('a pluralForm is provided', () => {
```
jeremyckahn, line 16: I suggest that we don't support this approach to pluralization. It strikes me as a potential pit of complexity to support the plural forms of the myriad entities within Farmhand. It seems simpler to standardize on "unit(s) of X" throughout the game. What do you think @lstebner?

src/game-logic/reducers/applyPrecipitation.test.ts
lstebner: done!

src/utils/pluralize.test.ts
jeremyckahn: We [discussed this on Discord](https://discord.com/channels/714539345050075176/847884323330850847/1535722685739569152) and agreed that this is generally useful and appropriate for this project. That said, it might be worth being careful about using it to pluralize entities in the game such as specific items, as that could become a lot to manage.

PR #754, opened 2026-08-23
refactor(types): [closes #703] make Factory class abstract and implement farmhand.Factory interface

### LINKED ISSUE #703: Make `Factory` class implement `farmhand.Factory` interface and declare as abstract
## Problem

`src/interfaces/Factory.ts` declares a `Factory` class but `src/farmhand.d.ts` declares a separate `Factory` interface. The class and interface are out of sync, and the class does not implement the interface.

Current `src/interfaces/Factory.ts`:

```typescript
export class Factory {
  generate(): farmhand.item | farmhand.item[] {
    throw new Error("generate() must be implemented by subclass")
  }
}
```

Current `src/farmhand.d.ts:348-350`:

```typescript
interface Factory {
  generate(): item | item[] | null
}
```

The interface allows `null` as a return value, but the class does not. The class is an abstract base class but is not declared as `abstract`. Subclasses (`OreFactory`, `CoalFactory`, `StoneFactory`, `ResourceFactory`) do not explicitly implement the interface.

### Step 1: Make `Factory` class implement the interface

```typescript
export class Factory implements farmhand.Factory {
  generate(): farmhand.item | farmhand.item[] | null {
    throw new Error("generate() must be implemented by subclass")
  }
}
```

### Step 2: Make `Factory` an abstract class

Since `Factory` is meant to be subclassed and its `generate()` method always throws, it should be abstract:

```typescript
export abstract class Factory implements farmhand.Factory {
  abstract generate(): farmhand.item | farmhand.item[] | null
}
```

### Step 3: Update subclasses to extend properly

Check all subclasses (`OreFactory`, `CoalFactory`, `StoneFactory`, `ResourceFactory`) and ensure they properly extend `Factory` and return the correct types.

### Step 4: Remove `Factory` declaration from `farmhand.d.ts`

Since the class now implements the interface, the duplicate declaration in `farmhand.d.ts:348-350` should be removed to avoid confusion.

### Step 5: Run type checker

```bash
npm run check:types
```

## Acceptance Criteria

- `Factory` class implements `farmhand.Factory` interface
- `Factory` is declared as `abstract` with an abstract `generate()` method
- All subclasses properly extend `Factory` and implement `generate()` with correct return types
- `Factory` interface removed from `farmhand.d.ts`
- `npm run check:types` passes with no errors
- `npm test` passes

jeremyckahn:
### What this PR does

- Makes `Factory` an abstract class that implements the global `farmhand.Factory` interface (`src/interfaces/Factory.ts`), with an abstract `generate(): farmhand.item | farmhand.item[] | null`. The single declaration stays in `src/farmhand.d.ts`, alongside every other `farmhand.*` type.
- Fixes latent fallout of the widened contract: `ResourceFactory.generateResources()` guards against wrapping a `null` return value into `[null]`.
- Adds a regression test locking that guard in (`ResourceFactory.test.ts`).

```bash
npm run lint          # clean, no disable comments needed
npm run check:types   # main + e2e projects pass
npm test -- --run     # 1150 tests; all green except one known unrelated flaky timeout below
```

Note: `field-purchase.test.ts` occasionally times out (default 5s) during full-suite runs under heavy machine load on some machines; it passes when re-run in isolation and is untouched by this change.

- Consciously deviates from step 4 of #703 ("remove the Factory declaration from farmhand.d.ts"): keeping it there as the single source of truth next to all other `farmhand.*` types. Moving it next to the class instead would require a `declare global` block inside an implementation file plus one scoped `no-shadow` disable; centralizing is what lets this PR stay free of both.
- The new guard in `ResourceFactory.generateResources()` only changes behavior if a subclass actually returns `null`; none of today's three subclasses (`OreFactory`, `CoalFactory`, `StoneFactory`) do — they all return arrays, unchanged here. Now covered by the added test.

### Additional information

Closes #703

### DIFF
--- a/src/factories/ResourceFactory.test.ts
+++ b/src/factories/ResourceFactory.test.ts
@@ -2,7 +2,6 @@ import { vi } from 'vitest'
 
 import { randomNumberService } from '../common/services/randomNumber.js'
 import { itemType, toolLevel } from '../enums.js'
-// eslint-disable-next-line no-unused-vars
 import { Factory } from '../interfaces/Factory.js'
 import { randomChoice } from '../utils/randomChoice.js'
 
--- a/src/factories/ResourceFactory.test.ts
+++ b/src/factories/ResourceFactory.test.ts
@@ -2,6 +2,8 @@ import { vi } from 'vitest'
 
 import { randomNumberService } from '../common/services/randomNumber.js'
 import { itemType, toolLevel } from '../enums.js'
+// eslint-disable-next-line no-unused-vars
+import { Factory } from '../interfaces/Factory.js'
 import { randomChoice } from '../utils/randomChoice.js'
 
 import ResourceFactory from './ResourceFactory.js'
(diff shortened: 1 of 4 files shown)


jeremyckahn:
@claude review

jeremyckahn:
Addressed in cb2aa36c — added `does not include resources when a factory generate returns null` to `ResourceFactory.test.ts`. It queues a `null` return on the cached ore factory (via the existing module mocks) and asserts `generateResources()` yields an empty array, locking in the guard. Lint/typecheck pass; full suite green apart from the one known unrelated flaky timeout in `field-purchase.test.ts` under load.

PR #755, opened 2026-08-27
refactor(cowpen): convert Cow class component to a function component

jeremyckahn:
### What this PR does

- Converts the `Cow` component from a class component to a function component using `useState`/`useRef`/`useEffect`. This is a pure refactor with no behavioral change.
- Class mechanics map to hooks as follows:
  - `this.state` → `useState` (`x`/`y` stay in a single `position` object so the tween render callback updates both coordinates in one render, as before)
  - `this.tweenable`, timeout IDs, and `isComponentMounted` → `useRef`
  - `componentDidMount`/`componentWillUnmount` → a mount `useEffect` and its cleanup
  - `componentDidUpdate` → effects keyed on `isSelected` and `cow.happinessBoostsToday`, with previous values tracked in refs
  - The static durations become module constants (they were not referenced externally; `CowProps` remains exported unchanged)

- `npx vitest run src/components/CowPen/CowPen.test.tsx` — all 15 tests pass (the only suite that renders `Cow`)
- `npm run lint` and `npm run check:types` pass
- Full test run shows only the pre-existing flaky failures also present on `develop` (`game-logic/field-purchase`, `shell/pending-notifications`, `shell/game-saving`, `game-logic/item-purchase` timing flakes) — no new failures
- Manual: open the game and watch cows wander; select/deselect a cow — movement pauses while selected and resumes after; flips, transitions, tooltip, and the hug heart animation all behave as before

- The test suite uses fake timers, so long-running tween behavior was not exercised end-to-end in CI. I've traced the call order of `random()` and the tween/cleanup lifecycles against the class version to keep them identical, but a quick manual playthrough is still worth it.

### Additional information

- `Cow` is only used by `CowPen.tsx` and `CowPen.test.tsx`; no other call sites or static references exist.

### DIFF
--- a/src/components/CowPen/Cow.tsx
+++ b/src/components/CowPen/Cow.tsx
@@ -5,6 +5,7 @@ import Typography from '@mui/material/Typography/index.js'
 import classNames from 'classnames'
 import { useCallback, useEffect, useState } from 'react'
 import { TweenState, Tweenable } from 'shifty'
+import { useIsMounted } from 'usehooks-ts'
 
 import { random } from '../../common/utils.js'
 import { LEFT, RIGHT } from '../../constants.js'
@@ -56,6 +57,7 @@ export const Cow = ({
   const [prevHappinessBoostsToday, setPrevHappinessBoostsToday] = useState(
     cow.happinessBoostsToday
   )
+  const isMounted = useIsMounted()
   const { x, y } = position
 
   const move = useCallback(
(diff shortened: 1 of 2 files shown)


jeremyckahn:
@claude review

jeremyckahn:
Both items addressed in e984cbb1:

1. **`newY` draw timing**: `newY` is now computed lazily immediately before the position tween call (after the optional flip await), exactly where the class version drew it, instead of upfront alongside `newX`. The `random()` call order across the flip is restored, so seeded playthroughs stay bit-for-bit reproducible.
2. **Position tween `render` typing**: the callback now takes a `TweenState` and uses the same `typeof === 'number'` guard pattern as the flip tween's `render` — no more `any`.

`CowPen.test.tsx` passes (19 tests), `tsc --noEmit` and `eslint` are clean.

PR #762, opened 2026-09-09
Add seasons and seasonal crop demand pricing

### LINKED ISSUE #85: Seasons
discussion #84 has been started for figuring out the specifics for this feature.

lstebner:
### What this PR does

Adds four in-game seasons (Spring → Summer → Fall → Winter, 10 days each, cycling) and lets crops optionally declare a high demand season (+20% sell price) and/or a low demand season (-20% sell price). Both are optional and independent; a crop with neither stays at 1x value year-round.

- New `season` enum and `highDemandSeasons`/`lowDemandSeasons` optional fields on `farmhand.item`.
- `getCurrentSeason(dayCount)` derives the season deterministically from `dayCount` (10 days/season) - no new persisted state field needed.
- `getSeasonalDemandStatus`/`getSeasonalDemandMultiplier` classify a crop against the current season and compute its ±20% multiplier.
- Applied to sale price in `sellItem.ts` and `Item.tsx`, guarded the same way as the existing achievement sale bonus (`#140`) so seeds bought and instantly resold can't exploit seasonal timing - only harvested crops are affected.
- UI: current season name shown centered in the header; "High Demand"/"Low Demand" text shown under each crop's sell price in the sell list.
- Demand data added to 16 crops per the agreed table (asparagus, carrot, corn, garlic, grape, jalapeno, olive, pea, potato, pumpkin, spinach, strawberry, sunflower, sweet potato, tomato, watermelon); onion, soybean, sugarcane, and wheat are left at steady demand.

this closes #85 

- `npx vitest run` - full unit/component suite passes (1175 tests), including new tests for `getCurrentSeason`, `getSeasonalDemandStatus`, `getSeasonalDemandMultiplier`, seasonal cases in `sellItem.test.ts`, and UI rendering in `Item.test.tsx`/`AppBar.test.tsx`.
- `npm run check:types` and `npm run lint` both pass.
- `e2e/tests/seasons.test.ts` (with fixture `e2e/fixtures/seasonal-demand.json`) exercises the header season indicator and sell-list demand labels end-to-end.
- Manually: `npm start`, then in the browser console `window.farmhand.setState({ dayCount: N })` to jump across season boundaries and watch the header label and each crop's sell-card demand text/price update.

- The seasonal bonus/penalty is currently only applied to harvested crop sales (not seeds, and not fermented/keg sales like wine) - flagging in case that scope should be wider.
- Demand-season assignments per crop follow the table worked out in planning; open to adjusting any of them.
- The +/- 20% was just a starting point. happy to change this

here's the seasonal demand chart for every crop, also happy to take suggestions on changes. crops with no high/low have a consistent price all year round.

| Crop | High Demand | Low Demand |
|---|---|---|
| Asparagus | Summer | Winter |
| Carrot | Spring | — |
| Corn | Summer, Fall | Spring, Winter |
| Garlic | Fall, Winter | Spring |
| Grape | Spring, Fall | Winter |
| Jalapeño | Spring, Summer | Fall, Winter |
| Olive | Winter | Fall |
| Onion | — | — |
| Pea | Spring | Winter |
| Potato | Fall, Winter | Spring, Summer |
| Pumpkin | Fall | Spring |
| Soybean | — | — |
| Spinach | Spring, Summer | Winter |
| Strawberry | Summer | Winter |
| Sugarcane | — | — |
| Sunflower | Summer | Winter |
| Sweet Potato | Fall, Winter | Spring |
| Tomato | Spring, Summer | Winter |
| Watermelon | Summer | Fall, Winter |
| Wheat | — | — |

### Additional information

Kept the UI simple for this first iteration. the name of the season is displayed in the header, and a simple "low demand" / "high demand" bit of text is added to the crop in the list of items.

image
Screenshot from 2026-09-08 20-35-14
Screenshot from 2026-09-08 20-34-50

https://claude.ai/code/session_01E6o6FVHMrogmHgkZEiXf9h


lstebner:
@claude review

lstebner:
Thanks for catching this. Fixed - the demand label in `Item.tsx` is now gated with the same `shopItemIds.has(id)` guard used for the price multiplier, so it only shows when the price actually reflects it (a new regression test covers selling `carrot-seed` during its would-be high demand season).

While at it: this surfaced two pre-existing e2e tests (`multiplayer/price-fluctuation.test.ts`, `stats/revenue.test.ts`) that hardcode a Carrot sell price/revenue against the `crops-mature` fixture, whose `dayCount` (6) falls in Spring - Carrot's newly-added high demand season. Updated their expected values to account for the +20% bonus (verified against the real reducer output, not hand-derived).

Also added an e2e test (`e2e/tests/seasons.test.ts`) that exercises the actual seasonal price adjustment end-to-end - not just the label/price text, but clicking Sell and confirming the credited money reflects the seasonal multiplier - using a new fixture with `valueAdjustments` pinned to 1 so the math is deterministic.

Pushed in 3f2210a4.

jeremyckahn (commented):
It's super exciting to see this finally happen! We've been talking about this for years. 🙂

I haven't had a chance to review the code or even give this much of a playthrough yet, but I wanted to give some early feedback based on the PR description and screenshots provided:

- What do you think about making seasons 15 days? I suppose it's arbitrary either way, but 10 feels a little too short to me. 15 feels like it gives the player a chance to "settle in" to the season a bit more and play to ones that are favorable to their goals more.
- It seems reasonable to declare significant UI/graphical changes out of scope for this first iteration, but what do you think about adding a simple effect of some kind on the Field, Forest, and Cow screens to indicate the current season? I'm thinking we could use a CSS `filter` or something to add an orange cast for Fall, blue for Winter, and perhaps subtle yellow for Summer.
- Could we add an indicator of what day of the season it is? Perhaps a tooltip on the season name at the top that says something like "Day X of Y of [season]"?
- For the "High demand"/"Low demand" text, what do you think about coloring the text accordingly? Perhaps green for high, red for low? (Agent: look for prior art for red/green colors to use within the `Paper`/item cards)
- Right now the crop value indicators (the up/down arrows next to the sell price) can indicate the opposite of the "High demand"/"Low demand" text. I think this will be confusing for players. I'm thinking that either the arrow should always be consistent with the season demand text, or we should adjust the demand text to say something less contradictory like "In season"/"Out of season."

I'll hold off on a code review for now in case significant changes are made given the above.

lstebner:
great feedback @jeremyckahn . no opposition to any of that from me, will make some updates!

e2e/tests/multiplayer/price-fluctuation.test.ts
```diff
@@ -14,8 +15,11 @@ test('uses server-based price values', async ({ page }) => {
     .locator('.Plot')
     .first()
     .click()
+  // crops-mature's dayCount (6) falls within Spring, carrot's configured
+  // high demand season, so the base fluctuation-adjusted price ($28.72) is
+  // boosted by SEASON_HIGH_DEMAND_BONUS.
   await expect(page.getByRole('complementary')).toContainText(
-    'CarrotSell price: $28.72Total: $28.72'
+    'CarrotSell price: $34.46Total: $34.46'
```
jeremyckahn, line 22: It seems safer to leave this test's expectations as-is (or at least tweak them to avoid seasonal value adjustments) and instead add an additional test that intentionally incorporates seasonal value adjustments. What do you think @lstebner?

e2e/tests/stats/revenue.test.ts
```diff
@@ -18,14 +18,17 @@ test('should track and manage revenue records', async ({ page }) => {
     .getByRole('button', { name: 'Sell' })
     .first()
     .click()
+  // crops-mature's dayCount (6) falls within Spring, carrot's configured
+  // high demand season, so the sold carrot's base revenue ($27.28) is
+  // boosted by SEASON_HIGH_DEMAND_BONUS to $33.02.
   await page.getByRole('button', { name: 'View your stats (s)' }).click()
   await expect(page.locator('#stats-modal-content')).toContainText(
-    "Today's Revenue$27.28Today's Losses$0.00Today's Profit$27.28Record Single Day Profit$27.28Current Profitability Streak0 daysRecord Profitability Streak0 days7-day Profit Average-$20.53Record 7-day Profit Average$0.00All-Time Total Revenue$27.28"
+    "Today's Revenue$33.02Today's Losses$0.00Today's Profit$33.02Record Single Day Profit$33.02Current Profitability Streak0 daysRecord Profitability Streak0 days7-day Profit Average-$20.53Record 7-day Profit Average$0.00All-Time Total Revenue$33.02"
```
jeremyckahn, line 26: Similar note about avoiding the modification of stable tests like this; it seems safer to add an additional, dedicated test case for season pricing.

e2e/tests/seasons.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  await expect(wheatCard.getByText('In Season')).not.toBeVisible()
+  await expect(wheatCard.getByText('Out of Season')).not.toBeVisible()
+})
+
+// seasonal-demand's dayCount (0) is SPRING, and its valueAdjustments for
+// carrot/corn/wheat are pinned to 1 (no daily fluctuation noise), so each
+// crop's sell price is just its base value times its seasonal multiplier:
+// carrot (SPRING high demand) 25 * 1.2 = 30, corn (SPRING low demand)
+// 70 * 0.8 = 56, wheat (no configured demand) 28 * 1 = 28.
+test('should apply the seasonal demand price adjustment when selling crops', async ({
```
jeremyckahn: Would it be worth demoting this to a unit test? E2E tests are quite heavy and slow, and I'm not sure that the lower-level price calculation logic being exercised here rises to the level of warranting E2E validation.

src/components/AppBar/AppBar.tsx
```diff
@@ -153,6 +163,16 @@ export const AppBar = ({
...
         {viewTitle}
       </Typography>
+      <Typography
+        {...{
+          className: 'season-display',
+          variant: 'h2',
+        }}
+      >
+        {`Day ${getDayOfSeason(dayCount)} of ${
+          seasonNameMap[getCurrentSeason(dayCount)]
```
jeremyckahn, line 180: On mobile, I recommend shifting this UI over to the left as it overlaps with the money indicator on smaller screens:

Image

jeremyckahn (changes requested):
I wasn't able to get through all the code yet, but I wanted to share my feedback so far. I'll review the rest when I can!

src/utils/getCurrentSeason.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    expect(getCurrentSeason(15)).toEqual(season.SUMMER)
+  })
+
+  test('returns FALL for the first day of fall', () => {
+    expect(getCurrentSeason(30)).toEqual(season.FALL)
+  })
+
+  test('returns WINTER for the first day of winter', () => {
+    expect(getCurrentSeason(45)).toEqual(season.WINTER)
+  })
```
jeremyckahn: nit: These tests are all pretty similar, so consider using `test.each` to templatize them.

src/enums.ts
```diff
@@ -290,3 +290,16 @@ export const grapeVariety = {
...
+ * @enum
+ */
+export const season = {
+  SPRING: 'SPRING',
+  SUMMER: 'SUMMER',
+  FALL: 'FALL',
+  WINTER: 'WINTER',
+} as const
+
+export type season = typeof season[keyof typeof season]
```
jeremyckahn, line 305: I know it's valid syntax to have a variable and type share a name (`season`), but it seems harder to follow than if they were different. What do you think about renaming this type to something like `seasonType` to clearly distinguish the two?

jeremyckahn (changes requested):
Finished the rest of my code review!

@claude given this PR affects a lot of complex subsystems (many of which we haven't looked at in a while), think deeply about any unhandled edge cases we should consider before merging this.

e2e/tests/multiplayer/price-fluctuation.test.ts
lstebner: we can't actually leave this one alone because carrot happens to be in demand in spring. so the price actually did change because of the addition of seasonal pricing. so, we either have to adjust the price, change the day the crop is sold on to make it a different season, or use a different crop for this test. i think each path leaves us with a similar amount of alteration.

e2e/tests/stats/revenue.test.ts
lstebner: same reasoning. though i also agree we should have tests that specifically test in and out of season pricing to make sure it works as expected.

e2e/tests/seasons.test.ts
lstebner: sure. i got into the mindset of e2e all changes and guess i got carried away haha

src/components/AppBar/AppBar.tsx
lstebner: great catch!

lstebner: i also noticed between 400-450 the title is shown in the header _and_ in the main panel so i'm going to bring those in sync to both change at 450

src/enums.ts
lstebner: this seems to be the pattern the file currently uses. changing this type to `seasonType` would make it the only one that does that (the others just happen to have "Type" in each name i.e. `const cropType` and `type cropType`). so this seems like a convention decision to make and then apply to the whole file. i say we leave as is for this PR

lstebner:
> **1. Loan garnishment is computed from the un-boosted price, so seasonal swings leak past garnishment.** 

updated

> **2. Value-fluctuation arrows can still contradict the "In Season"/"Out of Season" text.** 

this is by design now. the arrow indicates the market price, and the in/out-of-season indicator is a separate price mod

> **3. `AppBar`'s `.season-display` mobile overlap is still unresolved.** 

fixed
 
> **4. Season length (15 days) and the "New Year" milestone (365 days) don't line up.** 

i think this is fine. if we want we can change the year length to 360 to make them compatible, but i don't find it too bothersome.
 
> **5. Inventory can "age out" of the season it was harvested in.** 

this is intentional. the player has to weigh the bonus against the current market price and decide what works best for them.

> **6. No data-level guard against a crop declaring the same season as both high and low demand.** 

added

e2e/tests/multiplayer/price-fluctuation.test.ts
jeremyckahn: > change the day the crop is sold on to make it a different season

I'm thinking that's what makes the most sense here for the sake of observable stability. It might also be worth doing the following:

- Add a new similar test that uses a high demand season (essentially what have here now)
- Add a new similar test that uses a low demand season

I think that would cover our bases without materially changing a legacy test that validates established logic. At the moment, I don't know that we have evidence that seasonal price adjustments are fully compatible with multiplayer pricing (though I expect that they are). We could probably use `test.each` to keep this tidy.

e2e/tests/stats/revenue.test.ts
jeremyckahn: Between this and [the other thread](https://github.com/jeremyckahn/farmhand/pull/762#discussion_r3974804388), I think we're converging on having test coverage for all variety of seasonal adjustments here. 😄

e2e/tests/seasons.test.ts
jeremyckahn: All good! It's not wasted effort, the (temporary) e2e test was helpful for quick one-time validation. I just think we don't need to merge it in.

src/enums.ts
jeremyckahn: Ah, good catch. This must be an artifact of my scorched earth TypeScript conversion. 😅

I agree we should leave this as-is for now and consider addressing it later.

src/components/SeasonFilterTuner/SeasonFilterTuner.tsx
jeremyckahn, line 1: Love this! We discussed removing this on Discord, but we can use this comment as a reminder to do so before merging. 🙂

src/components/Stage/Stage.tsx
```diff
@@ -73,6 +80,16 @@ export const Stage = ({
     [stageFocusType.FOREST]: forestFloorBg,
   }[stageFocus as string]
 
+  // The Field, Cow Pen, and Forest screens get a season-specific visual
+  // filter (see seasonFilterMap) applied via the `.stage-background` layer
+  // below - scoped to just the background, not the crop/cow/tree sprites
+  // rendered on top of it.
+  const isFieldCowPenOrForest = ([
```
jeremyckahn, line 87: nit: Let's adjust this to be focused on the relevant outcome:
```suggestion
  const hasSeasonalColorShift = ([
```

(Or something like that.)

src/game-logic/reducers/sellItem.ts
jeremyckahn, line 1: @claude I think all these changes are correct, but do a sanity check re-review of this file to make sure we're not unintentionally wrecking any pricing logic.

e2e/tests/multiplayer/price-fluctuation.test.ts
```diff
@@ -1,11 +1,31 @@
...
+// in src/react-app-env.d.ts, which isn't part of this project's
+// TypeScript scope - so it's redeclared locally here with the shape this
+// file actually relies on.
+declare global {
+  interface Window {
+    farmhand?: {
+      setState: (partialState: Record<string, unknown>) => void
+    }
+  }
+}
```
jeremyckahn, line 16: This seems like something other E2E tests could benefit from. Could it moved somewhere such that it would apply to them?

e2e/tests/stats/revenue.test.ts
```diff
@@ -2,9 +2,30 @@ import { test, expect } from '@playwright/test'
...
+// in src/react-app-env.d.ts, which isn't part of this project's
+// TypeScript scope - so it's redeclared locally here with the shape this
+// file actually relies on.
+declare global {
+  interface Window {
+    farmhand?: {
+      setState: (partialState: Record<string, unknown>) => void
+    }
+  }
+}
```
jeremyckahn, line 16: Similar to the other comment, let's see if we can deduplicate this for the E2E tests.

jeremyckahn (changes requested):

jeremyckahn:
> > **4. Season length (15 days) and the "New Year" milestone (365 days) don't line up.**
> 
> i think this is fine. if we want we can change the year length to 360 to make them compatible, but i don't find it too bothersome.

@lstebner Can we revisit this decision? 365 is obviously the length of a real year, but this is going to seem confusing to players if we have two separate concepts of a year in the game. It seems like it would be a lot more straightforward if `EXPERIENCE_VALUES.NEW_YEAR` was granted at `state.dayCount % (SEASON_LENGTH_DAYS * 4) === 0`.

lstebner:
> > > **4. Season length (15 days) and the "New Year" milestone (365 days) don't line up.**
> 
> > 
> 
> > i think this is fine. if we want we can change the year length to 360 to make them compatible, but i don't find it too bothersome.
> 
> 
> 
> @lstebner Can we revisit this decision? 365 is obviously the length of a real year, but this is going to seem confusing to players if we have two separate concepts of a year in the game. It seems like it would be a lot more straightforward if `EXPERIENCE_VALUES.NEW_YEAR` was granted at `state.dayCount % (SEASON_LENGTH_DAYS * 4) === 0`.

yeah, that makes sense. i'll update it

e2e/tests/multiplayer/price-fluctuation.test.ts
lstebner: updated this to be shared

e2e/tests/stats/revenue.test.ts
lstebner: updated this

lstebner:
> > > **4. Season length (15 days) and the "New Year" milestone (365 days) don't line up.**
> > 
> > 
> > i think this is fine. if we want we can change the year length to 360 to make them compatible, but i don't find it too bothersome.
> 
> @lstebner Can we revisit this decision? 365 is obviously the length of a real year, but this is going to seem confusing to players if we have two separate concepts of a year in the game. It seems like it would be a lot more straightforward if `EXPERIENCE_VALUES.NEW_YEAR` was granted at `state.dayCount % (SEASON_LENGTH_DAYS * 4) === 0`.

this has been updated

src/components/Stage/Stage.tsx
```diff
@@ -85,39 +99,49 @@ export const Stage = ({
...
-            }
-          : {}),
-        // floorboard.png is a 144x48 (3:1) tile, unlike every other
-        // background image here which is square - keep it at the same 2x
-        // pixel-art scale as the rest (48px source -> 96px displayed) by
-        // matching that same ratio instead of the shared square size.
-        ...(stageFocus === stageFocusType.CELLAR
-          ? { backgroundSize: '288px 96px' }
-          : {}),
+        '& .stage-background': {
```
jeremyckahn, line 116: I noticed a major regression when scrolling down on mobile. It seems that scrolling down just pushes the whole stage area up and reveals the static Rainbow Cow loading background behind the game. This affects both nearly all the screens in some way. My read is that the backgrounds are no longer treated as fixed (which they should be presented as), so they scroll away. I presume it's related to these styling changes:

Image

Compared to Production:

Image

@claude do you know what might be causing this, and what the fix might be?

jeremyckahn (changes requested):
One major mobile regression to address, but otherwise this is looking good!

src/components/Stage/Stage.tsx
jeremyckahn: > ### Suggested fix
> 
> The most robust fix is defense-in-depth at the document level, not just patching the Stage layer: add an explicit `overflow: hidden` (and `height: 100%`) to `html, body` in `globalStyleOverrides` (`src/mui-theme.ts`), e.g.:
> 
> ```ts
> const globalStyleOverrides = {
>   html: { height: '100%', overflow: 'hidden' },
>   body: { height: '100%', overflow: 'hidden', overscrollBehavior: 'contain' },
>   ...
> }
> ```
> 
> This guarantees the document itself can never be scrolled, regardless of any transient sizing/paint gap in `.stage-background` or future changes — all scrolling stays confined to `.Stage`'s own `overflow: auto`, where it belongs. It's a one-line-ish, low-risk addition since `.farmhand-root` already covers the full viewport by design; nothing should currently depend on the document being scrollable.

I tried this locally and it just changes the failure mode:
image

@lstebner this might take some more iteration to properly address. :confused:

lstebner: woah! good find. that would have been a bad bug to ship. i'll take a look

lstebner: looks like this isn't just a mobile issue, it happens on desktop too (just noting)

lstebner: fix is in

Screenshot from 2026-09-16 08-28-09
Screenshot from 2026-09-16 08-27-54

jeremyckahn (approved):
LGTM! Awesome work @lstebner. This is a major improvement to Farmhand!

PR #767, opened 2026-09-17
feat(ci): Make release type a dropdown in release workflow

lstebner:
### What this PR does

Changes the `releaseType` input on the "Release New Version" workflow (`.github/workflows/run-release.yml`) from a free-text string to a `choice` dropdown with `patch`, `minor`, and `major` options. Default stays `patch`. This prevents typos when manually triggering a release from the Actions UI.

Trigger the "Release New Version" workflow manually from the Actions tab in GitHub and confirm the "Release type" input renders as a dropdown with `patch`/`minor`/`major` choices instead of a free-text field.

None.

### Additional information


lstebner:
@jeremyckahn if you have a strong preference for the freeform text input feel free to close this but i always worry about making a typo when i do a release so i wanted to suggest this improvement

jeremyckahn (approved):
Good idea @lstebner! We do lose some functional flexibility for setting an arbitrary version number with this, but we've never needed to do so and can't foresee a situation in which we would. So, I think this is fine.

lstebner:
> Good idea @lstebner! We do lose some functional flexibility for setting an arbitrary version number with this, but we've never needed to do so and can't foresee a situation in which we would. So, I think this is fine.

i actually didn't even know that was possible, i thought the string here was explicitly used to determine how to bump the version. i've added support for an "other" type that gives the custom input that used to exist

